PR #1, opened 2022-01-12
Update Configuration.md

jaklke:
The working directories are used during the configuration update operation. Mentioning 'configuration backups'  can be confusing. In the app itself, we also changed the tooltip to state it holds files (rather than configuration backups).


MariekeGO:
Update implemented based on feedback.

PR #9, opened 2022-01-25
Correction to list view creation procedure

MariekeGO:
- StartTime and EndTime moved from shape data to component options
- StartTime and EndTime removed from shape data field overview
- Markdown adjusted for easier editing


user-guide/part_2/visio/Creating_a_list_view.md
```diff
@@ -101,7 +118,7 @@ If you set the *Source* shape data field to “Elements” or “Services”, th
     View=MyView
     ```
 
-    An alternative syntax, *ViewName=\<name>*, can be used from DataMiner 10.0.5 onwards. From this version onwards, you can also use the "==" operator instead of the "=" operator. If the *ViewName* syntax is used, DataMiner will first try to filter by name, and then by ID in case the name cannot be found. If the *View* syntax is used, DataMiner will first try to filter by ID ,and then by name if the ID cannot be found. The filter can contain only one *View* or *ViewName* part.
+    An alternative syntax, *ViewName=\<name>*, can be used from DataMiner 10.0.5 onwards. From this version onwards, you can also use the "==" operator instead of the "=" operator. If the *ViewName* syntax is used, DataMiner will first try to filter by name, and then by ID in case the name cannot be found. If the *View* syntax is used, DataMiner will first try to filter by ID, and then by name if the ID cannot be found. The filter can contain only one *View* or *ViewName* part.
```
MariekeTest, line 121: The comma after ID should be removed.

PR #15, opened 2022-02-01
Removed XML WebService (removed since RN 27237)

jens-gellynck:


MariekeGO:
To avoid conflicts because of the ongoing change to the structure of the source files, this PR has been added as part of [this commit](https://github.com/SkylineCommunications/dataminer-docs/commit/4e07e288ccef078b6188ae18fd6c0093765a6705).

PR #28, opened 2022-02-09
DIS 2.36 update

spetebroodt:


SimonV-Skyline (commented):
See comments.
Will also still need to consider what to do with class library packages info.

PR #73, opened 2022-03-11
SLCloud.xml can still contain old IP address

jens-gellynck:
When the IP changes, SLCloud.xml can still contain the IP address, potentially causing SYN-floods on the network.
Added this file to the list of files you need to manually update when the IP address changes.


thomieboy:
The indentation does not seem to be displayed correctly here in GitHub (Not sure why)

MariekeGO (approved):

PR #122, opened 2022-04-04
ValidationState was still described as a bool

jens-gellynck:
Changed description of ValidationState from bool to the enum it actually represents


MariekeGO:
@PedroDebevere Could you verify this change?

PedroDebevere:
Suggested change is OK.

MariekeGO (approved):

PR #126, opened 2022-04-05
Replace SLNetRaw by SendSLNetMessage

jens-gellynck:
Engine.SLNetRaw can be dangerous as it circumvents a lot of security checks. Update the help to use SendSLNetMessages instead.


jens-gellynck:
@PedroDebevere could you also review these?

PedroDebevere:
Suggested changes are OK.

MariekeGO (approved):

PR #154, opened 2022-04-13
Correct case for field names

beckerrobin:


beckerrobin:
This "error" exists in the entire API docs. I think it's worth thinking about if you want it to maybe just look good (as it does now but the API returns an error) or actually be technically correct, with the initial letter as lower case. The other solution would of course be to change the API to accept the field names as they are stated in the docs.

MariekeGO:
> This "error" exists in the entire API docs. I think it's worth thinking about if you want it to maybe just look good (as it does now but the API returns an error) or actually be technically correct, with the initial letter as lower case. The other solution would of course be to change the API to accept the field names as they are stated in the docs.

Indeed, we'll have to correct this throughout the API docs. I'll look into it asap. Thanks for the feedback!

MariekeGO:
This change has been implemented throughout the API docs with a different commit.

PR #202, opened 2022-05-02
Added info around Unicode with websockets.

MaximBOU:
When setting the unicode tag on a protocol, websockets of type text can break. Some information explaining a good approach has been added.

### DIFF
--- a/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
@@ -81,7 +81,7 @@ When the WebSocket server closes the connection for some reason, the element goe
 
 By default, DataMiner sends the WebSocket messages as binary data (i.e. a frame with Opcode 0x2, RFC 6455). Some WebSocket servers will reply with an \[ACK\] packet but ignore the message as the server does not support binary formatted messages.
 
-If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, RFC 6455). 
+If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455#section-11.8)). 
 
 In case the server supports text frames, it should now respond to this command. Note that the WebSocketMessageType tag is only supported from DataMiner 9.5.1 (RN 14177) onwards.
 
--- a/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
@@ -87,7 +87,7 @@ In case the server supports text frames, it should now respond to this command. 
 
 ## Unicode protocols
 
-If the protocol is set to use Unicode and the response for the WebSocket is saved in a parameter of type "string", this behavior can cause issues in case Getparameter() is used to fetch the WebSocket response from within a QAction. When the Unicode tag is set, string parameters will be saved as UTF-16. The WebSocket, however, will try to store its response as UTF-8. When Getparameter() is used, the received value will therefore not be encoded correctly, which will cause special characters to not be shown correctly. To solve this, add the following tags to the QAction that parses the data:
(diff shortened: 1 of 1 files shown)


MariekeGO:

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -81,6 +81,29 @@ When the WebSocket server closes the connection for some reason, the element goe
...
+If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, RFC 6455). 
+
+In case the server supports text frames, it should now respond to this command. Note that the WebSocketMessageType tag is only supported from DataMiner 9.5.1 (RN 14177) onwards.
+
+## Unicode protocols
+
+If the protocol is set to use Unicode and the response for the WebSocket is saved in a parameter of type "string", this behavior can cause issues in case Getparameter() is used to fetch the WebSocket response from within a QAction. When the Unicode tag is set, string parameters will be saved as UTF-16. The WebSocket, however, will try to store its response as UTF-8. When Getparameter() is used, the received value will therefore not be encoded correctly, which will cause special characters to not be shown correctly. To solve this, add the following tags to the QAction that parses the data:
+
+```cs
```
PedroDebevere, line 92:
```suggestion
```xml
```

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -81,6 +81,29 @@ When the WebSocket server closes the connection for some reason, the element goe
...
+If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, RFC 6455). 
+
+In case the server supports text frames, it should now respond to this command. Note that the WebSocketMessageType tag is only supported from DataMiner 9.5.1 (RN 14177) onwards.
+
+## Unicode protocols
+
+If the protocol is set to use Unicode and the response for the WebSocket is saved in a parameter of type "string", this behavior can cause issues in case Getparameter() is used to fetch the WebSocket response from within a QAction. When the Unicode tag is set, string parameters will be saved as UTF-16. The WebSocket, however, will try to store its response as UTF-8. When Getparameter() is used, the received value will therefore not be encoded correctly, which will cause special characters to not be shown correctly. To solve this, add the following tags to the QAction that parses the data:
```
PedroDebevere, line 90:
```suggestion
If the protocol is set to use Unicode and the response for the WebSocket is saved in a parameter of type "string", this behavior can cause issues in case [GetParameter()](xref:Skyline.DataMiner.Scripting.SLProtocol.GetParameter(System.Int32)) is used to fetch the WebSocket response from within a QAction. If the Unicode tag is set, string parameters will be saved as UTF-16. The WebSocket, however, will try to store its response as UTF-8. When [GetParameter()](xref:Skyline.DataMiner.Scripting.SLProtocol.GetParameter(System.Int32)) is used, the received value will therefore not be encoded correctly, which will cause special characters to not be shown correctly. To solve this, add the following tags to the QAction that parses the data:
```

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -81,6 +81,29 @@ When the WebSocket server closes the connection for some reason, the element goe
 
 By default, DataMiner sends the WebSocket messages as binary data (i.e. a frame with Opcode 0x2, RFC 6455). Some WebSocket servers will reply with an \[ACK\] packet but ignore the message as the server does not support binary formatted messages.
 
-If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, RFC 6455).
+If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, RFC 6455).
```
PedroDebevere, line 84:
```suggestion
If the message you want to send only contains text and the server does not seem to support binary formatted messages, try to add `<WebSocketMessageType>text</WebSocketMessageType>` to the `<Command>`. This will result in the command being sent as UTF-8 encoded text (Opcode 0x1, [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455#section-11.8)). 
```

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -81,6 +81,29 @@ When the WebSocket server closes the connection for some reason, the element goe
 
 By default, DataMiner sends the WebSocket messages as binary data (i.e. a frame with Opcode 0x2, RFC 6455). Some WebSocket servers will reply with an \[ACK\] packet but ignore the message as the server does not support binary formatted messages.
```
PedroDebevere, line 82:
```suggestion
By default, DataMiner sends the WebSocket messages as binary data (i.e. a frame with Opcode 0x2, [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455#section-11.8)). Some WebSocket servers will reply with an \[ACK\] packet but ignore the message as the server does not support binary formatted messages.
```

PR #209, opened 2022-05-03
RN 33208

MariekeGO:
@wimme Could you check if this is OK to merge?


wimme (approved):
Some more info:

- License requests can be send to dataminer.licensing@skyline.be
- Once "Low-Code Apps" has been added, the new license file can be downloaded from the License Portal https://licensing.skyline.be/

PR #223, opened 2022-05-06
RN 33422

MariekeGO:
@JordyGit Could you check if this is OK to merge?


JordyGit:
@MariekeGO don't think this is needed, we don't do it for the other commands either which also have the buttons etc. The reason why we've added it to the RN was because we first had an initial version without those things for views.

PR #231, opened 2022-05-12
Using Azure Cognitive Search and fully reworked UI for search results page

stanley-skyline:
DoxFX's original search box is removed via code and a custom implementation is created that uses queries to https://docs-srch.search.windows.net (Azure Cognitive Search).

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -60,6 +60,7 @@
     "markdownEngineName": "markdig",
     "noLangKeyword": false,
     "keepFileLink": false,
-    "cleanupCacheHistory": false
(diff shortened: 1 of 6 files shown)


docfx.json
```diff
@@ -61,7 +60,6 @@
     "markdownEngineName": "markdig",
     "noLangKeyword": false,
     "keepFileLink": false,
-    "cleanupCacheHistory": false,
-    "disableGitFeatures": false
```
stanley-skyline, line 65: @BenBo-slc, can you explain why this was removed?

MariekeGO (approved):

PR #236, opened 2022-05-16
Bug fix new search (skip table of contents files)

stanley-skyline:
Results with toc.html files (table of contents) + without title are now skipped


stanley-skyline:
@BertVandenberghe, this resolves the issue when you search for "trending" for example.

PR #264, opened 2022-06-02
Initialize JSON YAxisResources variable

HansSolos:
This change has made the example better so it is more usable to the users


HansSolos (commented):
Finetune of change: YAxisResources became YaxisResources

MariekeGO (approved):

PR #265, opened 2022-06-03
DataMiner MR version info added for .NET 5

MariekeGO:
@baptistepa OK to add this info, based on https://community.dataminer.services/question/service-dataminer-cloudgateway-failed-to-start/?


baptistepa (approved):
Yes looks good. Thanks for adding this.

PR #277, opened 2022-06-08
Removed wrong options for group

MichielOda:
According to [Actions Overview](https://docs.dataminer.services/develop/devguide/Connector/LogicActionsOverview.html) and the specific pages [set next](https://docs.dataminer.services/develop/devguide/Connector/Actions/ActionSetNext.html) & [stop current group](https://docs.dataminer.services/develop/devguide/Connector/Actions/ActionStopCurrentGroup.html) these are not to be used with the Action/On == 'group'.


MariekeGO:
@PedroDebevere Could you confirm if this is OK to merge?

PedroDebevere:
OK to merge.

MariekeGO (approved):

PR #288, opened 2022-06-10
Update Protocol.Actions.Action.On-nr.md

SimonV-Skyline:


SimonV-Skyline:
Had a quick chat with Michiel and he is currently working on the same improvements but covering more than me.
So I'll close this PR to avoid double work for reviewers.

PR #294, opened 2022-06-13
DCP Admin app: New Deployments & Outgoing Shares pages

MariekeGO:
@AntonPlanckeSLC Could you confirm if this is OK to merge?

### DIFF
--- a/user-guide/Cloud_Platform/CloudAdminApp/Viewing_info_on_deployments.md
+++ b/user-guide/Cloud_Platform/CloudAdminApp/Viewing_info_on_deployments.md
@@ -1,0 +1,21 @@
+---
+uid: Viewing_info_on_deployments
+---
+
+# Viewing information on deployments
+
+In the DCP Admin app, you can view information about all the deployments that have been done to the DataMiner System via the Nodes page, via the Catalog, or using a GitHub pipeline with our GitHub action.
+
+To do so:
+
+1. In the Admin app, check whether the correct organization is mentioned in the header bar.
+
+1. If a different organization should be selected, click the organization selector in the top-right corner and select the organization in the list.
+
+   ![Organization selector](~/user-guide/images/CloudAdmin_Selector.png)
+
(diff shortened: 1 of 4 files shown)


user-guide/Cloud_Platform/CloudAdminApp/Viewing_info_on_deployments.md
```diff
@@ -0,0 +1,21 @@
...
+
+1. If a different organization should be selected, click the organization selector in the top-right corner and select the organization in the list.
+
+   ![Organization selector](~/user-guide/images/CloudAdmin_Selector.png)
+
+1. In the pane on the left, under *DataMiner Systems*, select your DataMiner System and select the *Deployments* page.
+
+   This page details what has been deployed, when, and by whom, and whether the deployment succeeded, is pending, or failed.
+
+   Click an item in the list to view more detailed information, including version information and event information that can be used for debugging. For example, in case a pending deployment is queued because of other deployments, you will be able to see this in the event information.
```
JordyGit, line 21: maybe we should drop the example?

user-guide/Cloud_Platform/CloudAdminApp/Viewing_info_on_deployments.md
```diff
@@ -0,0 +1,21 @@
...
+
+To do so:
+
+1. In the Admin app, check whether the correct organization is mentioned in the header bar.
+
+1. If a different organization should be selected, click the organization selector in the top-right corner and select the organization in the list.
+
+   ![Organization selector](~/user-guide/images/CloudAdmin_Selector.png)
+
+1. In the pane on the left, under *DataMiner Systems*, select your DataMiner System and select the *Deployments* page.
```
JordyGit, line 17: Don't know if we need to mention (in general that) the DMS button won't be shown if the user doesn't has the permission? Atm all buttons should be available available for members except the 'Keys' button (which is from Admin & Owner) if i remember correctly.

PR #302, opened 2022-06-13
Version info RN 33122 added

MariekeGO:
@Pieter-JanV I've added the version info for RN 33122 with this pull request. Could you check if this is OK to merge?


MariekeGO:
@PedroDebevere When I do a test run of this PR, I get "Badly formed XML comment" warnings while generating metadata:
image

Do you know what could be the cause of this? I've tried cleaning up the syntax, but this hasn't helped.

PR #316, opened 2022-06-20
Added package manifest schema documentation

PedroDebevere:


PedroDebevere:
@MariekeGO: we could add an additional entry next to the root element (e.g. "SchemaTypes"), which would then list all types defined in the schema.

MariekeGO:
@PedroDebevere I've added SchemaTypes to the TOC as suggested. OK to merge like this?

PedroDebevere:
@MariekeGO : OK to merge

MariekeGO (approved):

PR #335, opened 2022-06-24
Improved ClassLibraryExamples.md slightly

andr9528:
Fixed a few errors/inconsistencies in code example, and simplyfied/improved 'GetElementAfterStartupComplete(...)'

The commit contains a more detailed list of changes, along with links to some of the related Microsoft documentation.

Do note that during my changes i had copied the code into my VS as part of a precompiled QAction, to verify the syntax. I noticed that 'SLProtocol' does not contains a definition for '.GetDms()'. Might just be for me who is missing something, but i thought i would mention it, for others to potentially fix.

EDIT: I have check it with DocFX, where it looked fine.

### DIFF
--- a/develop/devguide/ClassLibrary/ClassLibraryExamples.md
+++ b/develop/devguide/ClassLibrary/ClassLibraryExamples.md
@@ -12,7 +12,7 @@ The following example creates a new element and then waits until the element is 
 /// <summary>
 /// The QAction entry point.
 /// </summary>
-/// <param name="protocol">Link with SLProtocol process.</param>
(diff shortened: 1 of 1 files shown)


MariekeGO:

MariekeGO (approved):

andr9528:
Found the SLProtocol.GetDms() extention method after digging arround. Dunno why my VS didn't suggest adding the needed using...

PedroDebevere:
Hi @andr9528 : thanks for the improvements and feedback! It's indeed strange that VS doesn't suggest the using, if the project is referencing the library it should show it as a suggestion.

PR #340, opened 2022-06-24
RN 33580

MariekeGO:


karlmasschelein (approved):
Verified upon the latest version of 'https://docs.dataminer.services/user-guide/Basic_Functionality/Visio/SRM/Embedding_a_Resource_Manager_component.html' and including an upcoming update in the same section.

PR #345, opened 2022-06-27
Update Creating_a_list_view.md

brmdias:
Adding a detail of the applicable values for ReservationInstance.Status.
If a user is building a new Visio, they need to know which values correspond to which Status.

I based on the following mapping: https://docs.dataminer.services/user-guide/Basic_Functionality/Visio/miscellaneous/Creating_a_list_view.html#list-view-filters:~:text=StateOfSelection%3A%20Available%20from,Canceled%20(7)
Better to double-check with Orchestration domain.


MariekeGO:
@HansSolos Could you check if this OK to merge?

MariekeGO (approved):

PR #354, opened 2022-06-29
Update Embedding_a_Service_Manager_component.md

karlmasschelein:
Doubt between use of shapedata Options versus ComponentOptions


HansSolos (approved):
readability is better indeed

MariekeGO (approved):

PR #355, opened 2022-06-29
updated NT_SNMP_GET docs

PedroDebevere:


develop/api/NotifyTypes/NT_SNMP_GET.md
```diff
@@ -148,24 +137,201 @@ if (response.Length == 1)
...
+    //// ...
+  }
+}
+```
+
+### TABLEV2
+
+To retrieve a table via the TABLEV2 method, the entry in oidInfo should be a semicolon-separated list of OIDs of the columns that should be retrieved (e.g. `1.3.6.1.2.1.2.2.1.2;1.3.6.1.2.1.2.2.1.3:TABLEV2`).
+
+This will result in GetBulkRequest PDUs being sent out. The number of maxRepitions to use in the GetBulkRequest can be specified in the elementInfo object. The default value is 6.
```
MariekeGO, line 231: Could it be that this should be "maxRepetitions" instead of "maxRepitions"?

PedroDebevere: @MariekeGO: indeed, that is a typo.

PR #369, opened 2022-07-06
RN 33784

MariekeGO:


PedroDebevere:
OK to merge. Suggestion: mention both dashboards app and monitoring app as the popup should now also show up in the mobile monitoring app.

PR #379, opened 2022-07-12
Connectors troubleshooting

spetebroodt:


develop/devguide/Connector/Troubleshooting/Investigating_a_protocol_thread_RTE.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Investigating a protocol thread RTE
+
+Below, you can read how you can find the root cause when the following alarm is shown in the Alarm Console:
+
+```txt
+Thread problem in SLProtocol.exe: [DriverName/DriverVersion] ElementName – Protocolthread
+```
+
+### User skills required
```
PedroDebevere, line 13: This line should probably start with just two '#' (i.e. ## User skills required). The linter in VS Code should give the following warning: MD001/heading-increment/header-increment: Heading levels should only increment by one level at a time [Expected: h2; Actual: h3]

develop/devguide/Connector/Troubleshooting/Investigating_a_protocol_thread_RTE.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Basic knowledge of drivers.
+- Basic knowledge of how to use DataMiner Cube.
+
+### What does this alarm mean?
+
+This alarm indicates that the protocol thread that is used to execute polling or logical flows has stopped showing activity for a specific amount of time. By default, this time is 15 minutes.
```
PedroDebevere, line 20:
```suggestion
This alarm indicates that the protocol thread that is used to execute polling or logical flows has stopped showing activity for a specific duration. By default, this duration is 15 minutes.
```

develop/devguide/Connector/Troubleshooting/QAction_exceptions/Identifying_the_source_of_an_exception.md
```diff
@@ -0,0 +1,128 @@
...
+```csharp
+System.IndexOutOfRangeException: Index was outside the bounds of the array.
+    at MailServer.get_Address()
+    at QAction.Process()
+    at QAction.Run(SLProtocol protocol)
+```
+
+This will be more useful, as you will have more information to go on and you will not keep looking at the `ProcessMethod`. However, it is not ideal yet, as there is no indication that the actual problem is in the `RetrieveCustomDomain` method.
+
+Also, note that adding the "throw ex" will generate two SonarCube warnings:
```
PedroDebevere, line 110:
```suggestion
Also, note that adding the "throw ex" will generate two SonarQube warnings:
```

develop/devguide/Connector/Troubleshooting/QAction_exceptions/Identifying_the_source_of_an_exception.md
```diff
@@ -0,0 +1,128 @@
...
+
+Getting a property is actually calling a `get_<PropertyName>` method in the background, which can be optimized away just like any other method. Adding the try-catch in the previous example prevented the optimization from the compiler, which let it show the get method call in the stack trace of the exception.
+
+### Summary
+
+- Like the SonarCube warning mentions, property getters should never cause exceptions. If exceptions are possible, consider catching them and gracefully returning an exception value, or making a method instead of a getter. A property getter is also considered something simple, meaning that the developer calling the property assumes that this can be called thousands of times in a loop without performance impact. If more complexity is behind it, a method is a better choice, as it flags to the developer that it might be better to call the result once, store it in a variable before the loop starts, and then use that variable inside the loop.
```
PedroDebevere, line 124:
```suggestion
- Like the SonarQube warning mentions, property getters should never cause exceptions. If exceptions are possible, consider catching them and gracefully returning an exception value, or making a method instead of a getter. A property getter is also considered something simple, meaning that the developer calling the property assumes that this can be called thousands of times in a loop without performance impact. If more complexity is behind it, a method is a better choice, as it flags to the developer that it might be better to call the result once, store it in a variable before the loop starts, and then use that variable inside the loop.
```

develop/devguide/Connector/Troubleshooting/QAction_exceptions/System_IndexOutOf­RangeException.md
```diff
@@ -0,0 +1,64 @@
...
+uint[] currentIds = new uint[] { Parameter.counter_data_offload_4003, Parameter.data_offload_folder_5000 };
+object[] currentParam = (object[])protocol.GetParameters(currentIds);
+int counterDataOffload_4003 = Convert.ToInt32(currentParam[0]);
+string dataOffloadFolder_5000 = Convert.ToString(currentParam[1]);
+string wrongDataOffloadFolder_5002 = Convert.ToString(currentParam[2]);
+```
+
+Accessing `currentParam[2]` will fail as only 2 parameters were requested. It could be that this parameter was requested in the past but got removed, and the fact that the processing of the parameter also had to be removed was overlooked.
+
+A second problem that could occur is that it is not sure if the object that is returned will be in the expected format. This could be a null object or it might not have the expected size.
```
PedroDebevere, line 50: sure => certain?

develop/devguide/Connector/Troubleshooting/QAction_exceptions/System_NullReference­Exception.md
```diff
@@ -0,0 +1,81 @@
...
+
+# System.NullReferenceException
+
+*"Object reference not set to an instance of an object"*
+
+While there is no need to instantiate the basic local variables, like int, bool, double, etc. most other objects need to be created by calling the constructor with the magic word "new". If methods or properties of the object are accessed without the object being instantiated, a null reference exception will be thrown.
+
+Three special cases:
+
+- String variables can be null, but do not have the "new" key word to be instantiated. Simply assign a value to them.
```
PedroDebevere, line 13: String variables can be null (as it is a reference type),

develop/devguide/Connector/Troubleshooting/QAction_exceptions/System_NullReference­Exception.md
```diff
@@ -0,0 +1,81 @@
...
+Verify every member before accessing it:
+
+```csharp
+if(person != null 
+&& person.LocationInfo != null 
+&& person.LocationInfo.Street != null 
+&& person.LocationInfo.Street.StartsWith("Test"))
+```
+
+The downside of this is that when the depth of items to check is too large, SonarCube will complain that there are too many conditional operators used in the expression. It will also affect the complexity calculation of the code.
```
PedroDebevere, line 40: SonarQube

develop/devguide/Connector/Troubleshooting/QAction_exceptions/System_NullReference­Exception.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Solution 1.B
+
+Since C#6 and later, a null-conditional member access operator `?.` is available, also known as the Elvis operator. This will return null when one of the members is null. This can then be combined with the null-coalescing operator `??`.
```
PedroDebevere, line 44: C# 6

develop/devguide/Connector/Troubleshooting/QAction_exceptions/System_NullReference­Exception.md
```diff
@@ -0,0 +1,81 @@
...
+
+Add an extra check to make sure the string is not null or empty before executing the method:
+
+```csharp
+if (!String.IsNullOrEmpty(folder) && folder.EndsWith("\\"))
+```
+
+### Solution 2.B
+
+Use the Elvis operator (requires C#6):
```
PedroDebevere, line 77: C# 6

develop/devguide/Connector/Troubleshooting/Why_is_serial_data_not_visible_in_DataMiner.md
```diff
@@ -0,0 +1,159 @@
+---
+uid: Why_is_serial_data_not_visible_in_DataMiner
+---
+
+# Why is serial data not visible in DataMiner?
+
+Below, you can find how you can determine the reason why info retrieved via serial communication is not shown in DataMiner.
+
+### User skills required
```
PedroDebevere, line 9: ## instead of ###

spetebroodt: changed to ##

PR #381, opened 2022-07-12
Update Overview_Support_Escalations_SLAs.md

bfsantos:
Replacing this section with an outside of business specific section. this one already exists in https://docs.dataminer.services/user-guide/Reference/Support_services_agreement.html#schedule-1-support-plan
Would also propose to move this Support services agreement from 'Reference' to this section on 'DataMiner Overview', so all support clauses are together


MariekeGO:
@bfsantos Based on a discussion with @SammyDW, I've replaced the link on the Support Services page with a link to the Support services agreement. This agreement will also be moved as you suggested.

MariekeGO (approved):

bfsantos:
> @bfsantos This file now no longer mentions what our regular business hours are. Should this information indeed no longer be mentioned? If so, we'll still need to remove the note on the Support Services page that refers to this page for information on business hours.

I removed it because it's a duplicate of this one. https://docs.dataminer.services/dataminer-overview/Support%20Services/Support_services_agreement.html#response-and-corrective-times Perhaps we can include a link to it?

PR #406, opened 2022-07-29
Update Using_SourceTree.md

janstaelensskyline:
Finalizing Task: https://collaboration.dataminer.services/task/189321


spetebroodt:
under a folder in the C driver > this should probably be "on the C: drive"?

janstaelensskyline:
Yes C: drive is fine. Could also be D: drive. Or any drive actually it's more that it should be a folder directly under a root drive. Perhaps we need to reword that a bit to make it clear. Any suggestions?

PR #416, opened 2022-08-04
RN 33879

MariekeGO:
@skyline-rge, could you check if this is OK to merge?

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_app/Available_visualizations/Charts/Line_and_area_chart.md
+++ b/user-guide/Advanced_Modules/Dashboards_app/Available_visualizations/Charts/Line_and_area_chart.md
@@ -15,6 +15,9 @@ To configure the component:
 
 1. Apply one or more parameter data feeds. See [Applying a data feed](xref:Configuring_dashboard_components#applying-a-data-feed).
 
+   > [!NOTE]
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Modules/Dashboards_app/Available_visualizations/Charts/Line_and_area_chart.md
```diff
@@ -104,6 +107,23 @@ To configure the component:
...
+1. Create a query data feed. See [Configuring query data feeds](xref:Configuring_GQI_feeds).
+
+1. In the *Component* > *Settings* tab, configure the following fields in the *Dimensions* section:
+
+   - *Query*: The query data feed you want to use.
+   - *X axis*: The column that should be used for the X-axis data.
+   - *Y axis*: The column that should be used for the Y-axis data.
+
+   > [!NOTE]
+   > If you want to show a trend chart, make sure the query results are sorted on the X-axis column.
```
skyline-rge, line 123: Maybe we should make this note a bit more general and give some more details.
I.e. something like:

`Data points are connected by a line in the order they appear in the query result. If you want to create a trend line, make sure the query results are sorted on the desired axis column.`

PR #452, opened 2022-08-23
Removed space in example

JordyGit:


JordyGit:
Possibly this example is also somewhere in the real docs about these features? (this is the release note)

JordyGit:
Maybe we even should remove examples from release notes and just refer to the docs? We consistently have all our [release notes](https://docs.dataminer.services/release-notes/Cloud/DCP_change_log.html) short and clean until the most recent ones we added examples, which we shouldn't have in a RN imo?

MariekeGO (approved):

MariekeGO:
> Maybe we even should remove examples from release notes and just refer to the docs? We consistently have all our [release notes](https://docs.dataminer.services/release-notes/Cloud/DCP_change_log.html) short and clean until the most recent ones we added examples, which we shouldn't have in a RN imo?

The RN info, unlike the docs info, never changes, which makes it an important source of information for someone using an older version of the software. If you want to know what something was like in a specific release only (e.g. for troubleshooting), before all sorts of changes get implemented in later versions, you need detailed information in the release notes. We therefore need to make sure the RNs contain sufficient information to remain useful even many versions later, when the documentation has become very different. We may be able to look into making them more concise, but only if that condition can remain fulfilled.

PR #474, opened 2022-08-31
Update Protocol.Params.Param.Interprete.Sequence.md

janstaelensskyline:


MariekeGO:
@janstaelensskyline, could you check if the reviewed version still matches what you wanted to say?

janstaelensskyline:
It's good.

PR #475, opened 2022-08-31
Update Viewing_info_on_deployments.md

KevinVerm:
Added tip section regarding deployment logging on the dataminer system.

@baptistepa can you also have a look?


baptistepa (approved):

MariekeGO:
@baptistepa, I've rephrased the first sentence in the tip. Is this still correct?

baptistepa (changes requested):
It is not mandatory to install an Artifact Deployer on each DMA, and an Artifact Deployer instance does not need to be next to a DMA at all. As long as it has access to the NATS bus it will be able to work. I would change the wording to something like: "You can find the logging on the server where the Artifact Deployer is installed under the following folder: "

MariekeGO (approved):

PR #479, opened 2022-09-01
Update GetHistogramTrendDataForParameter.md

PieterVanCompernolle:
Via investigation on https://collaboration.dataminer.services/task/153924, found out that the option 'IntervalAmount' is no longer implemented/supported.


PieterVanCompernolle:
Yes, the option is also not used in GetHistogramSnapshotForTableColumnParameter & GetHistogramSnapshotForView

MariekeGO (approved):

PR #492, opened 2022-09-05
RTDisplay true requirement on placeholder

SkylineThomasCR:
Is validated by DIS

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Display-pageOrder.md
+++ b/develop/schemadoc/Protocol/Protocol.Display-pageOrder.md
@@ -28,15 +28,16 @@ When a page starts with “-”, Cube will see this as a page separator. This is
 
 ### Including web interfaces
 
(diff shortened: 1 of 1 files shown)


SkylineThomasCR:
Not sure if an example of a parameter would further benefit the page
``` XML
<Param id="1">
	<Name>ThumbnailMosaicUrl</Name>
	<Description>Thumbnail Mosaic URL</Description>
	<Type>read</Type>
	<Interprete>
		<RawType>other</RawType>
		<Type>string</Type>
		<LengthType>next param</LengthType>
		<DefaultValue>probe/core/summary/thumbsMosaicDoc</DefaultValue>
	</Interprete>
	<Display>
		<RTDisplay>true</RTDisplay>
	</Display>
</Param>
```

MariekeGO:
I'm not sure if adding that example would bring much added value, but I've added cross-references instead to make it easier for users to find more information about the mentioned elements.
@PedroDebevere, could you advise and let us know if this is OK to merge?

PedroDebevere:
@MariekeGO : I think the added cross-references should be sufficient. OK to merge.

PR #494, opened 2022-09-06
Changes for DCP184323

SLC-Laurens-Vercruysse:
Updating Tombstone threshold docs.
Updating requirement of Cassandra 4.0 for CC.


MariekeGO:
@SkylineCookieMonster, could you check if anything still needs to be changed before this is merged?

PR #495, opened 2022-09-06
RN 34280

MariekeGO:
@de-webfee, could you check if this is OK to merge?

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_app/Using_dashboard_feeds/Configuring_GQI_feeds.md
+++ b/user-guide/Advanced_Modules/Dashboards_app/Using_dashboard_feeds/Configuring_GQI_feeds.md
@@ -385,7 +385,7 @@ From DataMiner 10.2.0/10.1.3 onwards, instead of specifying an exact filter valu
 
   If *Type* is set to *Query rows*, instead of a property, you can select the columns from the table containing the query rows. However, note that you will only be able to select columns that are compatible with the type of column you are filtering.
 
-From DataMiner 10.1.11 onwards, an additional option, *Return no rows when feed is empty*, is available. When you select this option, in case the feed is empty, an empty table will be returned instead of the entire table.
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Modules/Dashboards_app/Using_dashboard_feeds/Configuring_GQI_feeds.md
```diff
@@ -385,7 +385,7 @@ From DataMiner 10.2.0/10.1.3 onwards, instead of specifying an exact filter valu
 
   If *Type* is set to *Query rows*, instead of a property, you can select the columns from the table containing the query rows. However, note that you will only be able to select columns that are compatible with the type of column you are filtering.
 
-From DataMiner 10.1.11 onwards, an additional option, *Return no rows when feed is empty*, is available. When you select this option, in case the feed is empty, an empty table will be returned instead of the entire table.
+From DataMiner 10.2.10/10.3.0 onwards, an additional option, *When feed is empty, show* is available, which allows you to select what should be shown in case the field is empty: nothing, an empty table ("*empty values*"), or the full table ("*everything*"). Prior to this, from DataMiner 10.1.11 onwards, instead the option *Return no rows when feed is empty* is available. When you select this option, in case the feed is empty, an empty table will be returned instead of the entire table.
```
de-webfee, line 388: the **empty values** option will not always result in an empty table. The empty values will filter that specific column so that only its empty values remain.

| Column 1      | Column 2    |
| ----------- | -------------------- |
| A | (empty) |
| B | 2 | GQIPage |

Filtering Column 2 with an empty feed, with the empty values option enabled, would result in:

| Column 1      | Column 2    |
| ----------- | -------------------- |
| A | (empty) |

It also depends if the data supports filtering on empty values in GQI core (atm none does), then a fallback to the empty table (**nothing**) will be done

de-webfee (changes requested):

PR #501, opened 2022-09-08
Add detail about manager name required by ManagerStoreReinitializeCustomManagerRequest

peter-vanpoucke:
Next to the module name the `ManagerStoreReinitializeCustomManagerRequest` also needs the manager name. That's *DomManager* for this use case. Added that extra info.


thomieboy (approved):

user-guide/Advanced_Modules/DOM/DOM_ModuleSettings.md
```diff
@@ -23,6 +23,7 @@ helper.ModuleSettings.Create(settings);
 > When you update the module settings of a manager that is already running, this will only take effect when the manager is (re-)initialized. This happens when:
 >
 > - You send a `ManagerStoreReinitializeCustomManagerRequest` in a script. This will reinitialize the manager on all Agents in the cluster. If a manager is not present on one of the Agents, it will be ignored.
+Initialize the `ManagerStoreReinitializeCustomManagerRequest` message with the name of the module that needs to be reinitialized and *DomManager* as manager name.
```
MariekeGO, line 26:
```suggestion
Initialize the `ManagerStoreReinitializeCustomManagerRequest` message with the name of the module that needs to be reinitialized and *DomManager* as the manager name.
```

MariekeGO: Please also remove the return between the current lines 25 and 26 to keep the markdown syntax clean. If two lines should be in the same paragraph, no return is needed. If they should not be in the same paragraph, there should be an empty line between them (but this doesn't seem appropriate here).

MariekeGO (changes requested):

MariekeGO (approved):

PR #508, opened 2022-09-09
IDP CI Types - Export location

JensVandewalle:
Added default location to where CI Types are exported and where you can change it


MariekeGO:
@jaklke, is this OK to merge? Since we don't usually mention information about the hidden IDP elements, do we keep the second part of the note or should this be left out?

jaklke:
Indeed, I rather not have this mentioned in the Help, instead, I think we should make sure the location is configurable from the Visual Overview instead. 
Anything hidden and not available via the main IDP UI (The Visual Overview on the IDP Application together with the Interactive Automation Scripts provided) is not officially supported and we cannot guarantee the full feature set.
It could be features that have been disabled until fully implemented or auxiliary parameters and tables used in the background by the solution and that using without official support can cause system instability and is not advised.

MariekeGO (approved):

PR #528, opened 2022-09-14
Retrieved vs. custom type table columns clarified

MariekeGO:
Based on https://community.dataminer.services/question/retrieved-vs-custom-type-table-columns/


PedroDebevere (approved):
changes OK

PR #558, opened 2022-09-23
Update Displaying_a_DataMiner_Map_in_a_DataMiner_dashboard.md

KevinVerm:
Added example url's


MariekeGO:
@KevinVerm Please avoid adding duplicate information when possible. If something is already mentioned on a different page, add a link to that page instead of repeating the info (like in my latest commit here). This way you reduce the risk of people missing important notes or information getting updated in only one of the locations.

MariekeGO (approved):

PR #559, opened 2022-09-26
Low-Code Apps security info added

MariekeGO:
@wobog, is this sufficient for task [193925](https://collaboration.dataminer.services/task/193925)?


wobog:
> @wobog, is this sufficient for task [193925](https://collaboration.dataminer.services/task/193925)?

Looks good. For some reason, I would have the tendency to also add something like
"If you restrict the view and edit access, only the users with edit access wil be able to edit the app."

Not sure if you find that to be redundant or not.

MariekeGO:
> > @wobog, is this sufficient for task [193925](https://collaboration.dataminer.services/task/193925)?
> 
> Looks good. For some reason, I would have the tendency to also add something like "If you restrict the view and edit access, only the users with edit access will be able to edit the app."
> 
> Not sure if you find that to be redundant or not.

Done!

PR #562, opened 2022-09-28
Improve documentation for cloud DMZ setups.

Sven-at-Skyline:


LouisSkyline (approved):

baptistepa (approved):

MariekeGO:
@Sven-at-Skyline, could you check if everything is still OK after my changes?
Note that I've removed the exact number from "Install the 4 DxMs that need to be on the DMZ", because generally it's best not to add a number in such stem sentences if it's not absolutely necessary. When there's an update, typically such sentences get overlooked, and then there's a mismatch between them and the information that follows them.

Sven-at-Skyline:
@MariekeGO Thanks, lgtm!

MariekeGO (approved):

PR #563, opened 2022-09-28
Improved login info

MariekeGO:


JordyGit:
Documentation is correct, but actually the users never manually have to enter the command 'login' normally because the bot will automatically prompt them if this is needed :) We were required by Microsoft Teams to add the login/logout commands.

JordyGit:
After reading it again it's actually correct, my bad :)

MariekeGO:
> After reading it again it's actually correct, my bad :) would just indicate it's not needed because it will automatically prompt you

Better with my latest edit?

JordyGit (approved):

PR #590, opened 2022-10-04
First Time Setup

EdithLansens:
- Removed some trailing spaces
- Updated screenshots on keeping fork up to date
- Added subtitle 'First Time Setup' + info + subtitle 'Generating a test build in Visual Studio Code
- Added note about localhost


MariekeGO:
@PedroDebevere, could you take a look at this new first-time setup section and let us know if any changes are still needed?

CONTRIBUTING.md
```diff
@@ -582,26 +582,24 @@ The Skyline documentation team will then need to merge your pull request, so tha
...
+
     > [!CAUTION]
     > We recommend that you do not use any of the beta versions.
 
 1. Extract `docfx.zip` to a folder of your choice (e.g. `C:\DocFX`).
 
 1. Add the folder (e.g. `C:\DocFX`) to the environment variable **Path** (user variable or system variable).
 
-    On Windows 10 systems, do the following: 
+    On Windows 10 systems, do the following:
```
PedroDebevere, line 602: Suggestion: Add info for Windows 11 instead.

CONTRIBUTING.md
```diff
@@ -626,9 +624,40 @@ To install DocFX on a Microsoft Windows machine:
...
+   1. Start a build by pressing CTR+SHIFT+b.
+
+   1. Repeat the steps above for the file *Code Library.sln* in the `dataminer-docs\src\Base Class Library\` subfolder of the repository.
+
+1. Open Visual Studio Code and go to *Terminal* > *New Terminal*.
+
+1. In the terminal, enter `docfx metadata -f`.
+
+   If everything is OK, you will get no warnings.
+
```
PedroDebevere, line 659: Note that the need for building the solution is due to the change we had to make in the DocFX file regarding how the API docs is generated. Now it is generating the metadata from the DLL/XML. Previously, it was via the source code, which does not require a build (The long term goal is to be able to again generate it via the source code when that issue is fixed). Note that this also means that whenever there was an update to the code/documentation in the solutions, these solutions will need to be rebuilt again.

MariekeGO: I've made a couple of changes based on your feedback. Could you take another look?

CONTRIBUTING.md
```diff
@@ -626,9 +624,43 @@ To install DocFX on a Microsoft Windows machine:
...
+Before you make your first test build, we recommend the first-time setup detailed below. If you do not do this, you will still be able to make a test build, but there may be many warnings in the terminal, and some pages may not be included in the build.
+
+1. Check if your fork is up to date with the latest changes from the main dataminer-docs repository (See [Make sure your fork is up to date](#make-sure-your-fork-is-up-to-date).
+
+1. Make sure you have Visual Studio 2017 or higher installed.
+
+1. Install the .NET framework SDK Developer Pack. To do so, go to <https://dotnet.microsoft.com/en-us/download/visual-studio-sdks> and choose the .NET Framework 4.8 Developer Pack.
+
+1. Download and install [Build Tools](https://aka.ms/vs/17/release/vs_BuildTools.exe).
```
PedroDebevere, line 637: The Build tools are not required if you have Visual Studio installed. (In fact, DocFX states that is requires VS (so I assume using Build tools instead of VS would not work based on their documentation: https://dotnet.github.io/docfx/tutorial/docfx_getting_started.html). It also states VS2019 (or above) is required while here VS2017 is mentioned.

PedroDebevere (commented):
See remark about MSBuild tools/VS. Other changes are OK.

PR #591, opened 2022-10-04
Update Standalone_Elastic_Backup_Tool.md

AlexCO1:
Please reach out to Alex Collantes from Origin squad about changes to the Standalone Elastic Backup Tool. A new, fully backwards compatible, version of the tool that allows the use of an encrypted key instead of a plain text password has been implemented and tested. Also created a separate PasswordEncryptionTool to encrypt a plain text password and output a txt file with the encrypted form (uses DPAPI for user-bound encryption). Need assistance properly documenting these changes and maintaining an updated repository with the latest code.


MariekeGO:
This change should be implemented in the standard tool first. When that is done, we can document it.

PR #598, opened 2022-10-06
Added Linux path

jens-gellynck:
Included linux path


MariekeGO:
@jens-gellynck I've adjusted the layout because the tab layout should only be used for section headers, not for information in the middle of a procedure (otherwise the procedure will not be displayed correctly).

PR #600, opened 2022-10-06
Update The_Bookings_module.md

karlmasschelein:
The color of a booking is determined by booking property "Visualbackground". A booking timeline will use that color on a booking item by default. A booking listview can show this color by adding the property column "Visualbackground" and choose column type  'Color' to show the color instead of  a color textrepresentation.


MariekeGO:
@karlmasschelein I've made some more changes. Could you check if everything is still correct?

karlmasschelein (commented):
Looks fine to me

MariekeGO (approved):

PR #651, opened 2022-10-24
Added info for Failover Pairs and Cloud Connection

baptistepa:
Added a page that explains considerations to be taken into account when connecting systems that have failover pairs.


MariekeGO:
@baptistepa Instead of adding a new file, I propose we keep it simple and just add a note about the Failover setup in the cloud connection procedure. OK to merge like this?

baptistepa (commented):
OK for me

MariekeGO (approved):

PR #664, opened 2022-10-27
Service definition property placeholder

HansSolos:
Correction on the explanation of this placeholder. It will only show the value of the configured property.


HansSolos (commented):
Better indeed.

MariekeGO (approved):

PR #674, opened 2022-11-02
Add note about OU certificate field

SeppeDejonckheereSkyline:
Added remarks regarding the validation of the OU field and and about using special characters in that field


SeppeDejonckheereSkyline:
To be reviewed by @jens-gellynck

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Database_security/Security_Cassandra_general/Security_Cassandra_TLS.md
```diff
@@ -42,17 +42,20 @@ To generate the certificates, you will need two tools: *openssl* and the *Java k
    ```
 
    > [!NOTE]
-   > Make sure the OU is set to the **name** of your Cassandra cluster, typically **DMS**. You can find this *cluster_name* tag in the *cassandra.yaml* file.
+   > The **OU** is only validated when **internode encryption** is turned on in the *server_encryption_options*. Make sure it matches the *cluster_name* **exactly** or Cassandra will fail to start. You can find the *cluster_name* in the *cassandra.yaml* config file.
+   >
+   >We also recommend not to use non-*ASCII* characters and to avoid special characters in your cluster name. The Cassandra documentation is lacking on this front but we noticed Cassandra failing to start when the *cluster_name* contained certain special/non-*ASCII* characters.
```
jens-gellynck, line 47: It is recommended to only use ASCII characters in your Cassandra cluster name.

jens-gellynck (commented):
Small rephrasing, I typically also avoid stuff like "we recommend", sounds a bit more procedural. Also the thing about 1. 2. 3., it's best to always use 1.

jens-gellynck:
Tagging Cassandra chamption: @SkylineCookieMonster

MariekeGO:
> Small rephrasing, I typically also avoid stuff like "we recommend", sounds a bit more procedural. 

In fact, "we recommend" is currently preferred over "it is recommended", because this matches the Microsoft style guide: [https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/recommend](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/recommend). However, I definitely agree with the rest of your suggestion to rephrase the sentence.

jens-gellynck:
> > Small rephrasing, I typically also avoid stuff like "we recommend", sounds a bit more procedural.
> 
> In fact, "we recommend" is currently preferred over "it is recommended", because this matches the Microsoft style guide: https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/recommend. However, I definitely agree with the rest of your suggestion to rephrase the sentence.

That good to know!

MariekeGO (approved):

PR #710, opened 2022-11-15
List view update

MariekeGO:


karlmasschelein (changes requested):
Added changes look fine to me.
As the columntype 'Custom icon' seems to got broken and is not currently used by customers, I created task [Cube listview column support for custom icons broken](https://collaboration.dataminer.services/task/198216) to restore this functionality. Maybe we should omit this information temporarily.

MariekeGO:
> Added changes look fine to me. As the columntype 'Custom icon' seems to got broken and is not currently used by customers, I created task [Cube listview column support for custom icons broken](https://collaboration.dataminer.services/task/198216) to restore this functionality. Maybe we should omit this information temporarily.

Since it's available in the UI, I'll keep the option in the documentation, but I'll mention that it shouldn't be used at the moment.

karlmasschelein (commented):
Seems odd ? 'This column type can currently not be used.'
This column type can't currently be used.

MariekeGO:
> Seems odd ? 'This column type can currently not be used.' This column type can't currently be used.

I'll adjust this. However, note that contractions (e.g. can't) should not be used in the documentation. See [https://docs.dataminer.services/CONTRIBUTING.html#avoid-contractions](https://docs.dataminer.services/CONTRIBUTING.html#avoid-contractions).

PR #717, opened 2022-11-16
KI: SLLog issue when large alarm tree is closed

MariekeGO:


user-guide/Troubleshooting/Known_issues/KI_SLLog_issue_when_large_alarm_tree_is_closed.md
```diff
@@ -0,0 +1,71 @@
...
+            {     
+                  "RootAlarmID": "30274729"
+            } 
+      }
+}
+```
+
+The prefix before *activealarms* (*dms-* in the examples above) can be different depending on your configuration.
+
+The root alarm ID for the alarm tree can be found in the SLSearch.txt logfile. For example:
```
wdm-slc, line 56: This part (being present in the log) is only true when the RN has already been installed. Before the fix, SLLog would crash trying to add this logging to the logfile

I also learned that even after the SLLog crash fix, [RN34827](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm.aspx?ID=34827) will prevent this line from being logged at all, as it is observed as a "Known Error", so the information is not that helpful at the moment...

user-guide/Troubleshooting/Known_issues/KI_SLLog_issue_when_large_alarm_tree_is_closed.md
```diff
@@ -0,0 +1,71 @@
...
+
+When an alarm tree is closed and DataMiner tries to move it from the *activealarms* Elasticsearch index to the *alarms* index, a tree gets inserted in one bulk request. For trees that are very large (e.g. lengthy history, or a lot of properties or base alarms), the query could become too big and get refused by Elastic. The query was then logged via SLLog, which triggered a problem in that process because of the size of the log statement.
+
+## Fix
+
+Upgrade to DataMiner 10.2.0 [CU10] or 10.3.1.
+
+## Workaround
+
+Depending on the exact cause, a workaround can be to delete an excessively large alarm tree from the *activealarms* index. This can be done with a query in Kibana or Elasticvue, but the exact query depends on the DataMiner version and on the configuration of the system.
```
wdm-slc, line 22: might be good to have links to what Kibana or ElasticVue is?

user-guide/Troubleshooting/Known_issues/KI_SLLog_issue_when_large_alarm_tree_is_closed.md
MariekeGO: In that case, perhaps it would be better if I simply left out this workaround, as it probably won't be very helpful?

wdm-slc: Yes
The main reason for me to give that info was that with the SLLog crash fix, you will still have data building up in the activealarms table. A [follow task](https://collaboration.dataminer.services/task/196930) is being worked on at the moment.
Referring to the log and providing cleaning instructions was mainly for customers that installed the hotfix.
The delete queries for elastic are still useful for customers encountering the crash, before they have access to a hotfix. We tell them the ID to remove in such cases, or the dump analyzer can extract it as well

MariekeGO: I've made some changes. Could you take a look?

wdm-slc (approved):
looks good to me

PR #721, opened 2022-11-16
RN34432

EdithLansens:


EdithLansens:
@kdAtSkyline I have added the light bulb feature to the documentation. Could you do a final check of the changed file? Thanks in advance!

PR #728, opened 2022-11-18
IAS Button styles

lander-vanhaverbeke-skyline:
Adding documentation for IAS button styles (RN34904), and improving documentation on IAS text styles.

### DIFF
--- a/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
+++ b/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
@@ -599,8 +599,8 @@ namespace Skyline.DataMiner.Automation
 		/// <value>The style of the dialog box item.</value>
 		/// <remarks>
 		/// <para>It is possible to set a style on some dialog box items.</para>
-		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, ...).<br/>
-		/// The Style class is available from DataMiner 10.3.1 (feature release) and 10.4.0 (main release) onwards. On older DataMiner versions you can use the StaticText styles 'Title1', 'Title2' and 'Title3'.</para>
+		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, etc.).<br/>
+		/// The Style class is available from DataMiner 10.3.1/10.4.0 onwards. For older DataMiner versions, you can use the StaticText styles 'Title1', 'Title2', and 'Title3'.</para>
 		/// </remarks>
 		/// <example>
 		/// <code>
--- a/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
+++ b/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
@@ -599,7 +599,8 @@ namespace Skyline.DataMiner.Automation
 		/// <value>The style of the dialog box item.</value>
 		/// <remarks>
 		/// <para>It is possible to set a style on some dialog box items.</para>
-		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, ...).</para>
+		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, ...).<br/>
+		/// The Style class is available from DataMiner 10.3.1 onwards. On older DataMiner versions you can use the StaticText styles 'Title1', 'Title2' and 'Title3'.</para>
 		/// </remarks>
 		/// <example>
 		/// <code>
--- a/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
+++ b/src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
@@ -600,7 +600,7 @@ namespace Skyline.DataMiner.Automation
 		/// <remarks>
 		/// <para>It is possible to set a style on some dialog box items.</para>
 		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, ...).<br/>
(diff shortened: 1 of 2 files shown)


VanRobaysWouter (approved):

src/DataMiner/SLManagedAutomation/UIBuilder.cs
```diff
@@ -280,6 +280,31 @@ public class UIBuilder
...
 		public UIBuilder AppendButton(string destVar, string displayText) { return null; }
 
+
+		/// <summary>
+		/// Adds a button to this dialog box with the specified destination variable name, button text and style.
+		/// The supported button styles can be accessed through const strings on the Style.Button class.
+		/// </summary>
+		/// <param name="destVar">The name of the destination variable.</param>
+		/// <param name="displayText">The button text.</param>
+		/// <param name="style">The button style (see Style.Button for supported styled).</param>
```
Pieter-JanV, line 290:
```suggestion
		/// <param name="style">The button style (see Style.Button for supported styles).</param>
```

src/DataMiner/SLManagedAutomation/UIBuilder.cs
```diff
@@ -280,6 +280,31 @@ public class UIBuilder
...
+		/// <returns>This <see cref="UIBuilder"/> instance.</returns>
+		/// <remarks>
+		/// <para>When the button is pressed, the destination variable is filled with its own name.</para>
+		/// <para>This is a convenience method that will create a new instance of <see cref="UIBlockDefinition"/> of type Button and sets <see cref="RequireResponse"/> to <c>true</c>.</para>
+		/// </remarks>
+		/// <example>
+		/// <code>
+		/// UIBuilder uibDialogBox1 = new UIBuilder();
```
Pieter-JanV, line 298: The example doesn't compile. Here we use "uibDialogBox1", but in the next lines, we use uiBuilder. We should make it consistent. Note: this is the case as well for the other AppendButton example

src/DataMiner/SLManagedAutomation/UIBuilder.cs
Pieter-JanV: As an extra comment, should we make a separate section/page for the available styles. Now they can't be found in the Help (i.e. that you have None, Title1, Title2 and Title3 for the text styles and None and CallToAction for the button styles)

Pieter-JanV (changes requested):

src/DataMiner/SLManagedAutomation/UIBuilder.cs
```diff
@@ -249,11 +249,11 @@ public class UIBuilder
 		/// <returns>This <see cref="UIBuilder"/> instance.</returns>
 		/// <example>
 		/// <code>
-		/// UIBuilder uibDialogBox1 = new UIBuilder();
+		/// UIBuilder uiBuilder = new UIBuilder();
 		/// // ...
 		/// UIBlockDefinition blockTextBox = new UIBlockDefinition();
 		/// // ...
-		/// uibDialogBox1.AppendBlock(blockTextBox);
+		/// uiBuilder.AppendBlock(blockTextBox);
```
VanRobaysWouter, line 256: We should probably use the name uibDialogBox1 or something starting with uib. According the the docs the intellisense in cube is only triggered if the name starts with uib. I checked it in cube and uiBuilder does not trigger the intellisense

src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
```diff
@@ -594,23 +594,19 @@ public class UIBlockDefinition
...
-		/// <para>It is possible to add one of three title styles to a text block:</para>
-		/// <list type="bullet">
-		/// <item><description>“Title1” (highest level, all lower case)</description></item>
-		/// <item><description>“Title2” (medium level, all upper case)</description></item>
-		/// <item><description>“Title3” (lowest level, all upper case)</description></item>
-		/// </list>
-		/// <para>If you want to set the style of a text block to regular text, leave the style property empty.</para>
+		/// <para>It is possible to set a style on some dialog box items.</para>
+		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, ...).<br/>
+		/// The Style class is available from DataMiner 10.3.1 onwards. On older DataMiner versions you can use the StaticText styles 'Title1', 'Title2' and 'Title3'.</para>
```
VanRobaysWouter, line 603: Also add the main release where the feature will be available in, same for the AppendButton method

lander-vanhaverbeke-skyline: This is a new feature, so it will only be merged into RC, so no need to split it up into feature and main release. 
All DataMiner versions equal to or bigger than 10.3.1 will have this functionality.

src/DataMiner/SLManagedAutomation/UIBuilder.cs
lander-vanhaverbeke-skyline: You're right, I'll change them to uibDialogBox.

src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
MariekeGO: The main release version should be mentioned too, just for the sake of clarity. Otherwise, people on the main release track won't know whether it's available to them in 10.3.0, or if they need to wait for 10.4.0.

lander-vanhaverbeke-skyline: Ok, added.

MariekeGO (approved):

PR #729, opened 2022-11-18
Update Protocol.Params.Param.Type.md

AiltonLuz:
Extend the information regarding read bit parameters. NbrOfBits should be multiple of 8, otherwise, wrong values could be read from the group.


MariekeGO:
@PedroDebevere, could you confirm if this is OK to merge?

PedroDebevere:
@MariekeGO , this does not need to be a multiple of 8, there are implementations that set the NbrOfBits to values other than multiples of 8 which work.

PR #734, opened 2022-11-22
Naming convention adjustment

santiagoalz:
Due to character limitations of windows, use the new naming convention: SLC.CLib


MariekeGO:
@PedroDebevere, could you check if it is OK to merge this?

PR #740, opened 2022-11-24
Update Connectivity_xml_files_representing_chains.md

KevinVerm:
Info from RN https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm.aspx?ID=10633 added


MariekeGO (changes requested):
As this is not an example of a complete chain, which is what users will expect based on the title of the section where you added it, please move your example under "itemRelation" in the section below that. I think it might be best to add it in a note there, under the rest of the info about that attribute. Can you also adjust the example so that the closing tags are displayed, and add lines with "..." where you omit the Map tags? This so it's clear that this is only a partial example.

MariekeGO (changes requested):
Could you add a two-space indent in front of the new section, so that it's clear that it belongs under the "itemRelation" item?

MariekeGO (approved):

PR #766, opened 2022-11-30
Added Visio color placeholder

MariekeGO:
@ToonCasteele, could you check if this is OK to merge?


ToonCasteele:
Looks good to me. Possibly in the future we might have more things to retrieve the color from than the severity. But at that point we can just change the documentation :)

PR #782, opened 2022-12-05
RN34836

EdithLansens:
@MichielVerbruggeSky, could you verify whether RN 34836 was interpreted and added to the documentation correctly? Thanks in advance!


MichielVerbruggeSky:
Hi @EdithLansens, the documentation looks correct. You might need to add it to the monitoring app as well. Due to my change, string trending will also be supported on the trending pages in the monitoring application.

EdithLansens:
Thanks for the check, @MichielVerbruggeSky! I have now added a note in the section on the Monitoring app as well.

MariekeGO (approved):

PR #789, opened 2022-12-06
Creating visibility on Consistency Level, Replication Factor, data re…

SLC-Laurens-Vercruysse:
As discussed Monday 05/12 with Marieke. 
Added page Data replication and consistency in the Database/ Cassandra section. 
Also added information on consistency level, replication factor in DataMiner as well as consistency level configuration in db.xml page.


SLC-Laurens-Vercruysse (commented):
Looks excellent! It has better emphasis on the important bits now and the references are more clear. Thank you.

MariekeGO (approved):

PR #805, opened 2022-12-08
Update Overview_DCP.md

bfsantos:


MariekeGO (changes requested):
There also is a subtitle "DataMiner Cloud Services" on this page (line 60). Could you look into changing either the main title or that subtitle? Having duplicate titles on a page is not recommended, as that makes it impossible to create cross-references to the second instance of the title.
Possibly the structure of this page on the whole may need to be improved a bit, as it now appears to have a subsection dedicated to the main topic as the entire page, which doesn't make all that much sense.

MariekeGO:
Closed for now, as the title change is problematic. Changing the structure of the page can be done in another PR.

PR #835, opened 2022-12-19
Added note on how to Remove the Tenant ID

joaobarraca:
Added extra information on how to remove the tenant ID


MariekeGO (commented):
I've made a couple of changes. Could you check if this is still OK? Also, could you check whether there is no overlap between the second and third item of the note now? It's not entirely clear to me what the difference is between removing permissions and removing consent, or why a user would choose to do the one but not the other.

joaobarraca:
> I've made a couple of changes. Could you check if this is still OK? Also, could you check whether there is no overlap between the second and third item of the note now? It's not entirely clear to me what the difference is between removing permissions and removing consent, or why a user would choose to do the one but not the other.

there is no overlapping. if the user would remove the consent, the tenant id would still be linked to the organization. 
this is an added feature where the user can remove the linking of the tenant id to the organization

MariekeGO (approved):

PR #860, opened 2023-01-03
Update TCP-IP_socket_message_reference.md

KevinVerm:


MariekeGO (changes requested):
This file still contained a few conversion issues, so I've corrected these. I've also moved the note you added to the command, since it's most relevant there.

Could you still add an example that includes the domain name? That could make things much clearer for the user.

MariekeGO (approved):

PR #895, opened 2023-01-11
Clarify the TO address in ConnectionString configuration in a context…

wdm-slc:
… with Failover agents


MariekeGO:
@wdm-slc, this change conflicted with a change from an earlier PR that replaced part of this procedure with a link because it overlapped with another procedure. I've resolved the conflict by moving your addition to the linked procedure. OK to merge like this?

wdm-slc:
ok for me

PR #904, opened 2023-01-13
Update Parameter info to avoid confusion

ArneMaes0:
columnValues -> values since that is the one they use in the example.


MariekeGO:
@PedroDebevere, could you check if this is OK to merge?

PedroDebevere:
change is OK. In the second example that change should also be applied (after the paragraph that starts with "In case the column data make use of the protocol.Clear").

MariekeGO (approved):

PR #912, opened 2023-01-17
Update Configuring_external_authentication_via_an_identity_provider_u…

MichielMN:
…sing_SAML.md

Quick phrasing edit to avoid confusion

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
@@ -279,7 +279,7 @@ DataMiner supports Okta as identity provider as from version 10.1.11. Use Okta's
 > [!IMPORTANT]
 >
 > - Prior to DataMiner 10.3.0/10.3.2, it may not be possible to log in using Okta because of a software issue. We strongly recommend that you upgrade to DataMiner 10.3.0 or 10.3.2 to use this feature.
-> - DataMiner only supports the use of automatic user creation with Okta.
+> - When using Okta, automatic user creation must be enabled, it is not possible to import users from Okta.
 
 1. Launch the App Integration Wizard
 


user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -242,11 +242,12 @@ From DataMiner 10.2.0/10.1.12 onwards, users authenticated by Azure AD using SAM
      ...
    </DataMiner>
    ```
+1. User groups have to exist in DataMiner for users to be added to them (both for Groups claims set to true and false). See [Adding a user group] (Adding_a_user_group) for more info on how to do this
```
MariekeGO, line 245: The step to add user groups is already the second step of this procedure, which makes this a bit weird. When exactly should the user add the user groups? Possibly this should be rewritten as a note instead?
Also, make sure you leave an empty line above this line, and that you format cross-references correctly. In this case, it should be `[Adding a user group](xref:Adding_a_user_group)`.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -242,11 +242,12 @@ From DataMiner 10.2.0/10.1.12 onwards, users authenticated by Azure AD using SAM
...
 
    > [!NOTE]
    > If you set the *claims* attribute of the *Groups* element to "false", no claims will be used to add users to groups. In this case:
    >
-   > - The name of the group as specified in Cube will be used instead.
+   > - Instead of a claim for user groups replace [group claim name] with a security group that exists in DataMiner as described above.
```
MariekeGO, line 250: Ideally, this note should be under the same step as the one containing that "[group claim name]", so that it's more clear what this is about.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -279,7 +280,9 @@ DataMiner supports Okta as identity provider as from version 10.1.11. Use Okta's
 > [!IMPORTANT]
 >
 > - Prior to DataMiner 10.3.0/10.3.2, it may not be possible to log in using Okta because of a software issue. We strongly recommend that you upgrade to DataMiner 10.3.0 or 10.3.2 to use this feature.
-> - DataMiner only supports the use of automatic user creation with Okta.
+> - When Okta is used, automatic user creation must be enabled. It is currently not possible to import users and groups from Okta.
+> - Alternatively you can add local or domain users, see "User directories" and "Local users" under [User Management] (xref:User_management).
+>   After that you can point Okta to authenticate these users by following the below guide but ommitting the AutomaticUserCreation tag from Dataminer.xml
```
MariekeGO, line 285: Can these three lines be combined? I'd suggest to combine them as follows:
`> - When Okta is used, automatic user creation must be enabled. It is currently not possible to import users and groups from Okta. Alternatively, you can add local or domain users. See [User directories](xref:User_management#user-directories) and [Local users](xref:User_management#local-users). After that, you can have Okta authenticate these users by following the guide below, except that you omit the *AutomaticUserCreation* tag in *DataMiner.xml*.`

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -377,11 +380,12 @@ DataMiner supports Okta as identity provider as from version 10.1.11. Use Okta's
      ...
    </DataMiner>
    ```
+1. User groups have to exist in DataMiner for users to be added to them (both for Groups claims set to true and false). See [Adding a user group] (Adding_a_user_group) for more info on how to do this
```
MariekeGO, line 383: This line has similar issues as line 245: duplicate step in the procedure (creating groups is already mentioned on line 349), missing empty line, incorrect cross-reference. The change below this also has the same issue as line 250.

MariekeGO (changes requested):

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
MichielMN: Moved this line to the second step to make the original more clear

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
MichielMN: This refers specifically to what to do when not using a claim for group names so I think this is in the right spot
can't really put it next to the group claim name step as that would be more confusing and contradicting I think.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
MichielMN: done

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
MichielMN: Moved the line to the correct spot, fixed cross-reference, not really sure what you mean with empty line though

MariekeGO:
I've adjusted the procedure a bit, to avoid a situation where a step refers to something that is only explained later in the procedure (i.e. Groups claims), which can create confusion and isn't very user-friendly.

I've also added the empty line between the code block in line 245 and the note below it. It's a good practice to always place an empty line between two paragraphs, as otherwise you risk that they are not displayed correctly.

@MichielMN, can you take a look?

PR #1015, opened 2023-02-07
dataminer.services component added

EdithLansens:


MariekeGO:
@Glenndhaene, could you confirm if this is OK? As a [page related to CCA Support Services](https://docs.dataminer.services/dataminer-overview/Support%20Services/CCA_Support_Services.html) was added, it seemed better to us to also mention it under the dataminer.services features.

MariekeGO:
Rendered obsolete by #1029

PR #1024, opened 2023-02-08
Added page to upgrade dashboards when installing package.

sergiosuarezsky:


user-guide/Standard_Apps/EPM/EPM_I-DOCSIS/I-DOCSIS_upgrade.md
```diff
@@ -0,0 +1,14 @@
+---
+uid: I-DOCSIS_upgrade
+---
+
+# EPM I-DOCSIS upgrade
+
+To upgrade the Dashboards to work property:
```
MariekeGO, line 7: Could you explain this a little more: **Where** should the user replace these values? **Why** do these need to be replaced?  Are any **other actions** necessary when users upgrade? Is this necessary for **every upgrade, or only for a specific upgrade** (and which one)?
Also, for the sentences below, could you make it more clear what exactly it is that is being replaced? The way they are phrased now is quite confusing.

PR #1037, opened 2023-02-10
WebSocket Timeout Section

FlavioJA:
Added information related to the timeout in the WebSocket communication in Connectors.

### DIFF
--- a/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
@@ -96,9 +96,7 @@ These are possible reasons why the connection might be terminated:
 - Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition. Note that if the TCP retranmission time exceeds the WebSocket timeout time, the WebSocket might be closed depending on the implementation, since there was no activity for a period greater than the configured WebSocket timeout time.
 
 > [!NOTE]
->
-> - This behavior may vary depending on the other connections of the protocol.
-> - The timeout behavior of the element may vary depending on other connections configured on protocol level.
+> The timeout behavior of the element may vary depending on other connections configured on protocol level.
 
 ## Binary vs. Text Data Frames
 
--- a/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
@@ -93,7 +93,7 @@ These are possible reasons why the connection might be terminated:
 
 - Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
 
-- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition.
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition. Note that if the TCP retranmission time exceeds the WebSocket timeout time, the WebSocket might be closed depending on the implementation, since there was no activity for a period greater than the configured WebSocket timeout time.
 
 > [!NOTE]
 >
--- a/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
@@ -93,7 +93,7 @@ These are possible reasons why the connection might be terminated:
 
 - Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
 
-- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition. Note that if the TCP retranmission time exceeds the WebSocket timeout time, the WebSocket might be closed depending on the implementation, since there was no activity for a period greater than the configured WebSocket timeout time.
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition. Note that if the TCP retransmission time exceeds the WebSocket timeout time, depending on the implementation, the WebSocket might be closed, since there is no activity for a period greater than the configured WebSocket timeout time.
 
 > [!NOTE]
 > The timeout behavior of the element may vary depending on other connections configured on protocol level.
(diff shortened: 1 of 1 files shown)


MariekeGO:
@mmphs, could you take a look at this and let me know if it's OK?

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -79,23 +79,26 @@ When the WebSocket server closes the connection for some reason, the element goe
...
-Note that this behavior might vary depending on the other connections of the Connector. 
+- Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
 
-Note that the timeout behavior of the element might vary according to other connections configured at Connector level.
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition.
+
+> [!NOTE]
+>
+> - This behavior may vary depending on the other connections of the protocol.
+> - The timeout behavior of the element may vary depending on other connections configured on protocol level.
```
mmphs, line 101: What is **This behavior** referring to in the first paragraph? I thought it was the timeout behavior, but is is then repeated in the second paragraph.

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -79,23 +79,26 @@ When the WebSocket server closes the connection for some reason, the element goe
...
 
-* Policy Violation: The WebSocket connection is closed by the server due to a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
+- Ping timeout: The WebSocket connection has timed out because of a lack of activity. This can occur if the client or server does not receive a ping or pong message within a specified time frame.
 
-* Network Error: The WebSocket connection is closed due to a network error, such as a dropped packet or a network partition.
-Note that this behavior might vary depending on the other connections of the Connector. 
+- Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
 
-Note that the timeout behavior of the element might vary according to other connections configured at Connector level.
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition.
```
mmphs, line 96: A dropped packet will typically be retransmitted and not cause the closure of a connection.

mmphs (changes requested):
Technically correct, but isn't it a bit too much theoretical and broad?
Most of below concepts apply to communication on the TCP layer (besides ping / pong which is specific for Websockets) so you could write something similar in the Timeout section of Serial, Smart-Serial, SSH, SSL/TLS.

I'm wondering if we should not bundle it somewhere and document it in connection-oriented vs connectionless?

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
FlavioJA: Indeed, however the idea here was to indicate that in case of a packet drop, if the TCP retransmission of the packet exceeds the maximum timeout period specified by the WebSocket implementation, the connection may be closed by the server or client. I believe that this can be included as additional info.

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
FlavioJA: Indeed, that's duplicated info, meaning that only the second paragraph should be maintained.

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -77,6 +77,29 @@ For example, suppose you need to connect to ws://10.4.2.8:4601/x-nmos/events, th
...
+- Ping timeout: The WebSocket connection has timed out because of a lack of activity. This can occur if the client or server does not receive a ping or pong message within a specified time frame.
+
+- Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
+
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition.
+
+> [!NOTE]
+>
+> - This behavior may vary depending on the other connections of the protocol.
+> - The timeout behavior of the element may vary depending on other connections configured on protocol level.
```
mmphs, line 101:
```suggestion
> [!NOTE]
> The timeout behavior of the element may vary depending on other connections configured on protocol level.
```

develop/devguide/Connector/ConnectionsWebSocketsImplementation.md
```diff
@@ -77,6 +77,29 @@ For example, suppose you need to connect to ws://10.4.2.8:4601/x-nmos/events, th
...
+
+- Abnormal closure: The WebSocket connection is closed unexpectedly, because of an error or exception. This can be caused by issues such as network connectivity problems, server crashes, or protocol violations.
+
+- Endpoint shutdown: The endpoint (client or server) that initiated the WebSocket connection has shut down, which may be due to a process termination or system reboot.
+
+- Ping timeout: The WebSocket connection has timed out because of a lack of activity. This can occur if the client or server does not receive a ping or pong message within a specified time frame.
+
+- Policy violation: The WebSocket connection is closed by the server because of a violation of security policies, such as a cross-site scripting attack or unauthorized access attempt.
+
+- Network error: The WebSocket connection is closed because of a network error, such as a dropped packet or a network partition.
```
mmphs, line 96:
```suggestion
- Network error: The WebSocket connection is closed because of a network error, e.g. a network partition or a final dropped packet after retransmission.
```

mmphs (changes requested):

FlavioJA:
I believe that it would be great if we could include more information regarding connection-oriented vs connectionless in our documentation as an overall. However, for this particular case I consider that it's important to indicate what can cause a WebSocket to be closed since many DataMiner users are purely operators that might only consider the TCP socket timeout, ending up forgetting about the timeout time of the WebSocket itself. Therefore, I believe that this info can help in further troubleshooting.

mmphs:
@MariekeGO, looks good to me.

MariekeGO (approved):

PR #1050, opened 2023-02-14
Update Failover setup - DCP204995

SLC-Laurens-Vercruysse:
Following a maintenance window, we found that the customer hadn't anticipated dealing with the HTTPS configuration. Therefore added this bit.


MariekeGO:
@SLC-Laurens-Vercruysse I've made some changes to your pull request. Could you check and let me know if this is OK to merge?

MariekeGO (approved):

PR #1073, opened 2023-02-16
RN 35565 + 35057

EdithLansens:


GillesBara (approved):

user-guide/Advanced_Modules/Dashboards_app/Using_dashboard_feeds/Configuring_GQI_feeds.md
```diff
@@ -412,6 +412,11 @@ Joins two tables together. When you select this option, in the *Type* drop-down
 
 The *Inner* type of join only includes rows if they match the condition. *Left* displays all rows from the first table (i.e. the table on the left) and only the matching rows from the other table. *Right* does the opposite. *Outer* displays first the non-matching rows from the left table, then the matching rows from both tables, then the non-matching rows from the right table.
 
+From DataMiner 10.3.3/10.4.0 onwards, the *Row by row* option allows you to first execute the first table, and then execute the other table for each row that matches the join condition. To do so, first add the `showAdvancedSettings=true` option to the dashboard URL. If the *Row by row* option is not enabled, the join will execute both tables once, and directly combine their results. <!-- RN 35565 + 35057 -->
```
MariekeGO, line 415: This "execute the table" is a bit weird. Would it make sense to change this to "execute the query for the table"?

Also, the "To do so," introduction for the second sentence doesn't really work that well. I'd propose to change that like this:
```suggestion
From DataMiner 10.3.3/10.4.0 onwards, the *Row by row* option allows you to first execute the first table, and then execute the other table for each row that matches the join condition. This option is only available if you add the `showAdvancedSettings=true` option to the dashboard URL. If the *Row by row* option is not enabled, the join will execute both tables once and directly combine their results. <!-- RN 35565 + 35057 -->
```

EdithLansens:
Is this still OK to merge after the small edits we've made? @GillesBara

GillesBara (approved):

PR #1074, opened 2023-02-16
Update CCA_Support_Services.md

BartReynaert:
Just using the word 'Agents' might be confusing - 'DataMiner Agents' is unambiguous. Also, starting the sentence with 'If' is again confusing, since the 'condition' is already fulfilled at the beginning of this article.


MariekeGO (approved):

BartReynaert:
Note that the isolated word 'Agents' is used twice in this paragraph. Hence, accordingly to be replaced twice. :)

PR #1078, opened 2023-02-17
RN 35443

EdithLansens:


MariekeGO:
I see that the ModelHost DxM isn't mentioned here yet: https://docs.dataminer.services/user-guide/Reference/DataMiner_Extension_Modules.html
Could you add it to that list of available DxMs, with a short explanation of what it is?

EdithLansens:
> I see that the ModelHost DxM isn't mentioned here yet: https://docs.dataminer.services/user-guide/Reference/DataMiner_Extension_Modules.html Could you add it to that list of available DxMs, with a short explanation of what it is?

I have made a new pull request to include this: https://github.com/SkylineCommunications/dataminer-docs/pull/1080

PR #1085, opened 2023-02-21
RN 35654

EdithLansens:


release-notes/Cloud/DCP_change_log.md
```diff
@@ -6,6 +6,10 @@ uid: DCP_change_log
 
 The dataminer.services platform gets updated continuously. This change log can help you trace when specific features and changes have become available.
 
+#### 16 February 2013 - Fix - CloudGateway 2.10.3 - Share not created because of mismatch in local culture on hosting server [ID_35654]
+
+Prior to CloudGateway version 2.10.3, a share could not be created because of a mismatch in the local culture on the hosting server.
```
JordyGit, line 11: would say 'if there was a mismatch' instead of 'because of'
It only happened at one user so far we know, other users have always been fine

JordyGit (approved):

PR #1091, opened 2023-02-22
UI blocks - Add IAS toolkit TIP

HansSolos:
Add recommendation of IAS toolkit when creaing UI blocks


EdithLansens:
@HansSolos Would this tip not be better placed on this page if we want to add a disclaimer? https://docs.dataminer.services/develop/api/types/Skyline.DataMiner.Automation.UIBlockType.html

HansSolos:
> 

The TIP can be placed in both pages. As long as it is clear that there is a high-level implementation of UI blocks.

EdithLansens:
@HansSolos I moved up the tip so that it stands out even more.

EdithLansens (approved):

HansSolos (commented):
Approved

PR #1092, opened 2023-02-22
AddedUpgradeCassandraOnLinux

KevinVerm:
Added information on how to upgrade Cassandra on Linux


KevinVerm:
Draft was presented to Michiel Saelen, I didn't found him on GitHub, I'll send this PR link in Teams.

KevinVerm:
Added @SkylineCookieMonster

SkylineCookieMonster:
@KevinVerm Can you add that it is common practice in Cassandra to do rolling upgrades which means an upgrade can be done without downtime if your settings allow it? Rolling upgrades means basically upgrade one node bring it up again and then move to the next one. When going from one major range to another E.g. from 3.x to 4.x that it is best to first go to the latest version in the current range before going to the new range?

KevinVerm:
> @KevinVerm Can you add that it is common practice in Cassandra to do rolling upgrades which means an upgrade can be done without downtime if your settings allow it? Rolling upgrades means basically upgrade one node bring it up again and then move to the next one. When going from one major range to another E.g. from 3.x to 4.x that it is best to first go to the latest version in the current range before going to the new range?

Uploaded new commit with above remarks

spetebroodt (approved):

PR #1099, opened 2023-02-23
Update GRPCConnection type example

KarelLodefier:
Previous example was confusing, as "polling" and "zip" aren't valid options for the GRPCConnection type. Refer https://community.dataminer.services/question/grpc-connection-polling-value-required-when-configured-on-dms-running-the-latest-v10-3-2/ I expect further improvements will be provided by Bert B (as mentioned in his post).


MariekeGO:
I don't quite understand this change. You say "polling" and "zip" aren't valid options for the GRPCConnection type, but the example you've changed is not related to that type. It's a remoting connection.

PR #1136, opened 2023-03-03
Add remark about the required parameters for PA PLS

slc-peter-vanpoucke:
Add remark about the required parameters on a script task profile load script.

Although the screenshot show those parameters and the PLS template should be copied, it might be unclear that they're needed.

### DIFF
--- a/user-guide/Standard_Apps/PA/Getting_Started_with_PA/Creating_Activities/PA_Creating_script_tasks.md
+++ b/user-guide/Standard_Apps/PA/Getting_Started_with_PA/Creating_Activities/PA_Creating_script_tasks.md
@@ -141,8 +141,8 @@ uid: PA_Creating_script_tasks
       > [!NOTE]
       > In the C# block, never try to directly retrieve data from the DOM instance, as this can have unexpected results.
 
-      > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


user-guide/Standard_Apps/PA/Getting_Started_with_PA/Creating_Activities/PA_Creating_script_tasks.md
```diff
@@ -141,6 +141,9 @@ uid: PA_Creating_script_tasks
       > [!NOTE]
       > In the C# block, never try to directly retrieve data from the DOM instance, as this can have unexpected results.
 
+      > [!IMPORTANT]
+      > Check that the script has three parameters: *Info*, *ProcessInfo* and *ProfileInstance*.
```
MariekeGO, line 145: Are these parameters that the user still needs to add, or is this supposed to have happened sometime during the procedure above? If the user needs to add them, it might be better to change the creation of the parameters into a step of its own in the procedure.

slc-peter-vanpoucke: At line 84 it mentions to duplicate a template script, so the parameters should already be there. On occasions where people would transform a regular script it might be interesting to inform they're really required.

MariekeGO (approved):

PR #1146, opened 2023-03-07
RN 35323

MariekeGO:


SDumoulein (commented):
I suggest to use 'rearrange' rather than 'reposition' as that's the wording which is used in the Cube UI as well. Also note that they can be rearranged, it's not something that always happens when using dynamic zooming.

MariekeGO:
> I suggest to use 'rearrange' rather than 'reposition' as that's the wording which is used in the Cube UI as well. Also note that they can be rearranged, it's not something that always happens when using dynamic zooming.

Updated! Could you take a look?

user-guide/Basic_Functionality/Visio/positioning_shapes/Enabling_dynamic_zoom_on_dynamically_positioned_shapes.md
```diff
@@ -4,7 +4,7 @@ uid: Enabling_dynamic_zoom_on_dynamically_positioned_shapes
 
 # Enabling dynamic zoom on dynamically positioned shapes
 
-From DataMiner 9.5.1 onwards, dynamic zooming can be configured when shapes are positioned dynamically. When dynamic zooming is enabled, shapes will not change in size when you zoom in, but will instead be repositioned. Shapes will move away from each other when you zoom in and will move closer together when you zoom out.
+From DataMiner 9.5.1 onwards, dynamic zooming can be configured when shapes are positioned dynamically. When dynamic zooming is enabled, shapes will not change in size when you zoom in, but will instead be rearranged. Shapes will move away from each other when you zoom in and will move closer together when you zoom out.
```
SebastiaanSL, line 7: This one should be kept as 'repositioned'.

SebastiaanSL (changes requested):

PR #1157, opened 2023-03-08
Update About_redundancy_groups.md

edalslc:
Add limitations about SRM and Redundancy groups


MariekeGO:
@edalslc I've added an About page for SRM with this same note about redundancy groups. Could you take a look and let me know if this is OK for you?

edalslc (commented):
ok

MariekeGO (approved):

PR #1184, opened 2023-03-14
GMail removed because GMail not supported anymore

elkegovaert1:


MariekeGO:
@elkegovaert1, I'll merge this pull request to remove this example, but is there an alternative example you could add instead? I'd prefer if there is at least something to illustrate what this configuration looks like.

MariekeGO (approved):

PR #1194, opened 2023-03-15
Update FAQ.md

SimonV-Skyline:


MariekeGO:
Would there be a more user-friendly solution for updating this section than just adding a note at the top? One possibility I see is to update the question itself, so that it's clear that the answer is related to Gerrit only, and another is to rework the answer entirely so that it matches today's common practices. Since you say there are different alternatives that are used today, you could for instance start by explaining the best choices.

SimonV-Skyline:
> Would there be a more user-friendly solution for updating this section than just adding a note at the top? One possibility I see is to update the question itself, so that it's clear that the answer is related to Gerrit only, and another is to rework the answer entirely so that it matches today's common practices. Since you say there are different alternatives that are used today, you could for instance start by explaining the best choices.

Hi Marieke,
Thank you for the suggestions. I did consider those things as well, I was a bit unsure how to proceed tbh. Thing is that the "real" best practice is not yet ready, it will be relatively soon so currently there are different alternatives which I personally find better but not real best practice yet. However, in the meantime, I wanted to (even if a bit quick & dirty) already make it clear that Gerrit is really optional and not really a recommendation as I noticed many colleagues are doing it cause they (wrongly) are under the impression it is the way we recommend to do it.
So yeah, I think my change will make it a bit better and good enough even though not perfect as a temp solution up until we come up with the real new recommendation (which, tbw, will be moving everything to github and using Pull Requests just like for DM Docs)
Feel free to disagree on this of course :)

PR #1197, opened 2023-03-16
RN 33950 corrected

MariekeGO:


slc-peter-vanpoucke:
Looks good, thanks.
The name of the property was NewBookingId initially but was changed in the end.

PR #1217, opened 2023-03-22
Documentation for OpenConfig API

mmphs:
- Extended docfx configuration to scan for dlls
- Exclude Gnmi namespaces from documentation.


PedroDebevere (approved):
moved assembly from src to assemblies folder

PR #1230, opened 2023-03-24
Added note with reason for group-connection

SlcRobinME:
Added note to indicate reasoning on why it's important to add the group-connection attribute when using additional threads.

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Threads.Thread.md
+++ b/develop/schemadoc/Protocol/Protocol.Threads.Thread.md
@@ -60,3 +60,6 @@ To link a group to a thread, set the connection attribute to the connection ID o
   </Content>
 </Group>
 ```
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


develop/schemadoc/Protocol/Protocol.Threads.Thread.md
```diff
@@ -60,3 +60,6 @@ To link a group to a thread, set the connection attribute to the connection ID o
   </Content>
 </Group>
 ```
+
+> [!NOTE]
+> When multiple threads are used, it is important that each poll group has the [connection](xref:Protocol.Groups.Group-connection) attribute defined to indicate which connection/thread to use. This can avoid unexpected behavior in SLPort when the same group is executed on different threads in a similar time frame.
```
PedroDebevere, line 65:
```suggestion
> When multiple threads are used, it is important that each poll group has the [connection](xref:Protocol.Groups.Group-connection) attribute defined to indicate which connection/thread to use. This to avoid unexpected behavior in SLPort when the same group is executed on different threads in a similar time frame.
```

MariekeGO (approved):

PR #1255, opened 2023-03-29
Update PA_Integrating_Gateways.md

GelberMSkyline:
I believe we are missing the DomField definition in sequence 2 of the Exclusive split JSON sample


edalslc (approved):
ok for me

MariekeGO (approved):

PR #1258, opened 2023-03-30
Amazon Keyspaces Credentials Clarification

RobbeDGH:


user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
```diff
@@ -47,6 +47,38 @@ For Amazon Keyspaces, the Starfield certificate must be present on the local Win
 
    ![Add Certificate](~/user-guide/images/aks_add_certificate.png)
 
+## Generating credentials for Amazon Keyspaces
+
+In the Identity and Access Management (IAM) Service, you can generate new credentials that are used to access the Amazon Keyspaces service. Those credentials are then linked to the account that created them.
+
+In order to do this, the right permissions are required.
```
MariekeGO, line 54: Which permissions? This seems a bit too vague to be user-friendly. Or does this refer to the permissions you need to check in step 3 below? If it's the latter, I would leave out this sentence here, because it will likely put the reader on the wrong foot, making them think they need specific permissions before they can start looking at the procedure below.

user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
```diff
@@ -47,6 +47,38 @@ For Amazon Keyspaces, the Starfield certificate must be present on the local Win
...
+
+All of this takes place in the IAM service.
+
+1. Go to *IAM* > *Users*.
```
MariekeGO, line 58: How do you get to IAM? Where should users look to find this?

user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
```diff
@@ -47,6 +47,38 @@ For Amazon Keyspaces, the Starfield certificate must be present on the local Win
...
+
+   ![IAM](~/user-guide/images/Amazon_Keyspaces_IAM.png)
+
+1. Search for the account that should have access.
+
+   ![SearchUser](~/user-guide/images/Amazon_Keyspaces_SearchUser.png)
+
+1. Verify that this account has the permissions to generate Amazon Keyspaces credentials.
```
MariekeGO, line 66: What if the account doesn't have them? The way the procedure is written now, you just check if the permissions are there, and then just continue regardless of the outcome of the check. Something also seems to be missing between this step and the next: shouldn't the user log in with that account? If you assume that they are already logged in with this, that should be made clear from the start.

MariekeGO (changes requested):

user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
RobbeDGH: Removed this line.

user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
RobbeDGH: added a direct link

user-guide/Advanced_Functionality/Databases/Amazon_Keyspaces_Service.md
RobbeDGH: Added instructions on how to add the permission.

Also added a first step to log in on AWS and go to the IAM service. Clicking on this link automatically prompts a login screen.

MariekeGO (approved):

PR #1252, opened 2023-03-29
Add UIBlockType DownloadButton [RN35869] [DCP204668]

HansSolos:
Doucmentation of new UIBlockType 'DownloadButton '.

PR #1259, opened 2023-03-30
Button styles applicable to DownloadButton

lander-vanhaverbeke-skyline:


src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
```diff
@@ -600,6 +600,7 @@ public class UIBlockDefinition
 		/// <remarks>
 		/// <para>It is possible to set a style on some dialog box items.</para>
 		/// <para>The supported styles can be accessed through const strings on the Style class, subdivided per control type (Button, Text, etc.).<br/>
+		/// All Button styles can also be applied to a DownloadButton.<br/>
```
MariekeGO, line 603: Is the casing OK here? I'd only write "DownloadButton" if this refers to a specific object that is spelled that way in the code, but I can't find any mention of "DownloadButton" anywhere. If indeed this does not exist, I'd suggest the following spelling instead:
```suggestion
		/// All button styles can also be applied to a download button.<br/>
```

MariekeGO:
To be merged after #1252

MariekeGO (approved):

PR #1264, opened 2023-03-31
Installing Elasticsearch on linux machine

daudahabbasi:
Added the images and user guide to install Elasticsearch on linux machine.

PR #1265, opened 2023-03-31
Update Elasticsearch_database.md

daudahabbasi:
Link added for installing Elasticsearch on linux machine


MariekeGO:
This link is now added with #1264.

PR #1266, opened 2023-03-31
added modelhost and relationlearning documentation

lustpieter:


MariekeGO:
@lustpieter, I've fully reviewed the PR and made the necessary changes. Could you take a look and check if everything is still OK for you?

user-guide/Basic_Functionality/Trending/Advanced_analytics_trending.md
```diff
@@ -16,6 +14,7 @@ These features are:
 
 - [Working with pattern matching](xref:Working_with_pattern_matching)
 
-## Cloud-Based Advanced Analytics
```
lustpieter: You've removed this again.
Does it make sense to have 'Adding related parameters to a trend graph' in the same list?
Because it isn't provided by SLAnalytics, but by the ModelHost DxM.
This cannot be deactivated via *System Center* > *System settings* > *analytics config* as well.

user-guide/Basic_Functionality/Trending/Advanced_analytics_trending.md
```diff
@@ -4,9 +4,7 @@ uid: Advanced_analytics_trending
 
 # Advanced analytics in trend graphs
 
-## Local Advanced Analytics
```
lustpieter: This has been removed again.
Does it make sense to have the 'Adding related parameters to a trend graph' entry in here that does not adhere to the explanation right before it?
- RL does not use SLAnalytics directly. Instead it uses the ModelHost DxM.
- It cannot be (de)activated via '*System Center* > *System settings* > *analytics config*'

MariekeGO: Adjusted - could you take another look?

MariekeGO (approved):

PR #1282, opened 2023-04-04
Test build procedure updated

MariekeGO:


CONTRIBUTING.md
```diff
@@ -438,14 +438,16 @@ To be able to make a local test build, you need to have DocFX installed. DocFX i
...
-      > - If you only have .NET 7.0 installed, it is possible that you encounter an exception when trying to make a build. See [No instances of MSBuild could be detected](#no-instances-of-msbuild-could-be-detected).
+      - `dotnet restore "src/NuGetPackages"`
+      - `dotnet build "src/NuGetPackages" --configuration Release`
+      - `docfx metadata`
+      - `docfx build`
+      - `docfx serve _site`
 
-   1. Enter `docfx serve _site`.
+      > [!NOTE]
+      > If you make repeated test builds to check changes you have made, it may not be necessary to generate metadata again. In that case, you can already see the results of your changes with the `docfx build` and `docfx serve _site` commands.
```
PedroDebevere, line 450: here you could also mention that the metadata command is used to generate the API docs. If they only are updating markdown files then it is not needed to repeatedly execute docfx metadata

MariekeGO: OK after my changes?

PedroDebevere: Change is ok, the other note about the need to have the .NET SDK is still useful to keep as it is still a requirement

MariekeGO: Updated! Can you take a look?

PedroDebevere (approved):

PR #1292, opened 2023-04-05
Added maps deployment procedures

MarcosMMA:
Added maps deployment procedures for two different scenarios:
1. Systems with one Backend
2. Systems with multiple Backends


MariekeGO (changes requested):
This PR contains links to files on internal websites (svn.skyline.be and sharepoint). Should outside users have access to these files too?

If yes, we will have to make the files available in a different location (e.g. a Dojo download) and link to that instead.

If no, we will have to clearly mention that these procedures are for Skyline employees only, because without access to the links, it is not possible for users to follow these procedures.

miguel-slc:
Changes looks ok

MariekeGO (approved):

PR #1308, opened 2023-04-06
Added port 8222 again

SLC-Laurens-Vercruysse:
Port 8222 was mentioned as optional, however, the benefits of opening the port were not explained. Added port 8222 again to the list and amended the note.

### DIFF
--- a/user-guide/Advanced_Functionality/DataMiner_Agents/Configuring_a_DMA/Configuring_the_IP_network_ports.md
+++ b/user-guide/Advanced_Functionality/DataMiner_Agents/Configuring_a_DMA/Configuring_the_IP_network_ports.md
@@ -36,7 +36,7 @@ A DataMiner System makes extensive use of TCP/IP communication. Below, you find 
 >
 > - When viewing Stream via DataMiner Cube, access to port 23/tcp is not required. Access is only required when using a Telnet client. However, note that Telnet is by default disabled from DataMiner 9.6.5 onwards. For more information on how to enable this, see [DataMiner.xml](xref:DataMiner_xml)
 > - Prior to DataMiner 10.0.8, ports 7001, 7199, 9142 and 9160 are also opened during Cassandra installation. However, from DataMiner 10.0.8 onwards, only the essential ports 7000 and 9042 are opened.
-> - Prior to DataMiner 10.1.0 CU10 and 10.2.1, port 8222 is also opened for NATS monitoring. After those versions, we recommend opening the port manually in order to be able to debug NATS. With port 8222 you can extract metrics and performance indicators from the NATS message broker, which in turn may allow you to debug if required.  
+> - Prior to DataMiner 10.1.0 CU10 and 10.2.1, port 8222 is also opened for NATS monitoring. In later DataMiner versions, we recommend opening the port manually in order to be able to debug NATS. With port 8222, you can extract metrics and performance indicators from the NATS message broker, which in turn may allow you to debug if required.
 
 ## Graphical representation of IP communication within a DMS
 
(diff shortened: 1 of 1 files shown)


MariekeGO:
@SLC-Laurens-Vercruysse, could you also adjust this in the [DataMiner hardening guide](https://docs.dataminer.services/user-guide/Advanced_Functionality/Security/DataMiner_hardening_guide.html#configure-the-firewall)? Otherwise we'll have contradictory information in the documentation.

SeppeDejonckheereSkyline:
@SLC-Laurens-Vercruysse, because of the sensitivity of the NATS metrics and the lack of access control, I would not recommend to expose port 8222 via the firewall. With the port closed, the information remains accessible, but only directly from localhost, which seems like a better solution to me. What are your thoughts?

SeppeDejonckheereSkyline (commented):
Because of the sensitivity of the NATS metrics and the lack of access control, I would not recommend to expose port 8222 via the firewall. With the port closed, the information remains accessible, but only directly from localhost, which seems like a better solution to me. What are your thoughts?

SLC-Laurens-Vercruysse:
> Because of the sensitivity of the NATS metrics and the lack of access control, I would not recommend to expose port 8222 via the firewall. With the port closed, the information remains accessible, but only directly from localhost, which seems like a better solution to me. What are your thoughts?

Not sure here. It's an essential tool for debugging and would be very useful if we could reach these from within the DMS .
I don't see a great amount of harm if it remains within the DataMiner cluster and is not exposed to the outside world, especially since the information there tends to be kind of cryptic. If need be, we can have a chat in the office regarding this.

SeppeDejonckheereSkyline:
Had a quick chat with Laurens last week, I agree that we can recommend to open the port again but I would also like to add a note stating the the port should only be exposed on internal networks in order to prevent the leaking of sensitive information.

user-guide/Advanced_Functionality/DataMiner_Agents/Configuring_a_DMA/Configuring_the_IP_network_ports.md
```diff
@@ -35,7 +36,7 @@ A DataMiner System makes extensive use of TCP/IP communication. Below, you find
 >
 > - When viewing Stream via DataMiner Cube, access to port 23/tcp is not required. Access is only required when using a Telnet client. However, note that Telnet is by default disabled from DataMiner 9.6.5 onwards. For more information on how to enable this, see [DataMiner.xml](xref:DataMiner_xml)
 > - Prior to DataMiner 10.0.8, ports 7001, 7199, 9142 and 9160 are also opened during Cassandra installation. However, from DataMiner 10.0.8 onwards, only the essential ports 7000 and 9042 are opened.
-> - Prior to DataMiner 10.1.0 CU10 and 10.2.1, port 8222 is also opened for NATS monitoring. However, this port is no longer used from those versions onwards and is no longer opened during new installations.
+> - Prior to DataMiner 10.1.0 CU10 and 10.2.1, port 8222 is also opened for NATS monitoring. In later DataMiner versions, we recommend manually opening the port in order to be able to debug NATS. With port 8222, you can extract metrics and performance indicators from the NATS message broker, which in turn may allow you to debug if required. However, for maximum security, only expose port 8222 on internal networks to prevent the leaking of sensitive information.
```
EdithLansens, line 39: @SeppeDejonckheereSkyline Is this note OK?

SeppeDejonckheereSkyline: Looks good to me

PR #1321, opened 2023-04-11
RN 35999

MariekeGO:


skyline-rge (approved):
The note may indeed be a bit confusion. The Key property itself can indeed be null (which means a row key will be auto-generated) but when providing an actual row key this may not be null.
I'm not sure if there is a better way of conveying this.
Other than that, looks good to me.

skyline-rge (approved):

PR #1322, opened 2023-04-11
Test to check whether xrefs to .NET docs still compile through the workflow

mmphs:


mmphs:
Breaking change in docfx 2.65.0

PR #1374, opened 2023-04-25
RN 36248

MariekeGO:


miguel-slc:
Hi,
Please could we explain what how the bitrate is calculated?

MariekeGO:
> Hi, Please could we explain what how the bitrate is calculated?

I've added more info on this. Could you take another look?

miguel-slc:
Hi, It looks ok. Thanks!

PR #1384, opened 2023-04-26
Update Configuring_external_authentication_via_an_identity_provider_u…

KevinVerm:
…sing_SAML.md

Added prerquisite from RNhttps://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm.aspx?ID=35772

See the description field in this RN


KevinVerm:
@klaasdebeuf

MariekeGO:
When we documented this, @wimme informed us that this prerequisite would be added automatically when you upgrade to 10.3.5/10.4.0, so this is not something a user really needs to worry about. Have you come across a situation where this is a problem after all?

KevinVerm:
> When we documented this, @wimme informed us that this prerequisite would be added automatically when you upgrade to 10.3.5/10.4.0, so this is not something a user really needs to worry about. Have you come across a situation where this is a problem after all?

Hi @MariekeGO, no not yet. I searched the help for it, but I didn't find that those tags should be in the web.config file. I thought it was a prerequisite also since it was mentioned in the RN. Better informed than not informed. If it's added automatically already(so not task lurking with some domain) then probably this PR can be closed.

PR #1407, opened 2023-05-02
ArtifactDeployer 1.4.5 no longer requires direct internet access [ID_36284]

JordyGit:


JordyGit:
add warning min cloud gateway version and port 5100

PR #1461, opened 2023-05-11
Options measurement type

EdithLansens:
The page with UID MeasurementTypeOptionsOverview was not included in the toc.yml file. However, it gets referenced on the page about the options attribute of Param > Measurement > Type. Similar to pages like https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.Params.Param.SNMP.OID-options.html I have now added all options to the existing @ options page of Type, making a second, separate options page unnecessary.


EdithLansens:
Is this OK, @PedroDebevere?

PedroDebevere (approved):

PR #1486, opened 2023-05-17
Info added on DOM DevOps tools installation

MariekeGO:


MariekeGO:
Updated based on https://community.dataminer.services/question/dom-devops-tools/

PR #1487, opened 2023-05-17
Update DB_xml.md

robin-devos-skyline:
Based on DCP64018/RN15345
@Tom-Van-Bruaene and @bertbsky could you verify this addition? I found your names in the DCP+RN

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Offload_database/Configuring_data_offloads.md
+++ b/user-guide/Advanced_Functionality/Databases/Offload_database/Configuring_data_offloads.md
@@ -72,6 +72,9 @@ The offload rate for trend data and alarm data can be configured in System Cente
 
    - If all previous values should be removed from the database each time a new snapshot is offloaded, make sure the option *Only keep the latest parameter value in the database* is selected. If this option is not selected, all values will be kept.
 
+     > [!NOTE]
+     > This option corresponds with the `clean="true"` configuration in the Offload tag in DB.xml. If it is activated, DataMiner will truncate the data table in the offload database before new information is added. This way, the table always contains the latest snapshot information only.
+
 ## Disabling data offloads to the offload database on element level
 
 By default, data offloads are enabled for every element. In every *Element.xml* file, you should find the following setting:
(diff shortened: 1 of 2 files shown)


MariekeGO:
@robin-devos-skyline, after a quick search, I found that this was not added to the documentation because it's possible to do this in Cube instead with the "Only keep the latest parameter value in the database (clear previous values)" option. 
image

As we prefer it if people don't mess around with XML files if there's a better alternative available directly in Cube, it seemed better to leave this out. Should this be added after all? Does adding the option in DB.xml actually do something different than this option?

robin-devos-skyline:
@MariekeGO, I understand this perspective. 
The problem with this is, from my side, I only knew the clean option in the xml and I found nothing in the DM Docs, so I knew nothing. Even knowing that option in Cube, ie "Only keep the latest parame...", I had a hard time finding it in the DM Docs (I searched for "clear previous values" which seemed like more relevant terms). 
--> perhaps change "clear previous values" to "clea**N** previous values"? small change, but in a way it refers to the clean option, which does make a difference when searching.

Either way, final decision isn't for me to make. I know other xml tags/options/arguments aren't mentioned in the DM Docs either.

robin-devos-skyline:
> > @MariekeGO, I understand this perspective. The problem with this is, from my side, I only knew the clean option in the xml and I found nothing in the DM Docs, so I knew nothing. Even knowing that option in Cube, ie "Only keep the latest parame...", I had a hard time finding it in the DM Docs (I searched for "clear previous values" which seemed like more relevant terms). --> perhaps change "clear previous values" to "clea**N** previous values"? small change, but in a way it refers to the clean option, which does make a difference when searching.
> > Either way, final decision isn't for me to make. I know other xml tags/options/arguments aren't mentioned in the DM Docs either.
> 
> Perhaps we can add a note mentioning that this option in Cube corresponds with the clean="true" configuration in the Offload tag in DB.xml? That way it will show up if you search for that configuration.

Sounds like a good idea 👍

MariekeGO:
@robin-devos-skyline I've adjusted the PR based on the comments above. Could you take a look and let me know if this is OK for you?

robin-devos-skyline (commented):
Approved :)

MariekeGO (approved):

PR #1506, opened 2023-05-23
Update Troubleshooting_Trending.md

robin-devos-skyline:
Links added to provide details as it's not always easy to know or find these.

I wasn't sure if I had to add the links directly to the Mermaid shapes or via a note, i.e. in terms of being future-proof

Overview diagram > Trend Inspector shape, I removed the details within the shape since I added the link towards its DM Docs page


EdithLansens:
Typically, we include the links directly within the flowcharts. While this may not be the most update-friendly method, it is the most user-friendly one. If we were to place all links in accompanying notes, certain flowcharts would have an overwhelming list of links, potentially discouraging readers from navigating them because of the constant need to switch between the flowchart and the notes to follow instructions. However, if a link is relevant but doesn't fit within a specific flowchart shape, we can include it within a note.

EdithLansens (approved):

PR #1514, opened 2023-05-23
Update LogicQActions.md

DanielaOlive:
improved clarity on Idisposable (works only when entry point method is non-static as well). Example: non-static QAction class with a static Run method does not work.

### DIFF
--- a/develop/devguide/Connector/LogicQActions.md
+++ b/develop/devguide/Connector/LogicQActions.md
@@ -485,7 +485,7 @@ From DataMiner 10.2.9 onwards (RN 33965), DataMiner detects whether the [IDispos
 > [!NOTE]
 >
 > - DataMiner will only create an instance of the class containing the entry point method if this method is not a static method (see [Instance entry methods](xref:LogicQActions#instance-entry-methods)). Therefore, if you want to make use of the IDisposable functionality, make sure you use a non-static entry point method so that an instance gets created.
-> - This also applies to any other class the entrypoint may be in (see [Multiple entry methods](xref:LogicQActions#multiple-entry-methods)).
+> - This also applies to any other class the entry point may be in (see [Multiple entry methods](xref:LogicQActions#multiple-entry-methods)).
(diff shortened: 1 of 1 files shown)


develop/devguide/Connector/LogicQActions.md
```diff
@@ -480,7 +480,7 @@ For more information, refer to Skyline.DataMiner.Net.Messages.
 
 ## Implementing the IDisposable interface
 
-From DataMiner 10.2.9 onwards (RN 33965), DataMiner detects whether the [IDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable) interface is implemented on non-static QAction classes. DataMiner will then call the Dispose method when the QAction instance is released (i.e. when the element is stopped, removed or restarted).
+From DataMiner 10.2.9 onwards (RN 33965), DataMiner detects whether the [IDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable) interface is implemented on non-static QAction classes with a non-static entry point method. DataMiner will then call the method when the QAction instance is released (i.e. when the element is stopped, removed, or restarted).
```
PedroDebevere, line 483:
```suggestion
From DataMiner 10.2.9 onwards (RN 33965), DataMiner detects whether the [IDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable) interface is implemented on non-static QAction classes with a non-static entry point method. DataMiner will then call the [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) method when the QAction instance is released (i.e. when the element is stopped, removed, or restarted).

> [!NOTE]
> DataMiner will only create an instance of the class containting the entry point method if this method is not a static method (see [Instance entry methods](xref:LogicQActions#instance-entry-methods)). Therefore, if you want to make use of the IDisposable functionality, make sure that an instance gets created by having a non-static entry point method.
```

PedroDebevere (changes requested):

MariekeGO (approved):

MariekeGO (approved):

PR #1539, opened 2023-05-30
Update Changing_the_layout_of_the_Alarm_Console.md

VeerleLedoux:
corrected the information about suggestion events: those are no longer only generated by the behavioral anomaly detection feature. The suggestion events tab can also contain notifications about predicted alarms or detected trend patterns.


EdithLansens:
> corrected the information about suggestion events: those are no longer only generated by the behavioral anomaly detection feature. The suggestion events tab can also contain notifications about predicted alarms or detected trend patterns.

Is this change available in all DataMiner versions? Should I mention a DataMiner version that introduces notifications about predicted alarms or detected trend patterns, alongside behavioral anomaly detection?

VeerleLedoux:
The features will not be available in old DataMiner versions indeed. 
The page "Advanced analytics features in the Alarm Console" mentions the dataminer versions where the notifications for predicted alarms, trend patterns, and behavioral anomaly detection were added. Note that in a subsequent pull request I suggested a change to that page to have a clearer overview of the suggestion-generating features. If we simply refer to that page, I guess we are good? Refering to that page also had the advantage that if in the future we add an extra feature generating suggestion events, then we need to add it only to that one page.
So maybe we should then replace "This includes notifications generated by the [proactive cap detection feature](xref:Proactive_cap_detection), and notifications for [behavioral anomaly detection](xref:Working_with_behavioral_anomaly_detection) and [pattern matching](xref:Monitoring_of_trend_patterns)." by something like " see Advanced analytics features in the Alarm Console" as well?

VeerleLedoux (commented):
I approve the latest change

PR #1543, opened 2023-05-30
RN 36440

EdithLansens:
Hi @VeerleLedoux, could you take a look at this pull request to double-check whether I've correctly added release note 36440 to the user guide?


VeerleLedoux (changes requested):
The first line in the text was not entirely correct and complete. I suggest something like this: 
The value of the suggestion event mentions what kind of issue is expected. For instance, for predicted breaches of critical normal alarm thresholds the value will contain "Predicted Critical High" or "Predicted Critical Low". For trend which is forecasted to reach a parameter's minimum or maximum data range value, from DataMiner version 10.3.7/10.4.0 onwards, the value of the suggestion event mentions "Predicted maximum value" or "Predicted minimum value" followed by the actual maximum or minimum value of the parameter,
 along with the corresponding unit of measurement, as specified in the protocol. Prior to DataMiner version 10.3.7/10.4.0, the value of the suggestion event mentions the text "Predicted above range violation" or "Predicted below range violation".

MariekeGO (approved):

PR #1547, opened 2023-05-31
Changed url to pid

SlcADK:
If the value comes from a parameter, it should be pid.


EdithLansens:
@PedroDebevere, is this OK to merge?

PedroDebevere (approved):

PedroDebevere (approved):

PR #1548, opened 2023-05-31
Changed url to pid (use case 4)

SlcADK:
If the value comes from a parameter, it should be pid.


EdithLansens:
@PedroDebevere, is this OK to merge?

PedroDebevere (approved):

PR #1564, opened 2023-06-02
Update Protocol.Params.Param.Measurement.Type.md

AlexJBSilva:
Fixes the bullet text to match the link it is referencing to: Toggle button.

I suggest you also ask an expert to revise the explanation about [togglebutton](https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.Params.Param.Measurement.Type.html?q=togglebutton#togglebutton), because it is NOT consistent with the explanation in [Toggle button](https://docs.dataminer.services/develop/devguide/Connector/UIComponentsToggleButton.html) page .

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Params.Param.Measurement.Type.md
+++ b/develop/schemadoc/Protocol/Protocol.Params.Param.Measurement.Type.md
@@ -112,7 +112,7 @@ For example:
 </Measurement>
 ```
 
-See also
+See also:
 
 - [Options for measurement type “discreet”](xref:Protocol.Params.Param.Measurement.Type-options#options-for-measurement-type-discreet)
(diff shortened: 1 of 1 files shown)


PedroDebevere (approved):
To avoid confusion, I suggest to remove the sentence "In this case, the accompanying Protocol.Params.Param.Measurement.Discreets tag is used for the parameter of type READ or READ BIT. In rare cases it can be set on the parameter of type WRITE." from [https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.Params.Param.Measurement.Type.html?q=togglebutton#togglebutton](https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.Params.Param.Measurement.Type.html?q=togglebutton#togglebutton)

MariekeGO (approved):

PR #1580, opened 2023-06-08
PopupSkeleton and PopupDetails

spetebroodt:


MariekeGO (commented):
While this is already fine as it is, I think it would be even better if we could get rid of the large tables under "Attributes of the Detail tag" and "Type overview". If we could work with subtitles there, it would make the text easier to scan. As an added bonus, it would also allow more detailed cross-referencing, and it would make the page more edit-friendly.

PR #1588, opened 2023-06-12
Update CSV Export RN36467

TVincke98:
Add documentation for new feature of CSV Export for the table component


MariekeGO:
We always need to make sure the documentation supports all supported DataMiner versions, so information related to older supported versions can't just be replaced with the new functionality. I've adjusted the pull request accordingly.
We also need to mention both main and feature release version whenever relevant. I've assumed this will be merged in 10.4.0, but if it isn't, this will still need to be adjusted.
Note that I will revert this PR to draft for now. As soon as the feature has been merged to a DataMiner version, the PR can be marked as ready for review again.

MariekeGO (approved):

PR #1600, opened 2023-06-13
Note on indexing nodes config in DB.xml

MariekeGO:
Added based on https://community.dataminer.services/question/adding-a-three-node-elasticsearch-cluster/


wale-007 (approved):
All good.

PR #1606, opened 2023-06-13
Create Failed_to_get_log_file.md

robin-devos-skyline:
There are various cases which could lead to "Failed to get log file", I tried to nicely list them here (the ones I know).
Under "401 - Unauthorized", I uses Possible Cause 1 and 2, not sure if this is ok. I wanted to make a clear distinction as their Resolutions are different.

### DIFF
--- a/user-guide/Troubleshooting/Error_Messages/Failed_to_get_log_file.md
+++ b/user-guide/Troubleshooting/Error_Messages/Failed_to_get_log_file.md
@@ -48,6 +48,8 @@ This error code and message can have the following possible cause:
   netsh winhttp show proxy
   ```
 
+  If the result shows the proxy server used and the bypass list, a proxy server is indeed configured.
+
   **Resolution**: Add all IPs and hostnames of the DMAs to the proxy exception list. You can do so by opening a command prompt on the server and executing the following command:
 
   ```txt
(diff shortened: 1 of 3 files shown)


user-guide/Troubleshooting/Error_Messages/Failed_to_get_log_file.md
```diff
@@ -0,0 +1,83 @@
...
+```txt
+Failed to get log file (403 - Forbidden)
+```
+
+This error code and message can have the following possible cause:
+
+- **Possible cause**: This usually occurs when a proxy server is configured. To check if this is indeed the case, open a command prompt on the server and execute the following command:
+
+  ```txt
+  netsh winhttp show proxy
```
MariekeGO, line 48: What should you see when you execute this command? It might be useful to add this, so the user will know what to do based on the result.

robin-devos-skyline: The result is actually quite straightforward, it either shows "Direct access (no proxy server)." or it shows the proxy server used and the bypass list
image
In the next command, we update the bypass list

MariekeGO (approved):

PR #1611, opened 2023-06-14
DBconfiguration.xml sync note

MariekeGO:
Based on https://community.dataminer.services/question/is-the-dbconfiguration-xml-synchronized-with-the-failover-pair-during-the-failover-enable/.


BrentSLC (approved):
LGTM

PR #1634, opened 2023-06-16
Delete Overview_Support_warranty.md

michelhofmans:
I suggest to the delete this section. Warranty is something different than support. It doesn't fit here and has no added value.


MariekeGO:
@SammyDW, since you originally added this section, could you confirm if it's OK to delete this?

SammyDW:
@MariekeGO , spoke with Michel and I'm OK to delete this section. This paragraph will be possibly adapted/rewritten later on by Michel, and added to a new legal section on DataMiner Overivew.

MariekeGO (approved):

PR #1642, opened 2023-06-21
Update Adding_time_scoped_related_parameters_to_a_trend_graph.md

SkylineDennisDr:
I changed the first paragraph: it incorrectly referred the lightbulb icon on the top right of the trend graph, whereas a new lightbulb icon was introduced specifically for this feature. I also added some more info in that paragraph.

In the second paragraph, I removed "a" from "in a case".

Finally, the second item in the bottom section was reformulated.

### DIFF
--- a/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
+++ b/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
@@ -4,7 +4,7 @@ uid: Adding_time_scoped_related_parameters_to_a_trend_graph
 
 # Adding time-scoped related parameters to a trend graph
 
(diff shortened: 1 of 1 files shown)


user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
```diff
@@ -4,9 +4,9 @@ uid: Adding_time_scoped_related_parameters_to_a_trend_graph
 
 # Adding time-scoped related parameters to a trend graph
 
-From DataMiner 10.3.8/10.4.0 onwards<!--RN 36434 -->, you can use the light bulb in the top-right corner of a trend graph to add related parameters based only on the behavior of the parameters during the time range of a selected section of the trend graph.
+From DataMiner 10.3.8/10.4.0 onwards<!--RN 36434 -->, when you select a time range of a trend graph, a light bulb icon will be displayed. Click this icon to add related parameters based only on the behavior during the selected time range. This is different from the light bulb in the top right corner of the trend graph, which proposes related parameters based on the entire history. Currently, only parameters from the same DataMiner element can be proposed. The light bulb only shows parameters related to one parameter, even if multiple curves are visible on the trend graph. You can hover over the light bulb icon to see for which parameter relations are proposed.
```
MariekeGO, line 7: This paragraph is now quite confusing, because it's not clear from the fourth sentence onwards which of the light bulb icons you are talking about. I'd propose moving the information about the other light bulb feature to a separate, short note, with a cross-reference to a page with more information about this (i.e. [Adding related parameters to a trend graph](https://docs.dataminer.services/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_related_parameters_to_a_trend_graph.html)).
If you want a more extensive comparison between the two features, this should be placed on [this page](https://docs.dataminer.services/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Working_with_relation_learning.html), as this is where we currently explain the difference between the two.

PR #1644, opened 2023-06-21
Adding an element

EdithLansens:
- The previous "Adding and deleting elements" has now been split up into three different pages: "Adding elements", "Deleting elements", and "Duplicating elements".
- All different connection types (previously listed under "Adding an element") now have a separate page with instructions on how to configure the connection settings.
- GPIB-related documentation has been improved: 1. The old "Configuring the GPIB settings of a spectrum analyzer element" page has been removed and replaced with the new "GPIB connections" page, now a subpage to "Adding an element" instead of under "Installation and configuration of a spectrum analyzer". 2. An image has been created to simplify the concept of a GPIB gateway. 3. The "Installation and configuration of a spectrum analyzer" is now an introductory page instead of a mini-TOC page.


user-guide/images/GPIB.svg
MariekeGO, line 1: Could you modify this image so there is no hyphen in "IP-address"

PR #1653, opened 2023-06-21
Update Upgrading_a_DataMiner_Agent_in_the_Update_Center.md

robin-devos-skyline:
I wasn't able to find what the default option actually is (DM Docs, RNs, Tasks), I tested this myself locally to be certain. Also removed the last bit in the 2nd (simultaneous) option, it didn't make much sense to me knowing that the default option = the simultaneous option


MariekeGO (changes requested):
As a matter of fact, the default policy can be something else, depending on what is configured in System Center. This is explained in a note in the [procedure to upgrade using Taskbar Utility](https://docs.dataminer.services/user-guide/Advanced_Functionality/DataMiner_Agents/Upgrading_a_DMA/Upgrading_a_DataMiner_Agent_using_DataMiner_Taskbar_Utility.html?q=%22default%20Failover%22):

image

I'd therefore propose that instead of the changes in your PR, we add that note on the page related to upgrading a DMA in the Update Center as well.

robin-devos-skyline:
We learn something new everyday 🙈 
Could this Note be added to this page as well?

MariekeGO:
> We learn something new everyday 🙈 Could this Note be added to this page as well?

That's what I meant, actually 😊
I'll take care of it!

MariekeGO (approved):

PR #1654, opened 2023-06-21
Update Configuring_the_default_upgrade_options.md

robin-devos-skyline:
default policy = simultaneous policy

### DIFF
--- a/user-guide/Advanced_Functionality/DataMiner_Agents/Upgrading_a_DMA/Configuring_the_default_upgrade_options.md
+++ b/user-guide/Advanced_Functionality/DataMiner_Agents/Upgrading_a_DMA/Configuring_the_default_upgrade_options.md
@@ -24,7 +24,7 @@ To do so:
 
    - *Failover policy*: This option allows you to select the default policy for upgrading a Failover setup. This determines in which order the DMA in a Failover setup are upgraded by default.
 
-     > [!NOTE]
+     > [!IMPORTANT]
      > We highly recommend that you set this to *Upgrade main and backup Agent simultaneously*.
 
 1. After you have configured the options to your preferences, click the *Save* button in the lower right corner.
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/DataMiner_Agents/Upgrading_a_DMA/Configuring_the_default_upgrade_options.md
```diff
@@ -25,6 +25,6 @@ To do so:
    - *Failover policy*: Determines in which order the DMA in a Failover are upgraded.
 
      > [!NOTE]
-     > We highly recommend that you set this to *Upgrade main and backup Agent simultaneously*.
+     > We highly recommend that you set this to *Upgrade main and backup Agent simultaneously*, which is also the *default policy*.
```
MariekeGO, line 28: This change does not make sense. "Upgrade main and backup Agent simultaneously" is only the default policy if you configure it to be the default policy, which is exactly what the user does here. I'd propose that instead we update line 25-28 as follows to make this clear: 

```
- *Failover policy*: This option allows you to select the default policy for upgrading a Failover setup. This determines in which order the DMA in a Failover setup are upgraded by default.

     > [!IMPORTANT]
     > We highly recommend that you set the default Failover policy to *Upgrade main and backup Agent simultaneously*.
```

MariekeGO (changes requested):

robin-devos-skyline:
Yes, please disregard this commit
Looking back at it (with your comment in the other commit), it indeed doesn't make sense.

MariekeGO:
> Yes, please disregard this commit Looking back at it (with your comment in the other commit), it indeed doesn't make sense.

It does make clear that the current text isn't clear enough, so I'll adjust the PR the way I suggested.

MariekeGO (approved):

PR #1669, opened 2023-06-23
Extend upgrade info

wimme:
The docs seem to contain only information about upgrading the entire DataMiner system.

However, we also offer different upgrade options to update some components separately, without having to do an entire DataMiner upgrade.


MariekeGO:
@wimme Because putting the section about types of upgrades at the start of this page could potentially be confusing for the rest of the page (i.e. the user might not be sure if other sections are only for a DataMiner upgrade or also for another type of upgrade), I've moved it to the end of the page and adjusted it a bit. I've also simplified some things (e.g. it doesn't seem necessary to me to go into so much detail as to the process running GQI) and added some links. Could you check if everything is still OK for you after the changes?

wimme:
Reason I put it at top was to make more people aware of all the upgrade types, since we got quite some feedback from customers saying they were not aware of this. But it shouldn't be confusing indeed, at the bottom will indeed be less confusing.

Looks good to me!

MariekeGO (approved):

PR #1701, opened 2023-06-29
RNs 36761, 36780, 36781, and 36782

MariekeGO:


miguel-slc:
Hi ,

Regarding this paragraph:

`To make sure events at lower levels in the topology do not become too numerous, making it difficult to analyze the points of failure, root cause analysis has been implemented via alarm template conditions. This will suppress lower-level events when higher-level events are detected.`

Can we indicate on which topology levels will this work/has been tested? In case it works for any topology level. We don't need to add any extra comment.

`On the Cable Modems page of a Generic DOCSIS CM Collector element, it could occur that some rows in the Cable Modem Overview table incorrectly did not show any information in the CCAP Core ID column.`

Maybe it is better to rephrase it as follows: _On the Cable Modems page of elements running the connector 'Generic DOCSIS CM Collector'_

Regarding this paragraph:
`The EPM front-end element now no longer needs to create unique IDs for all cable modems in the system. This will improve the turnaround for new IDs, and new entities in the system will appear more quickly.`

For the FE element, can we include the connector name. Eventually we will have multiple FEs, so it is better if we mention the connector affected.

MariekeGO:
> Hi ,
> 
> Regarding this paragraph:
> 
> `To make sure events at lower levels in the topology do not become too numerous, making it difficult to analyze the points of failure, root cause analysis has been implemented via alarm template conditions. This will suppress lower-level events when higher-level events are detected.`
> 
> Can we indicate on which topology levels will this work/has been tested? In case it works for any topology level. We don't need to add any extra comment.
> 
> `On the Cable Modems page of a Generic DOCSIS CM Collector element, it could occur that some rows in the Cable Modem Overview table incorrectly did not show any information in the CCAP Core ID column.`
> 
> Maybe it is better to rephrase it as follows: _On the Cable Modems page of elements running the connector 'Generic DOCSIS CM Collector'_
> 
> Regarding this paragraph: `The EPM front-end element now no longer needs to create unique IDs for all cable modems in the system. This will improve the turnaround for new IDs, and new entities in the system will appear more quickly.`
> 
> For the FE element, can we include the connector name. Eventually we will have multiple FEs, so it is better if we mention the connector affected.

@miguel-slc, I've implemented the necessary changes. Could you take another look?

PR #1703, opened 2023-06-29
SLProtocol bitness update

Skyline-NicolasVC:
From Dataminer 10.3.9 onwards, SLProtocol will be 64-bit.
A soft-launch is provided to go back to 32-bit for compatibility reasons.
Updated documentation with this soft-launch.

### DIFF
--- a/user-guide/Reference/Soft-launch_options/Activating_SLProtocol_as_a_32_Bit_Process.md
+++ b/user-guide/Reference/Soft-launch_options/Activating_SLProtocol_as_a_32_Bit_Process.md
@@ -1,0 +1,50 @@
+---
+uid: Activating_SLProtocol_as_a_32_Bit_Process
+---
+
+# Activating SLProtocol as a 32-bit process
+
+From **DataMiner 10.3.9 onwards** SLProtocol will be 64-bit by default but a fallback soft-launch has been provided.
+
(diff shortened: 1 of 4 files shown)


FlorisCO:
We may want to mention in the how-to-64 page that the behavior changed to 64 by default from 10.3.9?

Skyline-NicolasVC:
added proposed change.

user-guide/Reference/Soft-launch_options/Activating_SLProtocol_as_a_32_Bit_Process.md
```diff
@@ -0,0 +1,50 @@
+---
+uid: Activating_SLProtocol_as_a_32_Bit_Process
+---
+
+# Activating SLProtocol as a 32-bit process
+
+From **DataMiner 10.3.9 onwards** SLProtocol will be 64-bit by default but a fallback soft-launch has been provided.
```
MariekeGO, line 7: Is this the change introduced by RN 36725? If yes, could you still add this RN number in comment here? This makes it easier to track down where an RN has been added later.

In addition, we also need to make sure to always mention both the feature and the main release version. For instance, in case it's 10.4.0, this is what I'd propose:

```suggestion
From DataMiner 10.3.9/10.4.0 onwards, SLProtocol is a 64-bit process by default. However, a fallback soft-launch has been provided.<!-- RN 36725 -->
```

MariekeGO (approved):

PR #1723, opened 2023-07-05
Specify n need to have CI Types and high-level overview to g…

jaklke:
…et started.


jaklke:
@MariekeGO , updated with the remarks.

MariekeGO (approved):

PR #1731, opened 2023-07-07
EPM GPON Documentation

ALizcanoR:
This is the first version of the EPM GPON Solution docs.
I added some external links as references for one of the pages, they seem required as we used such definitions to create the connectors and the data streams that are the base for this project.

If something is not right, please let me know, as this is my first contribution and it seems a rather big one. Thanks!


MariekeGO:
@ALizcanoR, I've reviewed the PR and made the necessary changes. Could you check if everything is still OK for you?

MariekeGO (approved):

PR #1756, opened 2023-07-13
Update Annotations life cycle status

SeppeDejonckheereSkyline:
Changed the life cycle status for annotations from 'end of engineering' to 'end of life'


user-guide/Reference/Software_support_life_cycles.md
```diff
@@ -34,7 +34,7 @@ Important announcements about the latest changes to DataMiner core functionality
 |---------|---------|---------|---------|
 | Jobs app | End of Engineering | Module is being retired. | End of Life to be confirmed |
 | Ticketing app | End of Engineering | Module is being retired. | End of Life to be confirmed |
-| Annotations | End of Engineering | Module is being retired. | End of Life as of DataMiner version 10.3.x (Q4 2022) |
+| Annotations | End of Life | From DataMiner version 10.3.x (Q4 2022) onwards, Annotations is End of Life. | End of Life |
```
MariekeGO, line 37: "is End of Life" is a bit odd. Would it be correct if we changed it like this?
```suggestion
| Annotations | End of Life | From DataMiner version 10.3.x (Q4 2022) onwards, Annotations can no longer be used. | End of Life |
```

SeppeDejonckheereSkyline: Actually it can still be used and it's enabled by default. You have to manually disable it via a softlaunch flag so 'can no longer be used' isn't really correct here.

MariekeGO: Since it's "End of Life", could we say "Annotations should no longer be used"?

SeppeDejonckheereSkyline: That sounds good to me

MariekeGO (approved):

PR #1763, opened 2023-07-14
RN 36784

MariekeGO:


miguel-slc (commented):
_In order to suppress alarms for routes, distributions, and FAT levels, the alarm template has been modified to include conditions for these levels. The alarm will be suppressed if the following conditions are met:_
The alarm template assigned to which element/connector?

The name of the feature is called 'Root Cause Analysis', but actually we are not analyzing the root cause of an alarm, but we are only removing alarms at the bottom level to reduce the number of alarms in the console. If a user would like to check alarms at the bottom level cause the alarm at the top level, they will not be able to see them because these alarms were disabled.
Can we rename the feature to 'Alarm Suppression at lower levels'?
Open to other suggestions!

MariekeGO:
> _In order to suppress alarms for routes, distributions, and FAT levels, the alarm template has been modified to include conditions for these levels. The alarm will be suppressed if the following conditions are met:_ The alarm template assigned to which element/connector?
> 
> The name of the feature is called 'Root Cause Analysis', but actually we are not analyzing the root cause of an alarm, but we are only removing alarms at the bottom level to reduce the number of alarms in the console. If a user would like to check alarms at the bottom level cause the alarm at the top level, they will not be able to see them because these alarms were disabled. Can we rename the feature to 'Alarm Suppression at lower levels'? Open to other suggestions!

I've updated the PR according to your suggestion! OK to merge like this?

miguel-slc:
It looks ok. Thanks for the changes!

PR #1764, opened 2023-07-14
Update DataMiner_Extension_Modules.md

PieterVanCompernolle:
updated with the evolution that was discussed in today's meeting


user-guide/Reference/DataMiner_Extension_Modules.md
```diff
@@ -4,12 +4,19 @@ uid: DataMinerExtensionModules
...
+# Evolution
+
+From the list of DxMs, we will extract some modules that are actually core modules created with the DXM technology and call them 'DMs'. DxMs and DMs will be treated differently during upgrades. DMs will be installed (or updated if already installed) during Core upgrade/update packages, while DxMs will only be updated when they were already installed. 
+DxMs will have a 'minimal version of the core' dependency, and that will be made transparent through the cloud admin page. When installing updates, the latest compatible update will be automatically selected.
 
 ## Packs
 
 For ease of installation, DxMs may be packaged together in a pack, such as the [Cloud Pack](xref:CloudPackages). These packs allow users to install several DxMs at once.
 
+## Available DMs
```
MariekeGO, line 18: Are any DMs available yet? If so, could you add these below this title? I'd prefer not to have a dummy section without any content here, as that may look quite confusing.

user-guide/Reference/DataMiner_Extension_Modules.md
```diff
@@ -4,12 +4,18 @@ uid: DataMinerExtensionModules
...
+A DataMiner Extension Module (DxM) is a service that can be installed, upgraded, and uninstalled without the need to reboot the DataMiner Agent. These modules will expand the core DataMiner software with additional features and capabilities. In most cases, DxMs will be backwards-compatible with all supported core DataMiner versions. However, this may not be possible in some cases, and then a minimum core DataMiner version will be indicated.
+
+## Evolution
+
+From the list of DxMs, we will extract some modules that are actually core modules created with the DXM technology and call them 'DMs'. DxMs and DMs will be treated differently during upgrades. DMs will be installed (or updated if already installed) during Core upgrade/update packages, while DxMs will only be updated when they were already installed. DxMs will have a 'minimal version of the core' dependency, and that will be made transparent through the cloud admin page. When installing updates, the latest compatible update will be automatically selected.
```
MariekeGO, line 11: To make this section more future-proof and get the phrasing more in line with the rest of the documentation, I'd suggest we rephrase this:

```suggestion
Some DxMs are considered core modules created using the DxM technology. These are known as "DMs". DMs are treated differently from other DxMs during upgrades. While other DxMs are only updated if they have already been installed, a DM will always be installed or updated with an upgrade.

DxMs depend on a minimum version of the core software. This is indicated in the Admin app. When you install updates, the latest compatible update will be automatically selected.
```

It would probably also be better to remove the "Evolution" title, as this won't make much sense in the long run, when what is written here has been implemented, and it's no longer a look ahead at how we want to work.

MariekeGO (approved):

PR #1765, opened 2023-07-14
Update DataMiner_Extension_Modules.md

PieterVanCompernolle:
Adding the ambition to be backwards compatible to all supported main release versions

### DIFF
--- a/user-guide/Reference/DataMiner_Extension_Modules.md
+++ b/user-guide/Reference/DataMiner_Extension_Modules.md
@@ -4,7 +4,7 @@ uid: DataMinerExtensionModules
 
 # DataMiner Extension Modules (DxMs)
 
-A DataMiner Extension Module (DxM) is a service that can be installed, upgraded and uninstalled without the need to reboot the DataMiner Agent. These modules will expand the core DataMiner software with additional features and capabilities. To increase the 'Time to Market', the DEV teams will try hard to be backwards compatible to all supported Core DataMiner Versions. Sometimes though, this will not be possible and a minimum Core version will be indicated. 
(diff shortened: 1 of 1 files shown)


user-guide/Reference/DataMiner_Extension_Modules.md
```diff
@@ -4,7 +4,7 @@ uid: DataMinerExtensionModules
 
 # DataMiner Extension Modules (DxMs)
 
-A DataMiner Extension Module (DxM) is a service that can be installed, upgraded and uninstalled without the need to reboot the DataMiner Agent. These modules will expand the core DataMiner software with additional features and capabilities.
+A DataMiner Extension Module (DxM) is a service that can be installed, upgraded and uninstalled without the need to reboot the DataMiner Agent. These modules will expand the core DataMiner software with additional features and capabilities. To increase the 'Time to Market', the DEV teams will try hard to be backwards compatible to all supported Core DataMiner Versions. Sometimes though, this will not be possible and a minimum Core version will be indicated.
```
MariekeGO, line 7: I'm not sure I entirely follow how this is linked with increasing the time to market, but perhaps it's not necessary to even add this? Maybe we can rephrase like this, to keep it simple and easy to understand?

```suggestion
A DataMiner Extension Module (DxM) is a service that can be installed, upgraded, and uninstalled without the need to reboot the DataMiner Agent. These modules will expand the core DataMiner software with additional features and capabilities. In most cases, DxMs will be backwards-compatible with all supported core DataMiner versions. However, this may not be possible in some cases, and then a minimum core DataMiner version will be indicated. 
```

MariekeGO (approved):

PR #1766, opened 2023-07-14
Update Installing_Cassandra.md

AlexanderGorbunovSkyline:
Added information about creating a superuser account.


SkylineCookieMonster (changes requested):
Can we say that it is advised to not use the default credentials for security reasons and that the information can be found on the Cassandra authentication to create a user for DataMiner? Would like to avoid that the information needs to be maintained in two places and now we are not telling them to remove the default user after creating this new super user...

AlexanderGorbunovSkyline:
Removed details on configuring a user; instead just pointed to the Cassandra authentication page. Does this look better?

MariekeGO (approved):

PR #1776, opened 2023-07-18
Update Exporting_and_importing_packages_on_a_DMA.md

SkylineThomasCR:
Fix wrong compatibility direction and make it a warning. Added backwards compatibility warning.
Make dll note a warning.

### DIFF
--- a/user-guide/Advanced_Functionality/DataMiner_Agents/Exporting_and_importing_packages_on_a_DMA.md
+++ b/user-guide/Advanced_Functionality/DataMiner_Agents/Exporting_and_importing_packages_on_a_DMA.md
@@ -15,8 +15,11 @@ This section consists of the following topics:
 > - For more information on how to migrate elements within one DataMiner System, see [Migrating elements in a DataMiner System](xref:Migrating_elements_in_a_DataMiner_System).
 > - For more information on how to import and export elements using CSV files, see [Importing and exporting elements](xref:Importing_and_exporting_elements).
 > - It is also possible to export element information to the printer or the clipboard. To do so, follow the same procedure as to export to a CSV file. See [Exporting elements to a CSV file](xref:Importing_and_exporting_elements#exporting-elements-to-a-csv-file).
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/DataMiner_Agents/Exporting_and_importing_packages_on_a_DMA.md
```diff
@@ -15,8 +15,11 @@ This section consists of the following topics:
...
 > - For more information on how to import and export elements using CSV files, see [Importing and exporting elements](xref:Importing_and_exporting_elements).
 > - It is also possible to export element information to the printer or the clipboard. To do so, follow the same procedure as to export to a CSV file. See [Exporting elements to a CSV file](xref:Importing_and_exporting_elements#exporting-elements-to-a-csv-file).
-> - Backwards compatibility of import packages is not supported. If you import a package created with a more recent DataMiner version into a system with an older version, this may not work.
 > - Exporting/importing logger table data saved in Elasticsearch is only supported from DataMiner 10.1.4/10.2.0 onwards.
+
+> [!CAUTION]
+> - Forwards compatibility of import packages is not supported. If you import a package created with a more recent DataMiner version into a system with an older version, this may not work.
+> - Backwards compatibility of import packages is not completely guaranteed. If you import a package created with an older DataMiner version into a system with a newer version, this will work in the majority of cases.
```
MariekeGO, line 22: I'm not sure if I entirely agree with the choice of terminology here. If we say the "import packages" themselves are not forwards compatible (rather than the importing feature), I'd expect that an older package is not compatible with a more recent system, instead of the other way around. Maybe we should just keep this simple and rephrase both bullet points as follows?

```suggestion
> - Importing a package created with a more recent DataMiner version into a system with an older version is not supported and may not work. Importing a package created with an older DataMiner version into a system with a newer version will work in most cases, but this is not guaranteed.
```

MariekeGO (approved):

PR #1777, opened 2023-07-19
Patch 13

AlexanderGorbunovSkyline:
Major restructuring of the OpenSearch installation guide.
Added:
- Firewall configuration
- Information about mounting the data drive
- additional info about configuring a single node database

Reordered:
- Initial configuration goes with disabled security (same as for Elastic)
- Security section goes after the installation & testing section.

Removed:
- `indices.query.bool.max_clause_count` setting should be set to "2147483647" (I strongly disagree with this recommendation).

Not sure:
- is the setting "http.port: 9200" necessary? I believe it can be commented out.
- what the setting "node.max_local_storage_nodes: 3" is for and whether it can be left as is for a single node.

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
+++ b/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
@@ -78,7 +78,7 @@ Make sure the firewall ports are open for OpenSearch. OpenSearch operates on TCP
 
 ### Mounting the data drive
 
-In most cases, a separate hard drive will be used to store database data. The data drive has to be mounted and path to the data directory has to be specified in the Openseconfiguration file (see below).
+In most cases, a separate hard drive will be used to store database data. The data drive has to be mounted and path to the data directory has to be specified in the parameter *path.data* in *opensearch.yml* configuration file. An example of *opensearch.yml* can be found in the section [Initial configuration](#initial-configuration).
 
 The following example command mounts the drive */dev/sdb1* to the directory */media/data*: `sudo mount /dev/sdb1 /media/data`
 
--- a/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
+++ b/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
@@ -6,289 +6,183 @@ uid: Installing_OpenSearch_database
 
 ## Compatibility
 
-Supported versions:
+OpenSearch is only supported on Linux. DataMiner is compatible with the following versions:
 
 - OpenSearch 1.X
 - OpenSearch 2.X
 
 > [!NOTE]
-> OpenSearch is only supported on Linux.
+> - We promote the use of Ubuntu LTS as the preferred Linux distribution. As such, the commands mentioned below will work on any Debian-based system, including Ubuntu.
 
-## Setting up the OpenSearch cluster
-
-See the [official documentation](https://opensearch.org/docs/latest/) on how to set up your OpenSearch cluster. The configuration is almost identical to that of an Elasticsearch cluster.
-
-> [!IMPORTANT]
->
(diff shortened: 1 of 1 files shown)


AlexanderGorbunovSkyline:
@KevinVerm, @SLC-Laurens-Vercruysse

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -6,115 +6,237 @@ uid: Installing_OpenSearch_database
...
+
+    `$ sudo ufw allow from [IP node DMA 2] to [IP node 2] proto tcp port 9200,9300`  
+
+    `$ sudo ufw allow from [IP node DMA 2] to [IP node 3] proto tcp port 9200,9300`  
+
+  - And so on.
+
+### Mounting the data drive
+
+In most cases, a separate hard drive will be used to store database data. The data drive has to be mounted and path to the data directory has to be specified in the Openseconfiguration file (see below).
```
MariekeGO, line 81: Is "Openseconfiguration" a typo or is this actually the name of a file?

Also, could you replace the "see below" reference with an internal cross-reference? That would make it easier for the user to know which of the sections below you actually mean.

AlexanderGorbunovSkyline: This was a typo. Corrected and added a cross-reference as suggested.

KevinVerm:
@SLC-Laurens-Vercruysse , I recall indices.query.bool.max_clause_count was set to below value in the previous iteration of the help, do you agree we should not recommend or set this in opensearch.yml or should we re-instate this?:

indices.query.bool.max_clause_count setting should be set to "2147483647" (I strongly disagree with this recommendation).

KevinVerm:
Not sure:

is the setting "http.port: 9200" necessary? I believe it can be commented out.

On stackoverflow(https://stackoverflow.com/questions/31677563/elasticsearch-failed-to-connect-to-localhost-port-9200-connection-refused), they do suggest to have it enabled. Couldn't hurt I guess. @SLC-Laurens-Vercruysse what is your take on this?

KevinVerm:
what the setting "node.max_local_storage_nodes: 3" is for and whether it can be left as is for a single node.

Below is the explenation from Google Bard:

The node.max_local_storage_nodes setting in Elasticsearch controls the maximum number of nodes that can share a single data directory. This setting is important for ensuring that the data on each node is unique and that no single node can become a bottleneck for the cluster.

The default value for node.max_local_storage_nodes is 1, which means that each node can only have a single data directory. This is the recommended setting for most production clusters. However, you may need to increase this setting if you have a large amount of data or if you want to improve performance by distributing the data across multiple nodes.

If you set node.max_local_storage_nodes to a value greater than 1, you must make sure that each data directory is on a separate physical disk. This is because Elasticsearch uses the fs.defaultFS setting to determine the location of the data directories. If you have multiple data directories on the same physical disk, Elasticsearch will only use one of them.

The node.max_local_storage_nodes setting is a cluster-wide setting, which means that it applies to all nodes in the cluster. If you change this setting, you must restart all of the nodes in the cluster for the changes to take effect.

Here is an example of how to set the node.max_local_storage_nodes setting in the Elasticsearch configuration file:

node.max_local_storage_nodes: 3
This setting would allow each node in the cluster to have up to 3 data directories.

Perhaps @SLC-Laurens-Vercruysse can share his take on this...

AlexanderGorbunovSkyline:
Found where the recommendation for indices.query.bool.max_clause_count=2147483647 comes from. 
Commit from Robbe De Ghein on 06/02/2023 https://github.com/SkylineCommunications/dataminer-docs/commit/158f9e4775e345c7913d5d2721a9e119375e91cf

Other questions:
* is the setting "http.port: 9200" necessary? - No, it can be commented out in the example config.
* what is the setting "node.max_local_storage_nodes: 3" for? - see https://www.elastic.co/guide/en/elasticsearch/reference/7.17/modules-node.html#max-local-storage-nodes

PR #1780, opened 2023-07-19
Update Migrating_the_general_database_to_Cassandra.md

SLC-Laurens-Vercruysse:
Added information on setting Heap size on Windows. Ideally a separate page should be made "Installing a Cassandra database on a DMA" which outlines that Cassandra is installed with DataMiner and mentions peculiarities about the configuration that we ship with. This would then have a better place there. However, that is tricky to do in the short term.

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_Cassandra.md
+++ b/user-guide/Advanced_Functionality/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_Cassandra.md
@@ -28,22 +28,6 @@ The migration procedure has the following prerequisites:
 
 - The DataMiner Agents did not restart in the last 48 hours.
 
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_Cassandra.md
```diff
@@ -28,6 +28,22 @@ The migration procedure has the following prerequisites:
 
 - The DataMiner Agents did not restart in the last 48 hours.
 
+> [!IMPORTANT]
```
MariekeGO, line 31: I think this warning may not be in the right location. To my knowledge, the wizard to migrate from MySQL to Cassandra is usually also used to install Cassandra itself, which means that it is not possible for the user to configure any Cassandra settings before they start the wizard.

MariekeGO (approved):

PR #1782, opened 2023-07-20
connector version semantics info

PedroDebevere:
Info about connector version semantics, which is taken from https://community.dataminer.services/documentation/protocol-version-numbering-explained/ is now included in docs.


SimonV-Skyline (commented):
the only remarks are related to the same thing, I felt like you were encouraging a bit too much the fact that multiple system versions of the same branch can be maintained. This is still something we want to avoid as much as possible.
Examples:
- if a new system version requires a new firmware version and all devices can update to that new firmware without restrictions, we won't maintain the former system version.
- if a new system version requires a new firmware version and some old hardware can't be updated to the newer firmware, we will (typically) only maintain the former range as long as the older device/firmware are supported by the vendor.

Made a couple of suggestions in that direction already

PR #1787, opened 2023-07-25
Clarification for shared static parameter behavior

Robin-Spruytte:
Added clarification for shared static parameter behavior. Version needs to be the same as well. (Could be added as a note if that would be better)

### DIFF
--- a/develop/devguide/Connector/LogicQActions.md
+++ b/develop/devguide/Connector/LogicQActions.md
@@ -117,7 +117,10 @@ The entry point method must be a public method of a public class.
 
 In the example above, the Run method of the QAction class is defined as a static method.
 
-It is also possible to define static fields in the QAction class. As only one copy of a static member exists, regardless of how many instances of the class are created, static fields are also shared between all the elements executing this protocol (and the same version) and persist as long as the SLScripting process is running.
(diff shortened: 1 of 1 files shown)


develop/devguide/Connector/LogicQActions.md
```diff
@@ -119,6 +119,9 @@ In the example above, the Run method of the QAction class is defined as a static
 
 It is also possible to define static fields in the QAction class. As only one copy of a static member exists, regardless of how many instances of the class are created, static fields are also shared between all the elements executing this protocol and persist as long as the SLScripting process is running.
 
+> [!NOTE]
+> Static fields are only shared between elements with the same DataMiner version.
+
```
EdithLansens, line 124: @Robin-Spruytte I have indeed changed this into a note. Is this OK to merge?

Robin-Spruytte: Hi Edith, I meant protocol version, not DataMiner version. There was some confusion whether the fields are shared between same protocols if the protocol version was different.

EdithLansens: Changed it to protocol version!

PR #1804, opened 2023-08-01
KI: Unable to connect to DMS using .Net Remoting

EdithLansens:


bertbsky:
Functionally OK. I would prefer to see the casing of the term `.Net Remoting` changed to `.NET Remoting` as that is the official product name from Microsoft.

EdithLansens:
@bertbsky, thanks for the feedback! I have changed all instances of .Net remoting in the documentation to .NET Remoting, so that we use one term consistently.

PR #1813, opened 2023-08-03
OpenSearch - remove linux note (now also supports windows)

RobbeDGH:
Since OpenSearch 2.4, windows is also supported

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
+++ b/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
@@ -11,9 +11,6 @@ Supported versions:
 - OpenSearch 1.X
 - OpenSearch 2.X
 
-> [!NOTE]
-> OpenSearch is only supported on Linux.
-
 ## Setting up the OpenSearch cluster
 
 See the [official documentation](https://opensearch.org/docs/latest/) on how to set up your OpenSearch cluster. The configuration is almost identical to that of an Elasticsearch cluster.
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -12,7 +12,7 @@ Supported versions:
 - OpenSearch 2.X
 
 > [!NOTE]
-> OpenSearch is only supported on Linux.
+> OpenSearch is supported on Windows from OpenSearch 2.4 onwards.
```
EdithLansens, line 15: Hi @RobbeDGH, OK to merge it like this?

RobbeDGH: i rather just remove it as I dont see a reason why people would still download 2.3 or lower

RobbeDGH: (context: we are already at 2.9)

spetebroodt: If OpenSearch can only be used on Windows from a certain version onwards, then I agree with Edith that we should mention that "as from" version.

EdithLansens: I agree that it is unlikely for people to download an OpenSearch version prior to 2.9 today. However, it is possible that some users might still be using an older version they had previously installed. Since our documentation explicitly states that we offer support for these older versions, it would be more thorough and user-friendly to retain the note for their benefit.

RobbeDGH: Good for me, feel free to merge

EdithLansens (approved):

PR #1816, opened 2023-08-04
Update configuring elastic backup -network share permissions

meitavelo:


LuisFreitas-Skyline (commented):
All seem to be fine

meitavelo (commented):
Modified both links:
- Link to restrictions for backup
- Link to PSTools

user-guide/Advanced_Functionality/Databases/Indexing_database/Elasticsearch/Backups/Configuring_Elasticsearch_backups_Windows_Linux.md
```diff
@@ -312,3 +312,23 @@ You have now finished configuring an Elasticsearch backup. If you do not want to
...
+
+## Troubleshooting
+
+### Network share access in Windows
+
+Whether you are triggering the backup via DataMiner or using an Elasticsearch REST API client such as Kibana, you need to make sure that the backup path is accessible on all the nodes in the cluster.
+
+If the network path requires credentials, the SYSTEM user must be granted access. You can grant this access as follows:
+
+1. Follow the instructions under [Restrictions for the backup path](xref:Configuring_Elasticsearch_backups_Windows#restrictions-for-the-backup-path).
```
MariekeGO, line 324: Just to be sure, are line 324 and 326 indeed two steps that should be executed one after the other? Or are these in fact two different alternatives to grant the access? If it's the latter, could you adjust this so this is more clear, and perhaps also add why you would use one alternative instead of the other?

MariekeGO (changes requested):
As this is presented as a troubleshooting entry, I'd expect information about what the problem is you are troubleshooting here, since that is where most users will start from when they are looking to solve the problem. What happens if the backup path is not accessible on all nodes in the cluster? Could you add this info?

user-guide/Advanced_Functionality/Databases/Indexing_database/Elasticsearch/Backups/Configuring_Elasticsearch_backups_Windows_Linux.md
meitavelo: They can use one of 2 options.

MariekeGO (approved):

PR #1819, opened 2023-08-08
sending email note

EdithLansens:


user-guide/Advanced_Modules/Automation_module/Automation_script_actions/Email.md
```diff
@@ -19,6 +19,7 @@ Use this action to send a notification via email.
    > [!NOTE]
    >
    > - If you want to specify multiple indices for one table parameter, use a semicolon “;” as separator.
+   > - If you want to specify multiple parameters for one element, service, or protocol version, assign them all within a single line.
```
EdithLansens, line 22: @tmsrosa, should ";" also be used as a separator here, similar to when specifying multiple indices for a table parameter? (see suggested change below)

EdithLansens, line 22:
```suggestion
   > - If you want to specify multiple parameters for one element, service, or protocol version, use a semicolon ";" as separator.
```

HansSolos: It should not.
Purpose of the additional info is to avoid that users create multiple lines for the same element/service/protocol.

Here's an example of how it should:
image

HansSolos: It should not mention the ;. it should mention the original text.

tmsrosa (approved):
Approved, everything looks good (just to be clear, approving the note ending with "assign them all within a single line").

PR #1844, opened 2023-08-11
Add permissions that are required for DIS to properly function

MichielOda:


develop/TOOLS/DIS/Reference/DIS_settings.md
MariekeGO, line 1: Are these user permissions required for DIS in general, or for the DMA tab of the DIS settings? Right now, the place where you've added this suggests the latter, but it isn't clear.

MichielOda: I meant the user permissions on DataMiner itself. I've changed it to 'user permissions' and made it a link to said permissions.

MichielOda: I've moved the text

MariekeGO (approved):

PR #1857, opened 2023-08-20
Workspace buttons colour

EdibSU:
Workspace buttons in the header are in fact white for all themes and not blue. Checked on DataMiner Cube 10.3.0.0 and [here](https://community.dataminer.services/courses/dataminer-operator/lessons/cube-sidebar-2/) at 37:00.


MariekeGO:
As the coloration likely depends on the DataMiner version and theme, and the color info is actually superfluous, I've now simply removed it. "The four squares" should be clear enough to the user.

MariekeGO (approved):

PR #1860, opened 2023-08-21
Update Backing_up_a_DataMiner_Agent_in_DataMiner_Cube.md

KevinVerm:


MariekeGO (changes requested):

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+```txt
+opensearch.service: start operation timed out. Terminating.
+opensearch.service: Failed with result 'timeout'.
+Failed to start OpenSearch.
+opensearch.service: Consumed 57.702s CPU time.
+``
+
+It could be that you may need to increase the start timeout for systemd. To do this, execute below steps:
```
MariekeGO, line 135:
```suggestion
It could be that you may need to increase the start timeout for systemd (see [systemd](https://systemd.io/)). To do this, execute the steps below:
```

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
```diff
@@ -0,0 +1,165 @@
...
+   sudo /bin/systemctl enable opensearch.service
+   ```
+
+1. Start the OpenSearch service again:
+
+   ```bash
+   sudo systemctl start opensearch
+   ```
+
+1. Execute the following command to verify that OpenSearch no longer goes into timeout:
```
SLC-Laurens-Vercruysse, line 161: This only verifies whether the opensearch service is able to remain running and does not stop, not really related to timout.

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
```diff
@@ -0,0 +1,165 @@
...
+   sudo rm -rf nodes
+   ```
+
+1. Restart OpenSearch on the cluster manager first and then on the data nodes.
+
+> [!IMPORTANT]
+> Deleting the nodes folder will result in loss of data. You should only do so with a new installation of OpenSearch.
+
+## SLSearch.txt logging mentions OpenSearch version is not officially supported
+
```
SLC-Laurens-Vercruysse, line 109: We currently support 1.x and 2.x. Maybe add info in case Opensearch 3.x is released (futureproofing)

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
SLC-Laurens-Vercruysse: Though in the scope of the procedure it is the correct thing to check, maybe we should expand a bit to say this is just to know if the service is running and remains so.

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
```diff
@@ -0,0 +1,165 @@
...
+If the *SLSearch.txt* log file mentions that the OpenSearch version is not officially supported, you can resolve this by upgrading your DMS to DataMiner 10.3.6 or higher.
+
+However, note that this has no functional impact, as the DMA will run fine even if you have not upgraded yet.
+
+## OpenSearch service going into timeout
+
```
SLC-Laurens-Vercruysse, line 115: Maybe a bit more context? This happens when starting the service on the Linux machine and is one of I assume several issues that may occur when attempting to start the service. 
I haven't ran into this yet, but it might be interesting to know why it is timing out, that might be beyond the scope of this article though.

SLC-Laurens-Vercruysse (changes requested):
Requested some minor changes. All in all looks good.

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
KevinVerm: Doubt if we should list specific versions here... It is supported so. Unless 3.0 would do something different regarding behavior.

user-guide/Troubleshooting/Procedures/Investigating_OpenSearch_Issues.md
KevinVerm: Not sure why suddenly one node stopped working, but we needed to get it up and running. Quick googling showed ElasticSearch sometimes showcased the same problem.

MariekeGO (approved):

PR #1862, opened 2023-08-21
stepper docs

TVincke98:


TVincke98:
> @TVincke98 From which DataMiner version onwards can users use the stepper component? I can't find any release notes on the stepper component yet for reference.

We're aiming to make this component available from 10.3.10 onwards.

EdithLansens (approved):

PR #1864, opened 2023-08-22
AzureAD - Make objectId and Application.Read.All optional

RobbeDGH:
Goes together with [RN37162 - AzureAD - Don't generate RTE when ObjectID...](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=37162)

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
@@ -213,7 +213,7 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
      > Do not use the *Object ID* under *Azure Active Directory > Enterprise applications > [your application name]*. This is a different Object ID, which will not work.
 
      > [!NOTE]
-     > From DataMiner 10.3.11/10.4.0 onwards the **Object ID** is optional
+     > From DataMiner 10.3.11/10.4.0 onwards<!-- RN 37162 -->, the **Object ID** is optional.
 
    - **Client Secret**: In the pane on the left, click *Certificates & secrets*.
 
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
@@ -205,12 +205,9 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
 
 1. Gather the following information:
 
-   - **Client ID**, **Object ID**, and **Tenant ID**: These GUIDs identify the application (DataMiner) in the Azure AD platform, and identify the users & groups directory on the Azure portal, respectively. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
+   - **Client ID** and **Tenant ID**: These GUIDs identify the application (DataMiner) in the Azure AD platform, and identify the users & groups directory on the Azure portal, respectively. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
 
      Creating an Enterprise Application will also create an app registration with the same name, but you will not find it under *owned application*.
-
-     > [!IMPORTANT]
-     > Do not use the *Object ID* under *Azure Active Directory > Enterprise applications > [your application name]*. This is a different Object ID, which will not work.
 
    - **Client Secret**: In the pane on the left, click *Certificates & secrets*.
 
@@ -230,9 +227,16 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
       ```xml
       <AzureAD
        tenantId="[GUID]"
-       objectId="[GUID]"
        clientId="[GUID]"
-       clientSecret="[the DataMiner application secret value]"
+       clientSecret="[the DataMiner application secret value]" />
+      ```
+
+      or
+
+      ```xml
+      <AzureAD
+       tenantId="[GUID]"
+       clientId="[GUID]"
        username="[username]"
        password="[password]" />
       ```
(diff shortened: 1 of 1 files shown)


RobbeDGH:
Also split up the small xml snippet because either the Username & Password are required or a ClientSecret but never the combination.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -205,13 +205,10 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
 
 1. Gather the following information:
 
-   - **Client ID**, **Object ID**, and **Tenant ID**: These GUIDs identify the application (DataMiner) in the Azure AD platform, and identify the users & groups directory on the Azure portal, respectively. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
+   - **Client ID** and **Tenant ID**: These GUIDs identify the application (DataMiner) in the Azure AD platform, and identify the users & groups directory on the Azure portal, respectively. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
```
MariekeGO, line 208: If we change the text like this, it will only be valid for the latest version of DataMiner; however, the documentation should always be valid for all supported versions. To solve this, it would be best to add version info here, for example:

```suggestion
   - **Client ID**, **Object ID** (prior to DataMiner 10.3.10/10.4.0 only), and **Tenant ID**: These GUIDs identify the application (DataMiner) in the Azure AD platform, and identify the users & groups directory on the Azure portal, respectively. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
```

Note that the release note you linked to has not been merged to DataMiner yet, so I just guessed the possible version in the example above.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -230,9 +227,16 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
       ```xml
```
MariekeGO, line 227: To make sure this example is valid for all supported versions, it would be better to change this to two examples, one prior to the change, one from that DataMiner version onwards.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -269,6 +271,26 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
 
    It is now possible to log in to DataMiner with any of the Azure AD user accounts you have added, using either the domain and username (DOMAIN\\user) or the email address.
 
+#### Configuring DataMiner to check if the client secret is about to expire
+
+From DataMiner 10.3/10.2.9 onwards, it is possible to configure DataMiner to check Azure Active Directory for the expiration of the client secret when authenticating (only when application querying is used). When the client secret will expire in less than a week, a warning will be shown in the alarm console.
```
MariekeGO, line 276:
```suggestion
From DataMiner 10.2.9/10.3.0 onwards, it is possible to configure DataMiner to check Azure Active Directory for the expiration of the client secret when authenticating (only when application querying is used). When the client secret will expire in less than a week, a warning will be shown in the Alarm Console.
```

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -269,6 +271,26 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
...
+
+For this 2 extra configurations are required:
```
MariekeGO, line 278:
```suggestion
The following additional configuration is required for this:
```

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -269,6 +271,26 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
...
+
+1. On Azure's side an extra permission: Microsoft Graph \> Application.Read.All – Application – Read applications
```
MariekeGO, line 280:
```suggestion
1. In Azure, make sure the following permission is enabled: *Microsoft Graph* \> *Application.Read.All – Application – Read applications*.
```

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_external_authentication_via_an_identity_provider_using_SAML.md
```diff
@@ -269,6 +271,26 @@ Once you have established a trust relationship between DataMiner (i.e. the servi
...
+1. On DataMiner's side an extra guid in DataMiner.xml: the **Object ID**. This GUID can be found on the same page where the **Client ID** and **Tenant ID** are found. You can find these fields on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
```
MariekeGO, line 281:
```suggestion
1. In DataMiner.xml, add the **Object ID**. Like the **Client ID** and **Tenant ID**, you can find this GUID on the root page of the application: *Azure Active Directory > App registrations > [your application name]*.
```
Note that this step may need to be expanded a bit, as to my knowledge this will also require a DataMiner restart.

MariekeGO (changes requested):
I've added a number of comments for things that need to change. My main concern is that the text currently does not take the different DataMiner versions into account, but only provides the info for the latest version. Could you make sure that users will find the correct information they need, regardless of which of the supported DataMiner versions they are using?

RobbeDGH:
I think its easier to revert the original propsal and add a note that mentions the object id being optional from certain versions onwards

MariekeGO (approved):

PR #1870, opened 2023-08-23
Update NATS Docs

LaurensVergote:


user-guide/Troubleshooting/Procedures/Investigating_NATS_Issues.md
```diff
@@ -119,14 +119,16 @@ A few details to pay attention to:
 
 This is an example of **nats-server.config** in a NATS cluster of 3 or more Agents:
 
+##### [Up to DataMiner 10.2.0(CU5)/10.2.7](#tab/NATS3PlusObsolete)
```
LaurensVergote, line 122: these tabs all show up in the same block (5 together instead of 3+2) and all non-codeblock text after the first tab is gone.

EdithLansens:
@LaurensVergote, the tabs should now be working. I have also changed the order of the tabs, as it is custom to place the instructions for the most recent version of DataMiner first and those of the oldest version last.

EdithLansens (approved):

PR #1873, opened 2023-08-23
Add bitRateTracking to MaintenanceSettings documentation

FlorisCO:
Don't know when this option was introduced, but it has been mentioned in TBD item 155, which was created in 2014.

### DIFF
--- a/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
+++ b/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
@@ -359,9 +359,19 @@ In this *MaxSize* tag, you can specify the maximum log file size.
 
 Default: 10 MB
 
-### Network
+### Network.bitRateTracking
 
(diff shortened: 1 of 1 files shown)


user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
```diff
@@ -75,6 +75,7 @@ This is an example of a *MaintenanceSettings.xml* file:
       <SkipRTE process="SLDataMiner.exe" thread="Database cleaning" />
     </FailoverOnRTE>
   </WatchDog>
+  <Network bitRateTracking="true">
```
EdithLansens, line 78: Is Network not missing a closing tag? Feel free to reject this suggestion if this is not the case :)

```suggestion
  </Network bitRateTracking="true">
```

FlorisCO: Hi Edith, you are right, I did forget the closing tag. However, it is not an open-close pair, so the tag goes at the end, pushed it as a new commit since I can't edit yours.

EdithLansens (approved):

PR #1874, opened 2023-08-23
Improve RawType, Type and Others

SkylineThomasCR:


develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -15,41 +26,28 @@ Specifies which type of content is allowed in the parameter.
 [Interprete](xref:Protocol.Params.Param.Interprete)
 
 ## Remarks
-
-If an incoming parameter value does not match the expected rawtype, DataMiner will not process it. The rawtype setting can therefore be considered as a kind of filter.
-
-Contains one of the following predefined values, detailed in the sections below:
+Before DataMiner 9.5,
```
Skyline-NicolasVC, line 29: And now? Don't leave me hanging with this cliffhanger...

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -58,46 +56,36 @@ The following characters and digits:
...
-```xml
-<RawType>other</RawType>
-```
+any characters, both alphanumeric and non-alphanumeric.
 
 ### signed number
 
 When a next param is defined in Interprete.Type, the result will be the same as with unsigned number. When a fixed length is defined in Interprete.Type, the first half of the values will be positive number, and second half will be negative number.
 
+todo sentence below is confusing and incorrect when using different lengths
```
Skyline-NicolasVC, line 78: The sentence below "Number between -128 and 127." is confusing? Or did you mean the one above?
Also... ToDo: the ToDo

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -111,26 +99,19 @@ Text consisting of the following characters:
...
 
 ### unsigned number
 
-DataMiner will convert the hex value to a number using a decimal conversion.
+DataMiner will convert the value to a number using a decimal conversion.
 
-Example: If the hex value is 31, the decimal conversion gives 49. The parameter will contain 49 (result will be the same as if you use your calculator to convert from hex to dec).
+Example: If the binary value is 0x31, the decimal conversion gives 49. The parameter will contain 49 (result will be the same as if you use your calculator to convert from hex to dec).
 
+todo sentence below is confusing and incorrect when using different lengths
```
Skyline-NicolasVC, line 111: ToDo: ToDo

Todotodotodoo todo todoooo tododo :)

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -111,26 +99,19 @@ Text consisting of the following characters:
...
 Numbers between 0 and 255.
```
Skyline-NicolasVC, line 112: Add list of options?
Length 1: 0 - 255
Length 2: 0 - 65535
Length 4: 0 - 4294967295
Do we support length 8?

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -58,46 +56,36 @@ The following characters and digits:
...
-<RawType>other</RawType>
-```
+any characters, both alphanumeric and non-alphanumeric.
 
 ### signed number
 
 When a next param is defined in Interprete.Type, the result will be the same as with unsigned number. When a fixed length is defined in Interprete.Type, the first half of the values will be positive number, and second half will be negative number.
 
+todo sentence below is confusing and incorrect when using different lengths
 Numbers between -128 and 127.
```
Skyline-NicolasVC, line 79: Give a list of options?
Length 1: -128 - 127
Length 2: -32768 - 32767
Length 4:-2147483648 - 2147483647
Do we support length 8?

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -58,46 +56,36 @@ The following characters and digits:
...
+### other
 
-```xml
-<RawType>other</RawType>
-```
+any characters, both alphanumeric and non-alphanumeric.
 
 ### signed number
 
 When a next param is defined in Interprete.Type, the result will be the same as with unsigned number. When a fixed length is defined in Interprete.Type, the first half of the values will be positive number, and second half will be negative number.
```
FlorisCO, line 76: I was really confused when I read "the first half" for the first time. Why not just specify the way it is, because 0 to UINT.Max doesn't map on INT.MIN to INT.MAX. With signed, if the most significant bit is high, the value will be negative according to the ones' complement: https://en.wikipedia.org/wiki/Signed_number_representations#Ones'_complement

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
```diff
@@ -58,46 +56,36 @@ The following characters and digits:
...
 
+todo sentence below is confusing and incorrect when using different lengths
 Numbers between -128 and 127.
 
-In the following example, the value -3 will be saved in database as FDFFFFFF and the value 3 will be 03000000:
-
-```xml
-<RawType>signed number</RawType>
-```
+The value -3 will be saved in database as FDFFFFFF and the value 3 will be saved as 03000000:
```
FlorisCO, line 81: Perhaps specify how we arrive at these "weird" values, it's because the integer value is converted to a byte array before being converted into a hexadecimal notation. The byte array conversion is subject to the system's endianness, which is little-endian for x86-32 and x86-64 CPUs, which DataMiner is currently always running on.

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.RawType.md
FlorisCO: I'm getting lost a bit, because interprete.length plays a significant role when your interprete.type is double. Namely, if it is 4, we use float, and with 8, we use double internally. Even though it is just stored as a set of bytes...

MariekeGO:
Closed because this is a work in progress that will not be ready any time soon. A new PR will need to be created when this is ready for review.

PR #1875, opened 2023-08-24
Added stepper tutorial

WoutMH:


MariekeGO (changes requested):
This currently fully overlaps with the existing explanation about the [stepper component in the user guide](https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.html), which is something we really want to avoid (not update-friendly). I see two ways we could deal with this: 

- Improve the existing page with any information and screenshots it needs to be more clear, and not add the tutorial.
- Change the tutorial to a much more specific example (e.g. creating a specific app that includes a stepper), and refer to the user guide for the steps that are already explained there. This way the tutorial has added value, and we don't risk that one page gets updated and the other remains outdated, causing contradicting information in our documentation.

WoutMH:
Changed the tutorial to me more specific example using a package from the catalog. @MariekeGO

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper/StepperTutorial.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Click the pencil icon next to the initial page of the application in order to edit it.
+
+1. Hover the mouse pointer over the bar with the "+" on the left until it expands to show the Visualizations pane, and add a [stepper](xref:DashboardStepper), a [form](xref:DashboardForm), and a [table](xref:DashboardTable) component by dragging them onto the page.
+
+   You can find the stepper and form components in the *General* category and the table in the *Tables* category.
+
+The end result of this step should look like this:
+
+![Components](~/user-guide/images/StepperComponents.png)
```
MariekeGO, line 43: This screenshot and the one for the next step were taken from a fairly large window, but as the image gets resized to fit in the docs layout, the text on docs will be very small and hard to read. Would it be possible to replace these with screenshots of smaller windows, so that the user will be able to see everything more clearly?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper/StepperTutorial.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Step 3: Add the data
+
+Add data to components by [applying a data feed](xref:Configuring_dashboard_components#applying-a-data-feed).
+
+Each of the components on the page has to receive some data:
```
MariekeGO, line 49: I think it would be useful if there were screenshots clarifying how this data is added, especially for creating a query and linking a feed from a table component to other components. Could you add these?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper/StepperTutorial.md
```diff
@@ -0,0 +1,91 @@
...
+
+To do so, use the *Appearance* setting in the component layout tab on the right.
+
+![Appearance presets](~/user-guide/images/StepperAppearance.png)
+
+## Step 6: Use the low-code app
+
+Now that everything is configured, you can publish and use your app.
+
+You can create new incidents using the form component and the save button in the header bar. You can inspect and update existing incidents by selecting them in the table component at the bottom.
```
MariekeGO, line 89: Could you make it more clear in the description of this final step what the purpose of the stepper component is in the app?

MariekeGO (changes requested):

PR #1878, opened 2023-08-25
Typo

EdibSU:
Added space after the first sentence and removed extra dot from the end of the second sentence.

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.Exceptions.Exception.md
+++ b/develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.Exceptions.Exception.md
@@ -4,7 +4,7 @@ uid: Protocol.Params.Param.Interprete.Exceptions.Exception
 
 # Exception element
 
-Defines an exception for an exception value you want to intercept of which the Interprete.Type is identical to that of the parameter in question..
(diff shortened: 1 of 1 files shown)


PedroDebevere (approved):

develop/schemadoc/Protocol/Protocol.Params.Param.Interprete.Exceptions.Exception.md
```diff
@@ -4,7 +4,7 @@ uid: Protocol.Params.Param.Interprete.Exceptions.Exception
 
 # Exception element
 
-Defines an exception for an exception value you want to intercept of which the Interprete.Type is identical to that of the parameter in question..
+Defines an exception for an exception value you want to intercept of which the Interprete. Type is identical to that of the parameter in question.
```
PedroDebevere, line 7:
```suggestion
Defines an exception for an exception value you want to intercept of which the [Interprete.Type](xref:Protocol.Params.Param.Interprete.Type) is identical to that of the parameter in question.
```

MariekeGO, line 7: "Interprete.Type" is in fact correct, as this is the way the Type subtag of the Interprete tag is indicated. I'll revert that change but will keep the other correction.

PR #1882, opened 2023-08-28
Changes to Stepper component doc

MariekeGO:
Includes corrections + some changes based on internal presentation to make the text more user-friendly.

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
@@ -19,7 +19,7 @@ To configure the component:
      > [!NOTE]
      > DOM definitions can only be applied to the stepper component within the Low-Code Apps module.
 
-   - When a DOM instance is applied, this will be used to display the complete history of each state the instance has traversed until its current state. In this case, the happy path is displayed based on the current state, which may not be the most happy path based on the DOM definition.
+- When a DOM instance is applied, this will be used to display the current state, preceded by the states the instance has traversed until its current state, and followed by the future states it will go through if it follows the normal flow.
 
 1. Fine-tune the component layout. In the *Component > Layout* tab, the following options are available:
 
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
```diff
@@ -4,23 +4,25 @@ uid: DashboardStepper
...
+   You can apply a DOM definition or DOM instance directly, or use e.g. a feed based on a table component, so that the stepper will display the information linked to the item selected in the table component.
+
+   - When a DOM definition is applied, this will be used to display states along the typical path of the definition, i.e. the "happy path". This path illustrates the states an instance would undergo following the standard workflow.
 
      > [!NOTE]
      > DOM definitions can only be applied to the stepper component within the Low-Code Apps module.
 
-   - When applying a DOM instance, the component will display the complete history of each state the instance has traversed until its current state. Based on the current state, it calculates and displays the remaining states along the happy path.
+   - When a DOM instance is applied, this will be used to display the complete history of each state the instance has traversed until its current state. In this case, the happy path is displayed based on the current state, which may not be the most happy path based on the DOM definition.
```
TVincke98, line 22: > which may not be the most happy path based on the DOM definition.

I find this a little confusing, the happy path based on the current state of an instance will never be the same as the happy path you see for a definition, it will be a subset of that path where the start of the path isn't the first possible state but the current state the instance is in.

TVincke98: The difference is that, since there isn't a current state, future states and past states in the case of a definition, it will always show the same happy path, i.e. the easiest path between the first possible state and the last possible state. Since an instance does have a current state and the happy path is the easiest path between the current state and the last possible state, this happy path will never be the same as the happy path shown for a definition. I think the second explanation is the best, but maybe you can say something like 'followed by the future states the instance will go through if it were to follow the normal flow from here'.

MariekeGO: How about this? 

```suggestion
- When a DOM instance is applied, this will be used to display the current state, preceded by the states the instance has traversed until its current state, and followed by the future states it will go through if it follows the normal flow.
```

PR #1884, opened 2023-08-28
Ubuntu LTS

EdithLansens:
Small change following [Dojo question](https://community.dataminer.services/question/which-version-of-ubuntu-to-use/)

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
+++ b/user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
@@ -14,6 +14,8 @@ Supported versions:
 > [!NOTE]
 > OpenSearch is supported on Windows from OpenSearch 2.4 onwards.
 
+We promote the use of Ubuntu LTS as the preferred Linux distribution. The procedure mentioned below will work on any Debian-based system, including Ubuntu.
+
 ## Setting up the OpenSearch cluster
 
 See the [official documentation](https://opensearch.org/docs/latest/) on how to set up your OpenSearch cluster. The configuration is almost identical to that of an Elasticsearch cluster.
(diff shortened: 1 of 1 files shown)


SkylineCookieMonster (approved):

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -30,6 +30,9 @@ See the [official documentation](https://opensearch.org/docs/latest/) on how to
 
 Below you can find an example of how to set up an OpenSearch cluster on premises on a Ubuntu system, in accordance with the [Debian installation guide](https://opensearch.org/docs/latest/install-and-configure/install-opensearch/debian/).
 
+> [!NOTE]
+> We recommend using the latest version of Ubuntu LTS as Linux distribution.
```
MariekeGO, line 34: I'm not entirely sure if this is the best place for this info, as it might be useful beyond the example. Perhaps we could move it to the compatibility section, and add that while this is the Linux distribution we've tested, OpenSearch also supports other Linux distributions (based on the comments below this question and answer: https://community.dataminer.services/question/elastic-application-or-open-search-is-better-for-dataminer/)?

SkylineCookieMonster: Hi Marieke,
Sure we can add it in a more general section for example in the compute requirements (https://docs.dataminer.services/user-guide/Reference/DataMiner_Compute_Requirements.html). There we are still missing OpenSearch. Indeed OpenSearch and Cassandra can be installed on a wide variety of Linux distros, but we advise using the latest Ubuntu LTS (for this we provide installation steps and we will also provide an installation script soon). RHEL systems might be supported, but we still see it as a nice to have and not MVP for the installer script ATM

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -14,6 +14,8 @@ Supported versions:
 > [!NOTE]
 > OpenSearch is supported on Windows from OpenSearch 2.4 onwards.
 
+We promote the use of Ubuntu LTS as the preferred Linux distribution. The procedure mentioned below will work on any Debian-based system, including Ubuntu.
```
MariekeGO, line 17: To make sure people can find more detailed info, I'd rephrase this as follows with a link:

```suggestion
We recommend using the latest version of Ubuntu LTS. However, OpenSearch also supports other Linux distributions. See [Operating system compatibility](https://opensearch.org/docs/latest/install-and-configure/install-opensearch/index/#operating-system-compatibility).
```

Maybe we should also add this in the note above, instead of adding it as a simple paragraph?

PR #1911, opened 2023-09-04
Update I-DOCSIS_Create_CCAP_CM_pair.md

EstebanSkyline:
Updating documentation for I-DOCSIS_Create_CCAAP_CM_Pair with explanation on how to create CCAPs and Collectors using a CSV file


user-guide/Standard_Apps/EPM/EPM_I-DOCSIS/I-DOCSIS_Create_CCAP_CM_pair.md
MariekeGO, line 1: Just to be sure, is it correct that the same script is used, regardless of whether you start from scratch or start from a CSV file? It seems a bit odd that in the first procedure we mention different steps in the wizard, while in the procedure from CSV there seems to be only one step: fill in the CSV path and run automation.

EstebanSkyline: Hi, Yes! it is correct what you said, in the team we discussed about it and it was better for us to add this functionality to the same script instead to have a different script with the same purpose, create CCAPs and Collectors automatically.

MariekeGO:
Includes KPIs from RN 37301.

PR #1919, opened 2023-09-05
DIS - Adding error messages and updating toc

MichielOda:


MariekeGO:
With the way this has been added to the documentation structure now, there really isn't much context for the user as to what these pages are for, and I'm afraid these may cause some confusion as a consequence – especially because of the location where they have been added right under the main DIS node. I see several possible ways we could solve this:

- Make a page for the MajorChangeChecker and Validator nodes with more information about what all the underlying pages are for.
- Move the MajorChangeChecker and Validator nodes  to be under a new top-level node in the DIS folder with more context information about the function of all of these pages. 
- Move the MajorChangeChecker and Validator nodes  to be under the DIS reference node.

What do you think?

MichielOda:
After discussing this internally, we'll leave this be for now. As it's adding almost a 1000 pages, this would also bring us closer to the limit again. 
In DIS we won't have functionality yet to make use of this so no need to have this merged. We'll create the PR again whenever we pick up that topic so that DIS can make proper use and check on how to handle the file limitations for Azure.

PR #1937, opened 2023-09-09
Update UIComponentsLabel.md

EdibSU:
* Removed obsolete tag Param/Information/Includes.
* Removed whitespaces from Param/Name tag, as this generates warning in Validator.
* Added text to Param/Information/Subtext tag, as this generates minor error in Validator.
* Removed Param/Information tag in *write* param for brevity. As per note on [UI Components Visualization](https://docs.dataminer.services/develop/devguide/Connector/UIComponentsVisualization.html) maybe it would make sense to remove Param/Information and Param/Display tags entirely in following examples: [Checkbox](https://docs.dataminer.services/develop/devguide/Connector/UIComponentsLabel.html#creating-an-editable-label-with-check-box), [Editable lable](https://docs.dataminer.services/develop/devguide/Connector/UIComponentsLabel.html#creating-an-editable-label), and CPU Load example.
* "It is possible to visualize a check box by providing a discrete entry with state disabled on the write parameter" indicates that state needs to be set to disabled in order to visualize a checkbox, this is not required.

### DIFF
--- a/develop/devguide/Connector/UIComponentsLabel.md
+++ b/develop/devguide/Connector/UIComponentsLabel.md
@@ -14,12 +14,6 @@ A label displays the current value of a standalone parameter. The text provided 
   <Description>Software Version</Description>
   <Information>
      <Subtext>Running software version.</Subtext>
-     <Includes>
-        <Include>time</Include>
-        <Include>range</Include>
-        <Include>steps</Include>
-        <Include>units</Include>
-     </Includes>
   </Information>
   <Type>read</Type>
   <Interprete>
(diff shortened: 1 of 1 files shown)


EdibSU:
Regarding visualization of checkbox it seems that if Param/Measurement/Type is string or number `state="disabled"` is not required in order to visualize checkbox, however if Param/Measurement/Type is discreet it is. In the case where Param/Measurement/Type is string or number at least one Param/Measurement/Discreets/Discreet is required in order to visualize checkbox.

MariekeGO (approved):

PR #1972, opened 2023-09-20
PDF report: private dashboards

EdithLansens:
I've created this pull request in response to a [Dojo question](https://community.dataminer.services/question/minimum-dashboard-rigths-for-scheduler/). A private dashboard will not appear in the dropdown list of available dashboards when using the action to upload a report to FTP or a shared folder.

Is this correct and OK to merge, @SebastiaanSL?


SebastiaanSL (approved):
This is indeed accurate. Do note that this is something which we've heard multiple times. So a change in the software in the future is possible.

PR #1994, opened 2023-09-25
Connector help info + template

MariekeGO:


SimonV-Skyline (approved):

develop/devguide/Connector/ConnectorHelp/Connector_help_pages.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Markdown syntax
+
+Use DocFX Flavored Markdown (DFM). See [Markdown syntax](xref:contributing#markdown-syntax).
+
+### Title
+
+The **title** of the help page must be the **exact name of the connector**.
+
+It is important that this is the same name as in the [DataMiner Catalog](https://catalog.dataminer.services/), so that the help page can automatically be displayed there.
```
jaklke, line 23: Having the title being exactly the same is not needed anymore to embed it into catalog. 
Instead, the following is mandatory regarding the .md file: 
1. it should be under /dataminer-docs-connectors/connector_help/connector-help
2. the file name should be the name of the connector where the spaces are replaced with underscore
The above two points probably don't fit under the Title section, but I do think we need to mention this somewhere on this page.

MariekeGO: I've changed this so the points you mentioned are now included.

jaklke (approved):
Perfect,  good for me!

PR #2016, opened 2023-09-28
Note on migration elements

MariekeGO:
Based on https://community.dataminer.services/question/element-migration-mismatch-in-the-number-of-elements/.


HansSolos (approved):
FYI: DELT migration will eventually be obsolete when swarming feature is released.
But for now, it's useful to have this information for the user.

PR #2026, opened 2023-09-28
mergeDBConfig

DieterDegrande:
Add changes for moving DBConfiguration.xml into db.xml

Note: doc changes for RN37446


MariekeGO:
> If a single search cluster is defined, it should work as normal. If multiple search clusters are defined (now in the DB.xml), there will probably be an issue still, the Cloud team will follow this up. Regardless, that clause needs adjusting so that only the case of multiple search clusters defined in the DB.xml doesn't work yet.

OK, I've asked the Cloud team to look into this, and in the meantime this PR can already be merged.

MariekeGO (approved):

PR #2033, opened 2023-09-29
Update Software_support_life_cycles.md

SeppeDejonckheere:
Add section on OS support

### DIFF
--- a/user-guide/Reference/Software_support_life_cycles.md
+++ b/user-guide/Reference/Software_support_life_cycles.md
@@ -61,3 +61,7 @@ Status overview:
 - **End of Engineering**: Feature will no longer be updated, but will still be supported.
 - **End of Support**: Feature will no longer be updated and no longer be supported.
 - **End of Life**: Feature or compatibility will be removed.
+
+### Operating System support
+
+Skyline provides support for all supported DataMiner versions that run on an operating system that is supported by Microsoft.


SeppeDejonckheere:
Not sure if my wording is the best. It's kind of difficult to write a sentence about 3 kinds of support

user-guide/Reference/Software_support_life_cycles.md
```diff
@@ -61,3 +61,7 @@ Status overview:
 - **End of Engineering**: Feature will no longer be updated, but will still be supported.
 - **End of Support**: Feature will no longer be updated and no longer be supported.
 - **End of Life**: Feature or compatibility will be removed.
+
+### Operating System support
+
+Skyline provides support for all supported DataMiner versions that run on an operating system that is supported by Microsoft.
```
MariekeGO, line 67: Instead of adding this as a separate section, I'd propose to add it as a note under "Third-party software support life cycle":

```suggestion
> [!NOTE]
> For all supported DataMiner versions, we support all Windows versions that Microsoft currently supports.
```

SeppeDejonckheere: I agree that a note is the better option here and the suggested change looks good to me.

MariekeGO (approved):

PR #2056, opened 2023-10-04
Known Issue Central Offload

miguel-slc:
Adding an extra know issue related to the offload database:
Information available in https://intranet.skyline.be/DataMiner/Lists/Known%20Bugs/DispForm.aspx?ID=171


user-guide/Troubleshooting/Known_issues/KI_offload_database_incorrect_integer_value.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Offload to MySQL offload database fails
+
+## Affected versions
+
+Any version of DataMiner
+
+## Cause
+
+If a MySQL offload database is used, the default configuration is incorrect: The option *STRICT_TRANS_TABLES* is enabled (**sql_mode**).
```
MariekeGO, line 13: Could you check if my interpretation here is correct? Also, I feel like we should explain why this option must not be enabled. Could you add this?

miguel-slc: Hi Marieke, yes, your interpretation is correct.
About the explanation of this option:
The option STRICT_ALL_TABLES is used to return an error when there are missing values in  data-change statements such as INSERT or UPDATE. If this option is enabled, MySQL will return an error and ignore the remaining rows from the statement. Since there could be a possibility that the data to be offloaded contains missing data (e.g. a data point that was not stored due to an error in the element), we would still like to offload the remaining rows to the offload database.

MariekeGO: I've added more info based on your feedback. Could you check if this is OK?

PR #2057, opened 2023-10-04
Update Cube_10.3.11.md

jaklke:
Clarified when this could occur.  If accepted, can this change also be ported to main releases, please?


jaklke (commented):
Thanks for providing the clear description! Good to merge for me

spetebroodt (approved):

PR #2058, opened 2023-10-04

ThibaultHeylen:
Moved a configuration to the correct place on the page. You can only set the value after installing Opensearch, can be confusing to note it before you have done this


user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
MariekeGO, line 1: Because the place where you put this note now is below an example, I'm worried people won't notice it. Perhaps we could keep the original note at the top, and then put a reminder below the example as you suggest?

ThibaultHeylen: That's also an option. To me it was confussing that we say to change a value in a file, while the file is not created yet. Seems like people will go over it, not find it, then continue and forget about it later when the file is actually created.

MariekeGO (approved):

PR #2070, opened 2023-10-06
RN 37233

EdithLansens:
Hi @TVincke98, could you please review this pull request to confirm the accurate inclusion of release note 37233 in the user guide?

I am a bit unsure whether this release note introduces a noticeable change for users. In DataMiner 10.3.10, a path was calculated similar to a happy path, and now, from DataMiner 10.3.11/10.4.0 onwards, it is an actual happy path. If this isn't something users will notice as a difference between 10.3.10 and 10.3.11, it may be better to replace the text by one standard line: _"Instances without a history display the happy path from the initial state to the current state as if they followed the standard flow."_

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
@@ -23,7 +23,7 @@ To configure the component:
 
      > [!NOTE]
      >
-     > - If the DOM history saving behavior of a module has been set to *Disabled* in [the `DomInstanceHistorySettings` object](xref:DOM_DomInstanceHistorySettings), the stepper component will not show a history of traversed states for any DOM instances. From DataMiner 10.3.11/10.4.0 onwards<!--RN 37233-->, instances without a history calculate the happy path from the initial state to the current state and then to the final state. In DataMiner 10.3.10, instances without a history calculate the happy path from the current state to the final state.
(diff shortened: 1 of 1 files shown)


TVincke98:
Hi Edith, the difference between the behaviour before and after RN37233 is that before, when an instance had no available history, all that was shown was the path from the current state to the last possible state of that instance. After this change (and only when no history is available), a path is shown from the first possible state to the current state, followed by the path to the final state that was already shown before. Both the path from the first to the current and the path from the current to the final state are happy paths, and are calculated in a similar way.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Stepper.md
```diff
@@ -23,7 +23,7 @@ To configure the component:
 
      > [!NOTE]
      >
-     > - If the DOM history saving behavior of a module has been set to *Disabled* in [the `DomInstanceHistorySettings` object](xref:DOM_DomInstanceHistorySettings), the stepper component will not show a history of traversed states for any DOM instances. Instances without a history calculate the path from the initial state to the current state as if they followed the standard flow, similar to a happy path.
+     > - If the DOM history saving behavior of a module has been set to *Disabled* in [the `DomInstanceHistorySettings` object](xref:DOM_DomInstanceHistorySettings), the stepper component will not show a history of traversed states for any DOM instances. From DataMiner 10.3.11/10.4.0 onwards<!--RN 37233-->, instances without a history display the happy path from the initial state to the current state. In DataMiner 10.3.10, instances without a history calculate the path from the initial state to the current state as if they followed the standard flow, similar to a happy path.
```
EdithLansens, line 26:
```suggestion
     > - If the DOM history saving behavior of a module has been set to *Disabled* in [the `DomInstanceHistorySettings` object](xref:DOM_DomInstanceHistorySettings), the stepper component will not show a history of traversed states for any DOM instances. From DataMiner 10.3.11/10.4.0 onwards<!--RN 37233-->, instances without a history calculate the happy path from the initial state to the current state and then to the final state. In DataMiner 10.3.10, instances without a history calculate the happy path from the current state to the final state.
```

Is this more accurate then, @TVincke98?

TVincke98 (approved):

MariekeGO (approved):

PR #2086, opened 2023-10-10
Update requirements when removing field descriptors from a section definition (RN 37395)

slc-peter-vanpoucke:
References: [RN37395] [DCP219482]


Skyline-ThomasGH (approved):

MariekeGO:
I get the impression that with this PR, the behavior from before RN 37395 is no longer mentioned; however, the documentation must be valid regardless if which supported DataMiner version a user is using, so this may still need to be adjusted.

slc-peter-vanpoucke:
> I get the impression that with this PR, the behavior from before RN 37395 is no longer mentioned; however, the documentation must be valid regardless if which supported DataMiner version a user is using, so this may still need to be adjusted.

Indeed, forgot about that. I'll update the change.

Skyline-ThomasGH (approved):

MariekeGO (approved):

PR #2092, opened 2023-10-10
RN 37360

EdithLansens:
Hi @wimme, could you please review this pull request to confirm the accurate inclusion of release note 37360 in the user guide?


SebastiaanSL (approved):
Users will be expecting this to work with node keys, so to me it's not necessary to explicitly reference it. With that being said, the content is correct.

EdithLansens:
> Users will be expecting this to work with node keys, so to me it's not necessary to explicitly reference it. With that being said, the content is correct.

Thanks for the feedback! We will not include it in the user guide then.

PR #2097, opened 2023-10-11
[STaaS] Error for API Deployment

GuillaumeVadeBe:
Added the error that will be shown when API Deployment is still enabled in the SoftLaunchOptions while it shouldn't be.


LouisSkyline (approved):

LouisSkyline (changes requested):
Also mention the error shown in cube, as this is the one users will see most likely first

joaobarraca (approved):

MariekeGO (approved):

PR #2151, opened 2023-10-19
Skip verify server certificate validation documentation

mmphs:


user-guide/Reference/DataMiner_Extension_Modules.md
MariekeGO, line 1: @mmphs The new section on gRPC configuration is a bit too extensive to still be a good fit in this concise overview of DxMs. I think it would be better to rewrite this a bit and add it as a "how-to" procedure "Configuring gRPC-related settings" under dataminer.services > Advanced configuration. What do you think?

jaklke: dataminer.services cover items related to connected agents to the (cloud) services. The Although the CommunicationGateway DXM can be easily deployed when being cloud connected, this setting is not related how DataMiner connects to the cloud.
I think configuration items for this DXM probably fit better under the DataMiner Modules (like the Aggregator DXM). If we do this, then the text for this DXM from Reference > DataMiner Extension Module should also direct to this new page under DataMiner Modules..

MariekeGO: I agree that it would be better not to add this under dataminer.services. However, I'm also a little hesitant to add this as another module under DataMiner modules. CommunicationGateway seems to me to be more a DxM that handles things in the background, rather than a module with all sorts of functionality a user can play with — which is what that section is primarily meant for. I see several other possibilities:

- Reference > Skyline DataMiner folder > More information on certain files and folders > CommunicationGateway.gRPC.config.json

  This would place the procedure with other more unusual small tweaks that are done directly in files. It's a different perspective: we're not looking at the config from the perspective of the DxM, but from the perspective of the file where the user needs to do the changes. We should then link to it from the DxM section to make sure that users can find it from that angle as well.

- Reference > DataMiner Extension Modules > Communication Gateway

  The advantage of this location is that it could be an example for other more extensive DxM pages. The disadvantage is that users may not expect this configuration info here, and it might look a little lost here.  

Other locations may also be suitable, depending on why a user might want to implement this particular configuration. If it's a rather unusal configuration that is only used in very specific cases, personally I'd be in favor of adding it under the Skyline DataMiner folder, since that is where other such procedures usually end up.

mmphs: Thanks for thinking along @MariekeGO, I like the perspective of starting from the config file itself and then linking to it from the DxM page. However this might be confusing seeing that this file is not stored in the Skyline DataMiner folder but under Program Files\Skyline Communications\DataMiner CommunicationGateway.

Would it be an option to start a new section on files outside of the Skyline DataMiner folder?

MariekeGO: A section on any files outside the Skyline DataMiner folder might create expectations we can't live up to, because if we have to add every possible config file outside of that folder, that will take ages. Maybe we can instead approach this from the configuration angle. In which contexts would someone want to configure this? Is it when they're setting up their DMS, when they're configuring security, ... ?
For instance, if we consider this to be a security config procedure, we can add it under Advanced DataMiner functionality for administrators > Security > Advanced security configuration.

mmphs: This configuration needs to be used when they're configuring security indeed. More specifically to weaken it actually. 
When deploying the CommunicationGateway DxM to start pulling data in from gRPC services. (biggest use case currently is OpenConfig but more gRPC services are following in a fast pace), we're very strict on validating the certificate (hostname should match, should not be expired etc.). With this configuration, they can configure hosts for which that should not be the case, so even with an invalid certificate, we will still allow the connection.

So writing it down in the Security > Advanced security configuration seems like a good approach to me, if it's allowed to actually write down information in there to 'weaken' the security?

On top of that, this configuration file now only contains this very specific security related item but might grow in the future with other (non-security related) settings. But I guess when that will be the case we can re-evaluate?

MariekeGO: If we add this under security, we'll just need to add a caution box to make sure it's definitely clear that this is something you should only do in exceptional circumstances because it can weaken security. And we can indeed re-evaluate later, when we know more about other settings.

MariekeGO (approved):

PR #2155, opened 2023-10-20
Update DSI_OpenConfig_Middleware.md

LaurensMT:
Added new options that have been made available with NuGet version 1.0.3

### DIFF
--- a/develop/DSI/DSIOpenConfig/DSI_OpenConfig_Middleware.md
+++ b/develop/DSI/DSIOpenConfig/DSI_OpenConfig_Middleware.md
@@ -419,11 +419,11 @@ public static object ConvertEpochTimeUtcTicksToOleAutomationTime(DataMinerConnec
 }
 ```
 
-#### Complex value
+##### Complex value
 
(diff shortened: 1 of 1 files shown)


develop/DSI/DSIOpenConfig/DSI_OpenConfig_Middleware.md
```diff
@@ -405,6 +419,14 @@ public static object ConvertEpochTimeUtcTicksToOleAutomationTime(DataMinerConnec
 }
 ```
 
+#### Complex value
```
MariekeGO, line 422: @LaurensMT, these two titles (line 422 & 426) don't really seem to fit here — or at least not at this header level. All the other level-4 headers here are how-to instructions, while these two seem to be reference sections about specific values. Should these actually be level-5 headers meant as subsections of "Converting the value before doing the set"? Or otherwise, could you adjust them so they are also how-to instructions, in line with the other level-4 headers here?

LaurensMT: @MariekeGO, I've changed it into level-5 headers

MariekeGO (approved):

PR #2178, opened 2023-10-24
Create DataMinerProbes

robin-devos-skyline:
A page to clarify what a DM Probe is, as it's been asked a couple of times in Dojo before.

PR #2179, opened 2023-10-24
Create upgrading_a_dataminer_probe.md

robin-devos-skyline:
upgrading a probe is a bit different to a regular DMA


MariekeGO:
Included in #2178

PR #2180, opened 2023-10-24
Update toc.yml

robin-devos-skyline:
Added 2 things
- Dataminer Probes
- Upgrading a Dataminer Probe Also, not sure if there's a certain guideline to the menu-items (do they need to be alphabetically organized?)


MariekeGO:
Included in #2178

PR #2202, opened 2023-10-30
Managed services now deprecated

MariekeGO:


PieterVanCompernolle (approved):
I did not do a full Review, but looked at a couple of pages and it looks great to me!

PR #2207, opened 2023-10-31
RN 37651

MariekeGO:


release-notes/CommunicationGateway/CommunicationGateway_1.2.2.md
```diff
@@ -0,0 +1,27 @@
+---
+uid: CommunicationGateway_1.2.2
+---
+
+# Communication Gateway 1.2.2
+
+## New features
+
+#### Option to skip verify server certificate validation [ID_37651]
```
mmphs, line 9: Is it correct that we are skipping a header level here?
I've seen that in other articles as well so it probably is but I was wondering what's the reasoning behind it?

MariekeGO: Yes, we always use level 4 for release note titles, so that they look the same everywhere. That means that we sometimes need to skip a header level, like here for instance.

mmphs (approved):

PR #2221, opened 2023-11-03
RNs 37148 + 37171 + 37670

EdithLansens:
Hi @tobedeprez and @SkylineDennisDr, could you please review this pull request to confirm the accurate inclusion of release notes 37148, 37171, and 37670 in the user guide? Thanks in advance!


user-guide/Basic_Functionality/Protocols_and_templates/Alarm_templates/Configuring_alarm_templates/Configuring_anomaly_detection_alarms.md
```diff
@@ -4,7 +4,74 @@ uid: Configuring_anomaly_detection_alarms
...
+
+   - **Disabled**: No anomaly alarms are generated for this parameter.
+
+   - **Customized**: You have customized the anomaly alarm settings for this parameter.
+
+   - **Smart**: Anomaly alarms are generated for this parameter based on what DataMiner Analytics determines to be an anomaly.
+
+1. To save your changes, click *OK* or *Apply* in the lower right corner of the alarm template editor.
+
+## [Prior to DataMiner 10.2.0 [CU21]/10.3.0 [CU9]/10.3.12](#tab/tabid-2)
```
tobedeprez, line 74: I'm not sure how this is usually documented, but there are now three possibilities depending on your Cube and DataMiner version:

- Old Cube (i.e. a version older than 10.2.0 [CU21]/10.3.0 [CU9]/10.3.12) and old DataMiner (i.e. earlier than 10.3.12/10.4.0): configuration is still done as described in this last tab
- New Cube and old DataMiner: configuration also works via a pop-up here, but it only has the options that were previously available via the procedure below
- New Cube and new DataMiner: configuration works via de procedure above

Since one can also work with a different version of Cube than is included in the upgrade package (maybe it is even the default?), whether you have the second option or the third does not depend on the version of DataMiner you have installed, but only on the version of Cube.

user-guide/Basic_Functionality/Protocols_and_templates/Alarm_templates/Configuring_alarm_templates/Configuring_anomaly_detection_alarms.md
```diff
@@ -4,7 +4,74 @@ uid: Configuring_anomaly_detection_alarms
...
+
+     1. Configure alarms for specific types of anomaly detection:
+
+        - *Level shift > Level increases*: Enables or disables alarms for level increase anomalies.
+
+        - *Level shift > Level decreases*: Enables or disables alarms for level decrease anomalies.
+
+        - *Outlier > Upward spikes*: Enables or disables alarms for upward spikes.
+
+        - *Outlier > Downward spikes*: Enables or disables alarms for downward spikes.
```
tobedeprez, line 39: I think you should also include some explanation about the custom thresholds that can be set for level shifts and outliers:

- You can set absolute thresholds by selecting *Absolute* in the drop down menu next to each change point type: this will alarm level shifts based on the size of the jump (resp. outliers based on how far the outlier is from the surrounding points in absolute numbers). e.g. if you select *Absolute* for level increases and fill in 150 in the box for severity major, then a parameter jumping from, say, 100 to 255 will get a major alarm (since 255 - 100 = 155 > 150). If it jumped from 100 to 245 on the other hand, no alarm (and also no suggestion event) is generated. 
- You can set relative thresholds by selecting *Relative* in the drop down menu next to each change point type: this will alarm level shifts based on the size of the jump *relative to the value where the jump started*. This is then expressed in %. E.g. in the same example as above, if we select relative (instead of absolute) for level increases and set a value of 100% in the box for severity major, then the jump from 100 to 255 (i.e. 155% increase) or from 100 to 245 (i.e. 145% increase) will both get major alarms, but a jump from 200 to 250 will not get an alarm, since that is only an increase with 25%.

tobedeprez (changes requested):

user-guide/Basic_Functionality/Alarms/Working_with_alarms/Advanced_analytics_features/Behavioral_anomaly_detection.md
```diff
@@ -15,7 +15,7 @@ You can also configure alarm templates to have alarms generated instead of sugge
 
 Please note the following regarding suggestion events:
 
-- Currently, no suggestion events are generated for outliers and unlabeled change points.
+- Currently, no suggestion events are generated for unlabeled change points. Prior to DataMiner 10.2.0 [CU21]/10.3.0 [CU9]/10.3.12<!--RN 37171-->, no suggestion events are generated for outliers and unlabeled change points.
```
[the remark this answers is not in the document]
SkylineDennisDr: Hi, I believe the old line saying that no suggestion events are ever generated for outliers and unlabeled change points is still correct and no changes are required...

EdithLansens:
Hi @tobedeprez and @SkylineDennisDr, I have made some changes to the pages based on your input: [commit f6cc093](https://github.com/SkylineCommunications/dataminer-docs/pull/2221/commits/f6cc0937d12c6d88a8a87cdcc54c100c4f736e3f). Could you review these changes?

user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Working_with_behavioral_anomaly_detection.md
```diff
@@ -31,7 +31,7 @@ From DataMiner 10.0.0/10.0.2 onwards, the DataMiner Analytics software can detec
...
+If a change point other than an unlabeled change is unexpected, it will be considered anomalous. Prior to DataMiner 10.3.12/10.4.0<!--RN 37171-->, if a change point other than an outlier or unlabeled change is unexpected, it will be considered anomalous. Level shifts that have a different direction than previous recent jumps or that jump to a previously unseen level will typically be labeled “anomalous”. Similarly, trend or variance changes will be labeled “anomalous” when no earlier trend or variance changes in the same direction were detected during the last weeks. A flatline will be considered anomalous when no recent flatline change point of approximately the same length or longer is detected.
```
tobedeprez, line 34: Don't know how detailed we want to go here, but an outlier will only be considered an anomaly if alarming for outliers is turned on (at least for now, this might change in the future cfr. DCP223402). I think in the context of this file, it is more correct to just leave the old explanation.

tobedeprez (approved):

PR #2246, opened 2023-11-08
Overhaul SAML configuration

MariekeGO:


BertVandenberghe (approved):
Looks great! Thanks!

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
```diff
@@ -0,0 +1,474 @@
...
+   1. Under *Source attribute*, we recommend selecting *sAMAccountName*.
+
+   1. Optionally, expand the advanced options to add a filter or customize the name of the group claim.
+
+   1. Click *Save*.
+
+   ![Group claim configuration](~/user-guide/images/SAML_Group_claim_config.png)
+
+   > [!NOTE]
+   > The account name of the group will only be sent via SAML when the groups are synchronized. Otherwise, the ID of the group will be sent instead.
```
MichielMN, line 408: As a clarification here: where the groups need to be synchronized from is an on prem AD, I'm also looking into a new option added recently by microsoft to send the names of "cloud-only groups" (their words) I am trying to test this today to confirm this works as I expect. If so, we could also add that as an alternative option

MariekeGO: I'd suggest that we add this in a new pull request later; then we can already merge these improvements, as they'll make quite a difference for the users already.

MichielMN: I would just add the "synchronized from an on prem AD" to the docs now, since that can be very confusing to read

PR #2250, opened 2023-11-09
Update Working_with_behavioral_anomaly_detection.md

SkylineDennisDr:
In RN37827, we have changed the meaning of the color of the bars underneath the trend graph that indicate the location of change points.


user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Working_with_behavioral_anomaly_detection.md
MariekeGO, line 1: As the documentation has to be valid for any supported DataMiner version, could you adjust this so it's clear in which version you get which behavior? Right now, only the latest behavior is described, which will be confusing for users with older versions.

SkylineDennisDr: Added version information to the text!

MariekeGO (approved):

PR #2294, opened 2023-11-20
Backup procedure for Cassandra cluster using Medusa

kapembe:
How to configure Medusa, how to take a backup and how to restore a backup.

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Cassandra_cluster_backup_restore/Backup_Restore_Cassandra_Cluster.md
+++ b/user-guide/Advanced_Functionality/Databases/Cassandra_cluster_backup_restore/Backup_Restore_Cassandra_Cluster.md
@@ -15,10 +15,12 @@ Medusa serves as an Apache Cassandra backup system, offering a command-line inte
 
 ## Configuring the firewall and generating SSH keys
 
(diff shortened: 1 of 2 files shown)


meitavelo:
@ThibaultHeylen

MariekeGO (changes requested):
Could you go over your procedure again and add more detail? Right now, many of the steps are too vague for users to easily be able to follow. For instance, for line 14/18: where should the user specify that command? Is this the same regardless of which OS they are using or does your procedure apply for a specific OS? For line 16: where do they find the public key and where do they paste it? Make sure there is no missing info in the procedure and that it's always clear what the users need to do and where.
Note also that I've now moved the "Generate SSH keys" section to the top because I get the impression they can't do the backup unless this has been done, but perhaps it might be better to actually combine the backup procedure with the procedure to generate SSH keys, as one continuous set of steps?

spetebroodt:
Hi @meitavelo Can you double-check the corrected text? Also, two different Linux flavors are mentioned on this page: Ubuntu LTS and CentOS (link to procedure specific for CentOS?). Is this correct? I thought Ubuntu LTS is recommended.

spetebroodt (approved):

PR #2303, opened 2023-11-21
GQI Extensions - How to add or modify links to DataMiner objects (using metadata)

skyline-rge:
Includes:

- How to link DataMiner objects to a GQI row using metadata in ad hoc data sources (+ example)
- How to modify links to DataMiner objects using metadata in custom operators (+example)
- Added API reference for GenIfRowMetadata, RowMetadataBase, ObjectRefMetadata
- Extended API reference for GQIRow/GQIEditableRow

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
@@ -6,7 +6,7 @@ uid: Ad_hoc_Metadata
 
 GQI allows query rows to be linked to certain DataMiner objects. This is achieved via metadata that informs the Dashboards and Low-Code Apps of what a row represents. The metadata can then be used to feed these DataMiner objects from one component to another.
 
-From DataMiner 10.4.1/10.4.0 onwards, you can also add this metadata to the rows of your ad hoc data source, just like the rows from built-in data sources.
+From DataMiner 10.4.0/10.4.1 onwards, you can also add this metadata to the rows of your ad hoc data source, just like the rows from built-in data sources.
 
 To link a GQI row to a DataMiner object, add a reference to the object in the [Metadata property](xref:GQI_GQIRow#properties) of the row.
 
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
@@ -1,0 +1,22 @@
+---
+uid: Ad_hoc_Metadata
+---
+
+# Linking rows to DataMiner objects
+
+Just like the built-in data sources, you can link the rows of your ad hoc data source to specific DataMiner objects. This will allow you to feed these objects in the Dashboards and Low-Code Apps to other components.
+
+To link a GQI row to a DataMiner object, add a reference to the object in the [Metadata property](xref:GQI_GQIRow#properties) of the row.
+
+## Example
+
+The following ad hoc data source exposes a table of elements contained within a specific view.
+
+:::code language="csharp" source="./SLC-GQIDS-ViewElements.cs":::
+
+## See also
+
+- [Supported references to DataMiner objects](xref:GQI_ObjectRefMetadata#dmaobjectref)
+- [ObjectRefMetadata](xref:GQI_ObjectRefMetadata)
+- [GenIfRowMetadata](xref:GQI_GenIfRowMetadata)
+- [GQIRow](xref:GQI_GQIRow)
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
@@ -6,7 +6,7 @@ uid: Ad_hoc_Metadata
 
 GQI allows query rows to be linked to certain DataMiner objects. This is achieved via metadata that informs the Dashboards and Low-Code Apps of what a row represents. The metadata can then be used to feed these DataMiner objects from one component to another.
 
-Just like the rows from built-in data sources, you can now also add this metadata to the rows of your ad hoc data source.
+From DataMiner 10.4.0 onwards, you can also add this metadata to the rows of your ad hoc data source, just like the rows from built-in data sources.
 
 To link a GQI row to a DataMiner object, add a reference to the object in the [Metadata property](xref:GQI_GQIRow#properties) of the row.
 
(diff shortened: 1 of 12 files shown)


skyline-rge:
Related to [RN37879: GQI - Ad Hoc data source/Custom operator - Support for row metadata](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=37879).

@WaHaeck can you have a look too?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
```diff
@@ -0,0 +1,22 @@
+---
+uid: Ad_hoc_Metadata
+---
+
+# Linking rows to DataMiner objects
+
+Just like the built-in data sources, you can link the rows of your ad hoc data source to specific DataMiner objects. This will allow you to feed these objects in the Dashboards and Low-Code Apps to other components.
```
WaHaeck, line 7: Was a bit confused by 'Just like the built-in data sources', it sounded like its expected from me (as user) to know how that's done. Might be clearer to first explain that certain built in data sources contain metadata to map the row on a DataMiner object + why you should do that (feeds).

WaHaeck (approved):
Very nice!

WaHaeck (approved):

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Custom_Operator/CO_Metadata.md
```diff
@@ -0,0 +1,28 @@
+---
+uid: CO_Metadata
+---
+
+# Modifying links to DataMiner objects
+
+In a custom operator, you can link row or modify any existing links to DataMiner objects.
```
SebastiaanSL, line 7: This sentence does not read nicely.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Custom_Operator/CO_Metadata.md
```diff
@@ -0,0 +1,28 @@
...
+These metadata links allow you to feed the objects in the Dashboards and Low-Code Apps to other components.
```
SebastiaanSL, line 8: In Dashboards and Low-Code Apps (drop the "the")

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
```diff
@@ -4,7 +4,7 @@ uid: Ad_hoc_Metadata
 
 # Linking rows to DataMiner objects
 
-Just like the built-in data sources, you can link the rows of your ad hoc data source to specific DataMiner objects. This will allow you to feed these objects in the Dashboards and Low-Code Apps to other components.
+Just like the rows from built-in data sources can contain metadata that allow you to feed certain DataMiner objects in your Dashboards and Low-Code Apps, so too can you link the rows of your ad hoc data source to these DataMiner objects.
```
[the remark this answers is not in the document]
skyline-rge: Done.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Custom_Operator/CO_Metadata.md
skyline-rge: Done.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Custom_Operator/CO_Metadata.md
skyline-rge: Done.

SebastiaanSL (approved):

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
```diff
@@ -0,0 +1,24 @@
+---
+uid: Ad_hoc_Metadata
+---
+
+# Linking rows to DataMiner objects
+
+GQI allows query rows to be linked to certain DataMiner objects. This is achieved via metadata that informs the Dashboards and Low-Code Apps of what a row represents. The metadata can then be used to feed these DataMiner objects from one component to another.
+
+Just like the rows from built-in data sources, you can now also add this metadata to the rows of your ad hoc data source.
```
MariekeGO, line 9:
```suggestion
From DataMiner 10.4.1/10.5.0 onwards, you can also add this metadata to the rows of your ad hoc data source
```

MariekeGO (changes requested):
The only thing I'm missing in this PR right now is version info. If the classes/procedures you added are only available or supported from 10.4.1/10.5.0 onwards, could you make sure this is clear? Otherwise users with older DataMiner versions may get confused.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Metadata.md
skyline-rge: Done.

skyline-rge:
Note: Probably this was already suggested many times before, but I do want to stress again that we should think about having docs per version instead of interspersing version info everywhere.
As a user, I just want to select the version I need once and then read the docs without the extra clutter😢

MariekeGO:
> Note: Probably this was already suggested many times before, but I do want to stress again that we should think about having docs per version instead of interspersing version info everywhere. As a user, I just want to select the version I need once and then read the docs without the extra clutter😢

We have chosen to have all documentation in one place to make it easy to maintain, to allow easy cross-referencing, and to have one search with results throughout the documentation. Splitting the documentation into one version per DataMiner version is not feasible with our current approach. There are not only the limitations of the repo that we need to keep in mind (max 15000 pages), but with the different release tracks and the differences between server, Cube, and Web apps, this would be a very complex and time-consuming change. In addition, because changes to the documentation are not necessarily limited to the latest DataMiner versions only, we'd risk having to spend a lot more time on keeping the documentation up to date if we were to do this (i.e. checking all relevant versions and doing multiple updates where now one is enough).
This being said, mentioning the version info is very important with our current implementation, and we need to make sure we mention **both main and feature release** wherever relevant. Could you check your PR and make sure this is indicated?

skyline-rge:
I added the FR version now too. I was confused thinking MR 10.4.0 would release before FR 10.4.1.
I think a version indication for the new type reference files won't be relevant.

> We have chosen to have all documentation in one place to make it easy to maintain, to allow easy cross-referencing, and to have one search with results throughout the documentation. Splitting the documentation into one version per DataMiner version is not feasible with our current approach. There are not only the limitations of the repo that we need to keep in mind (max 15000 pages), but with the different release tracks and the differences between server, Cube, and Web apps, this would be a very complex and time-consuming change. In addition, because changes to the documentation are not necessarily limited to the latest DataMiner versions only, we'd risk having to spend a lot more time on keeping the documentation up to date if we were to do this (i.e. checking all relevant versions and doing multiple updates where now one is enough).

I did not want to imply that there were no good reasons to document the version info as we currently do or that the team had not carefully considered it.
Just wanted to give my honest feedback since, regardless of the technical and practical challenges, I believe we'll need to move towards a different system eventually if we want to keep this clear and user-friendly. Which is of course the ultimate goal.

MariekeGO (approved):

PR #2305, opened 2023-11-21
DOM - It's now possible to define CRUD actions for DomInstances on DomDefinition level

VanRobaysWouter:
RN37963 DCP156079


MariekeGO (commented):
I see the release note also mentions "The ScriptSettings object in the DomDefinition has to be null if you want the ModuleSettings ScriptSettings to be used. Just making the values empty is not sufficient!" Is this something that should also be added here (e.g. in a note)?

VanRobaysWouter:
> I see the release note also mentions "The ScriptSettings object in the DomDefinition has to be null if you want the ModuleSettings ScriptSettings to be used. Just making the values empty is not sufficient!" Is this something that should also be added here (e.g. in a note)?



> I see the release note also mentions "The ScriptSettings object in the DomDefinition has to be null if you want the ModuleSettings ScriptSettings to be used. Just making the values empty is not sufficient!" Is this something that should also be added here (e.g. in a note)?

Added note for this

MariekeGO (approved):

PR #2311, opened 2023-11-22
Update Protocol.HTTP.Session-timeout.md

EdibSU:
Is this a mistake? DataMiner will use this value instead of one specified in Session tag, but this value is the value in Session tag. Maybe it will overrule some other timeout setting and not the one in Connection tag?


EdibSU:
Seems like this is the same description as one provided for [Connection](https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.HTTP.Session.Connection-timeout.html). Maybe entire sentence requires an update?

PedroDebevere (approved):
Reworked a bit to make it more clear and added cross ref links.

MariekeGO (approved):

PR #2313, opened 2023-11-22
Update Security_Elasticsearch.md

KevinVerm:
@SeppeDejonckheere

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Database_security/Security_Elasticsearch.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Database_security/Security_Elasticsearch.md
@@ -104,6 +104,31 @@ To configure TLS encryption for client-server communication:
    bin\elasticsearch-keystore add xpack.security.http.ssl.keystore.secure_password
    bin\elasticsearch-keystore add xpack.security.http.ssl.truststore.secure_password
    ```
+   For Ubuntu the procedure is the following:
+
+   1. Navigate to the bin folder using the following command:
+
+      ```bash
+      cd /usr/share/elasticsearch/bin
+      ```
+
+   1.Set the password for the keystore using the following command:
+
+      ```bash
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Database_security/Security_Elasticsearch.md
```diff
@@ -105,6 +105,31 @@ To configure TLS encryption for client-server communication:
...
+   1. Set the password for the truststore using the following command:
+
+      ```bash
+      sudo ./elasticsearch-keystore add xpack.security.http.ssl.truststore.secure_password
+      ```
+
+   > [!NOTE]
+   >
+   > - Depending on the flavor of operating system (Windows, Ubuntu, or another OS), the procedure can be different. Note that we generally recommend Ubuntu.
+   > - If the password needs to be updated, because you ran the generate-tls-certificates again and you acquired a new password for the keystore and truststore, the commands above to set the keystore and truststore passwords in `/usr/share/elasticsearch/bin` need be executed again. These will overwrite the previous password.
```
MariekeGO, line 131: This sentence is a bit ambiguous. Is it correct if I rephrase this as follows?

```suggestion
   > - If the password needs to be updated, after you have run generate-tls-certificates again and acquired a new password for the keystore and truststore, you need to execute the commands above to set the keystore and truststore passwords in `/usr/share/elasticsearch/bin` again. These will overwrite the previous password.
```

KevinVerm: Yes that's a bit better rephrased I think.

KevinVerm: Good to have an extra set of eyes

MariekeGO (approved):

PR #2319, opened 2023-11-23
Update STaaS migration limitations

ruben12321:
Update the docs about the limitation for migrations on STaaS


user-guide/Advanced_Functionality/Databases/STaaS/STaaS.md
MariekeGO, line 1: What are the release notes that introduced this change? Usually we add these in comment to make it easier to trace which RN introduced which change.

ruben12321: For Cassandra cluster it's RN37283 and for Cassandra single it's RN37408

MariekeGO (approved):

PR #2360, opened 2023-12-04
added Remote PHY Monitoring package information

MariaCUSkyline:


user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Installation.md
```diff
@@ -0,0 +1,63 @@
...
+uid: Remote_PHY_Monitoring_Installation
+---
+# Installing Remote PHY Monitoring
+
+1. Make sure the following **prerequisites** are met:
+
+   - Your DataMiner System uses DataMiner 10.3.0 or higher.
+
+     > [!IMPORTANT]
+     > If you are using a version older than 1.0.4.0, upgrade the DataMiner web apps version to DataMiner 10.3.11 or higher.
```
MariekeGO, line 11: Is "1.0.4" the version of the solution here? This seems odd in a procedure for first-time installation. Would it make sense to instead add as a requirement that the web apps must be upgraded to 10.3.11 or higher in all cases?

user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Components.md
```diff
@@ -0,0 +1,28 @@
...
+## Low-code apps
+
+The Remote PHY Monitoring Solution includes the following low-code apps:
+
+- **R-PHY Monitoring:** This low-code app allows you to easily monitor all Remote PHY nodes in the system.
+
+- **R-PHY Analog & RF:** This low-code app monitors the analog and RF KPIs for all the Remote PHY nodes in the system. This allows on-site technicians to have remote visibility on how the Vecima R-PHY nodes are performing during on-site installation or troubleshooting.
+
+> [!NOTE]
+> All elements related to the low-code apps must use the **Production** version of the connector.
```
MariekeGO, line 28: Could you detail which elements are related to the low-code apps? Users might not know this, so this could cause questions.

MariekeGO (changes requested):

user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Components.md
MariaCUSkyline: It is written on the components page that right know the solution only support Vecima RPM elements. Do you think I should add it here too?

user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Installation.md
MariaCUSkyline: Yes, we can write that, but in February version 1.0.4.0 will be released and in that version we will no longer have to update the web applications.

user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Components.md
MariekeGO: In that case, we could adjust this note like this to avoid any possible confusion:

```suggestion
> All elements related to the low-code apps (i.e. the Vecima RPM elements) must use the **Production** version of the connector.
```

user-guide/Standard_Apps/Vecima Remote PHY Monitor/Remote_PHY_Monitoring_Installation.md
MariekeGO: I'd suggest to add this as a requirement as follows then:

```
   - If version 1.0.4.0 or older of the solution is used, the DataMiner web apps must be upgraded to DataMiner 10.3.11 or higher.
```

MariekeGO (approved):

PR #2364, opened 2023-12-04
Extend GetParameterByPrimaryKey documentation

AndreRodriguesSkyline:
Update the documentation of the GetParameterByPrimaryKey method to correctly represent the behavior of the method when information templates are present.

### DIFF
--- a/src/DataMiner/SLManagedAutomation/Element.cs
+++ b/src/DataMiner/SLManagedAutomation/Element.cs
@@ -760,7 +760,7 @@ namespace Skyline.DataMiner.Automation
 		/// </code>
 		/// </example>
 		/// <remarks>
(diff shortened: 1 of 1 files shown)


src/DataMiner/SLManagedAutomation/Element.cs
```diff
@@ -759,6 +759,9 @@ public class Element : IActionableElement
 		/// var value = element.GetParameter(1002, "1");
 		/// </code>
 		/// </example>
+		/// <remarks>
+		/// <para>When the primary key is in a column hidden by an information template, this method will return an empty value. We recommend using the <see cref="GetParameter(int, string)"/> method.</para>
```
PedroDebevere, line 763:
```suggestion
		/// <para>If a column is hidden by an information template, the value for a cell of that column can no longer be retrieved by this method (null is returned). However, the value can stilll be retrieved using the <see cref="GetParameter(int, string)"/> method. Using the GetParameter method is therefore recommended.</para>
```

MariekeGO (approved):

PR #2376, opened 2023-12-06
Update STaaS.md - DBConfiguration.xml is deprecated

joaobarraca:


user-guide/Advanced_Functionality/Databases/STaaS/STaaS.md
```diff
@@ -78,7 +78,7 @@ To ensure data resilience for potential recovery scenarios, protecting against u
 To **migrate existing data** to STaaS, the following limitations apply:
 
 - Migrating logger tables is supported from DataMiner 10.3.11 onwards<!-- RN 37283 --> for systems using a [dedicated clustered storage setup](xref:Dedicated_clustered_storage), and from DataMiner 10.3.12 onwards<!-- RN 37408 --> for systems using a Cassandra database per DMA.
-- Migration of an OpenSearch or Elasticsearch database that is defined in [DBConfiguration.xml](xref:DBConfiguration_xml) instead of [DB.xml](xref:DB_xml) (e.g. because a setup with multiple OpenSearch/Elasticsearch clusters is used) is not yet supported.
+- Migration of a setup with multiple OpenSearch/Elasticsearch clusters is not yet supported.
```
MariekeGO, line 81: What if people use an older DataMiner version that still uses DBConfiguration.xml? Is it only the multiple OpenSearch/Elasticsearch clusters config that was the problem there? Or also the configuration of the size limit for file offloads?

joaobarraca: if people use older versions of DataMiner, it still applies. Yes it was only related to multiple clusters being defined there

MariekeGO (approved):

PR #2417, opened 2023-12-18
Overview FAQ corrections

MariekeGO:
@SammyDW, could you take a look at this PR and confirm if it's OK to merge? It contains a number of corrections for things that went under the radar in an earlier PR. I've also added some cross-references to STaaS information that has recently been added in the user guide.


SammyDW (approved):
Thanks for the checks!

PR #2428, opened 2023-12-19
Troubleshooting Web Apps

wimme:
Some troubleshooting information which hopefully makes it easier for everyone to know what to do when running into an issue with one of the DataMiner web apps.


spetebroodt:
@wimme I reviewed the changes you made.

I suggest the "troubleshooting - web" page be moved to the "troubleshooting procedures" section. As this new page doesn't contain any flowcharts :-)

If we move the file, we could rename it to e.g. "Investigating web app issues"

spetebroodt (approved):

PR #2429, opened 2023-12-20
Indexing documentation changes

MariekeGO:
- Adjusted existing documentation to make it clear that you can use OpenSearch with storage per DMA. 
- Also added the possibility to use an Elasticsearch database on a separate server with storage per DMA (similar to OpenSearch).
- Modified the structure of the indexing section to make more sense in light of the changes above
- Removed tips referring to deprecated functionality.


user-guide/Advanced_Functionality/Databases/Indexing_database/Indexing_Database.md
```diff
@@ -30,4 +18,35 @@ While these are not recommended, we also still support setups with storage per D
 
 - [Specific GQI data sources](xref:Query_data_sources)
 
-![indexing database per DMS](~/user-guide/images/Indexing_database_per_DMS.svg)
+> [!IMPORTANT]
+> Elasticsearch is only supported up to version 6.8. As such, we recommend using OpenSearch instead.
+
```
SLC-Laurens-Vercruysse, line 23: This is double. DataMiner supports only Elasticsearch 6.8.x, but Elasticsearch itself no longer provides support for that version. 

Perhaps it would be good to clarify this in this main page. Note that I'm okay with using the abridged phrase "Elasticsearch is only supported up to version 6.8" on all the other pages. It would get a bit big and unnecessarily cumbersome otherwise.

SLC-Laurens-Vercruysse (changes requested):
Looks good on the whole. One remark concerning on the main page of indexing, other than this I wonder whether we should add a section that explains that the end-user would be responsible to install Opensearch on Windows on his own, that there is nothing in the way of a DataMiner feature/installer that does this for them at the moment. 

For that, however, extra feedback may be requested from Platform Delivery.

SLC-Laurens-Vercruysse (approved):

PR #2452, opened 2023-12-22
Release of new tools open up CICD for more (all) platforms

janstaelensskyline:


MariekeGO:
@janstaelensskyline, could you solve the remaining issue that causes the build to fail?
The TOC contains a reference to the UID "Deploying_Automation_scripts_from_a_repository", which doesn't exist. Maybe this is a file you still need to add?

develop/CICD/Azure DevOps/CICD_Azure_Devops_Examples.md
```diff
@@ -0,0 +1,54 @@
+---
+uid: CICD_Azure_DevOps_Examples
+---
+
+# Azure DevOps CI/CD Examples
+
+## Basic Deployment Example
+
+A basic pipeline for upload to catalog and/or deployment to cloud-connected agents.
+It's recommended to combine this with quality control beforehand such as executing static code analysis and running tests.
```
MariekeGO, line 10:
```suggestion
This is a basic pipeline for uploading to the catalog and/or deployment to DMAs connected to dataminer.services.

We recommend combining this with quality control beforehand, such as executing static code analysis and running tests.
```

develop/CICD/Azure DevOps/CICD_Azure_Devops_Examples.md
```diff
@@ -0,0 +1,54 @@
...
+---
+
+# Azure DevOps CI/CD Examples
+
+## Basic Deployment Example
+
+A basic pipeline for upload to catalog and/or deployment to cloud-connected agents.
+It's recommended to combine this with quality control beforehand such as executing static code analysis and running tests.
+
+## Creating a dataminer.services key
```
MariekeGO, line 12: Should "Creating a dataminer.services key" be considered part of the "basic deployment example"? Because right now, they're the same header level, so it looks as if this is a completely new section unrelated to what came before.

janstaelensskyline: Yes, I changed this to be part of the basic deployment example

MariekeGO (changes requested):
For all the examples, I feel like a bit more information is needed, as right now users may struggle to know what exactly they should do. For example:
- Under "Azure DevOps CI/CD examples" > "Azure DevOps Pipeline", you only mention what variables are needed and then give a yml example. But where and how can they use this example? 
- Under "Command line CI/CD examples", do the two titles after "Creating a dataminer.services key" represent two different possibilities, and can the user choose depending on the system they are using? Without any explanation, some users may get confused, because the document structure itself suggests these are consecutive steps. 

Could you go over the example pages again and make sure such things are more clear? I want to avoid users getting frustrated because they only find partial information. (Note that if you can make things clear by adding a cross-reference to a page with more info, that will also help considerably and be less work for you 😊).

develop/toc.yml
```diff
@@ -5105,14 +5105,21 @@
...
     topicUid: CICD_on_GitHub
     items:
+    - name: GitHub example workflows
+      topicUid: CICD_GitHub_Examples
     - name: GitHub starter workflows
       topicUid: github_starter_workflows
     - name: GitHub validation requirements
       topicUid: github_validation_requirements
-    - name: Deploying from GitHub
+    - name: MarketPlace action
```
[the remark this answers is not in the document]
janstaelensskyline: I'm not sure I completely agree. The missing information is likely a more detailed "how-to" very specific to the CICD technology you're wanting to use. All of which is a simple 'google' away or will likely already be known for any cicd engineer wanting to make such a set-up. Concourse, Azure Devops, GitHub Actions, GitLab, ...  all of those have their own help & documentation that tells you what to push until you get to your "empty script". At that point you can copy paste one of the examples we prove and add to it. (or add our example to an existing script)

I'd like to avoid adding too much extra industry-standard information and focus more on the DataMiner specific information you need to immediately get things up and running in your technology of choice.

Would it be good enough you think, if I just add some links to the help pages of those technologies?

---------
There is a command line in Windows and a command line in Linux (ubuntu).  Windows usually goes with Powershell these days and Ubuntu has the Terminal with bash scripts. So that page shows both options, you can pick which one based on your Operating System. It doesn't make much sense to see that as consecutive steps as they are for different operating systems.

I can add an extra line or two, mentioning that there is an example in powershell for windows users and an example with ubuntu terminal for ubuntu users wishing to use the command line.

---------

I was afraid of changing the file names and UID's of the existing "Deploying from Github" because I think we have linked to that several times in answers, e-mails and dojo posts. However that name doesn't make sense any more, it was too generic and it makes it confusing for the users.  The page is talking about the Github marketplace action that allows deployment. But for most users we would refer them now to use the platform-independent .NET Tools we just released.

If you think it's OK, I can look into adjusting the filename & UID of the old page then and not just the treecontrol displayname and internal title.

MariekeGO (approved):

PR #2458, opened 2023-12-29

paulohenriques70:
Updated mostly regarding https://collaboration.dataminer.services/task/223099


user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -267,21 +236,72 @@ plugins.security.system_indices.enabled: true
 plugins.security.system_indices.indices: [".plugins-ml-model", ".plugins-ml-task", ".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state", ".opendistro-reports-*", ".opensearch-notifications-*", ".opensearch-notebooks", ".opensearch-observability", ".opendistro-asynchronous-search-response*", ".replication-metadata-store",".opendistro-anomaly-detection-state", ".opendistro-reports-*", ".opensearch-notifications-*", ".opensearch-notebooks", ".opensearch-observability", ".opendistro-asynchronous-search-response*", ".replication-metadata-store"]
 ```
 
+##### Change the admin password and remove other users
+
+Generate a new hash for the admin user as detailed under [Configure a user](https://opensearch.org/docs/latest/install-and-configure/install-opensearch/debian/#configure-a-user) in the OpenSearch documentation. You will need to remove all the users except *admin* and replace its hash with the one generated from the new password:
```
MariekeGO, line 241: This intro is a little confusing. As a user, I'd think I need to go the linked page in the OpenSearch documentation and do what is described there first, but if I understand it correctly, I'd actually only need to do the procedure detailed below this intro. Could you change this? You could for instance move the reference to the OpenSearch documentation to a "TIP" below the procedure.

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
```diff
@@ -267,21 +236,72 @@ plugins.security.system_indices.enabled: true
...
+
+1. Generate a hash for the new password of the *admin* user:
```
MariekeGO, line 243: This isn't entirely clear to me. Do you need to create a new password first before you do this (if yes, this should still be added to the procedure), or is the creation of the new password included in this step?

MariekeGO (changes requested):

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
paulohenriques70: Indeed, the new password is created in step 1. where a hash key is generated from the new password chosen at the time, and then in step 2., this hash key is added to the yaml file, replacing the old one there.
Basically here we are changing the default password, not entering it in clear text but instead, adding its corresponding hash key. Steps 1. and 2. try to describe this.

user-guide/Advanced_Functionality/Databases/Indexing_database/OpenSearch_Database/Installing_OpenSearch_database.md
paulohenriques70: Agree, it is a bit confusing. I might be wrong, but I have the feeling in other pages and procedures in our documentation, many times we provide the link to some external site to the official documentation and next we start describing in more detail the steps on that external page. May give you some examples later.
I will try to reformulate this section and submit for review.

MariekeGO (approved):

PR #2462, opened 2024-01-02
DOM - Empty list values are now blocked when the field is required

VanRobaysWouter:
[RN38238] [DCP222884]


Skyline-ThomasGH:
I would suggest adding a slightly longer explanation (including the version) to the note under: https://docs.dataminer.services/user-guide/Advanced_Modules/DOM/DOM_objects/DOM_SectionDefinition.html#fielddescriptor. You can then have the notes in those three tables refer to that one. (Will require a header above the note so it can be clearly linked)
E.g. 'Empty list values are equivalent to not providing a `FieldValue`.' where 'Empty list values' is the link for example.

VanRobaysWouter:
> I would suggest adding a slightly longer explanation (including the version) to the note under: https://docs.dataminer.services/user-guide/Advanced_Modules/DOM/DOM_objects/DOM_SectionDefinition.html#fielddescriptor. You can then have the notes in those three tables refer to that one. (Will require a header above the note so it can be clearly linked) E.g. 'Empty list values are equivalent to not providing a `FieldValue`.' where 'Empty list values' is the link for example.

Gave it a try, not sure if a header above the note is the best approach

user-guide/Advanced_Modules/DOM/DOM_objects/DOM_SectionDefinition.md
```diff
@@ -81,6 +81,8 @@ There are also special types of `FieldDescriptors` that are purpose-made to stor
 >
 > Using `FieldDescriptors` that have the same ID in multiple `SectionDefinitions` might result in inconsistent behavior. When a `FieldDescriptor` with the same ID is used in a name definition, the name of the `DomInstance` might differ depending on the first section available in the `DomInstance` that has a value assigned for the `FieldDescriptor` with that ID. During the validation of changes to the `FieldDescriptor`, this might result in DataMiner incorrectly detecting that a `FieldDescriptor` is no longer in use, which may cause the removal of a descriptor that is actually still in use.
 
+### List FieldValues
```
Skyline-ThomasGH, line 84:
```suggestion
### Multiple values
```

Skyline-ThomasGH: The header does indeed feel a bit weird with the note under it. Since that note is quite long, maybe we can just move the content under that header. It seems fitting to me that it get its own section.

Skyline-ThomasGH (approved):
Thanks

MariekeGO (approved):

PR #2472, opened 2024-01-04
dcf, link to new nuget package instead of referring to the old SVN

SimonV-Skyline:


develop/devguide/Connector/AdvancedDcfHelper.md
```diff
@@ -4,6 +4,6 @@ uid: AdvancedDcfHelper
 
 # DCF Helper Class
 
-When implementing DCF, the DCF helper class should be used. The latest version of this helper class and documentation about how to use it are available on SVN.
+When implementing DCF, the DCF helper class should be used. The latest version of this helper class can be found on [the public NuGet store](https://www.nuget.org/packages/Skyline.DataMiner.Core.ConnectivityFramework.Protocol).
 
 Usage of this library requires a minimum DataMiner version of 9.5.0.0-7104.
```
JoaoSE-Skyline, line 9: Is the minimum version still accurate given the usage of NuGets?

SimonV-Skyline: Good point, I now changed it to 10.0.10. I don't think we need to bother explaining what to do when having an older DM version as those are getting out of support anyway.

MariekeGO (approved):

PR #2476, opened 2024-01-04
Change based on 38330

MariekeGO:
Initial correction based on RN 38330. Note that when this RN has been merged to a DataMiner version, we should also add a note here to make sure people using earlier versions do not get confused by the old description with "DataMiner Agent".


HansSolos (approved):

HansSolos:
Gerrit link: https://devcore3/review/#/c/83063/

PR #2483, opened 2024-01-05
Added examples to ConnectionsWebSocketsCommunication.md

ArneMaes0:
Added examples on how to implement websockets commands responses and pairs.

### DIFF
--- a/develop/devguide/Connector/ConnectionsWebSocketsCommunication.md
+++ b/develop/devguide/Connector/ConnectionsWebSocketsCommunication.md
@@ -4,7 +4,7 @@ uid: ConnectionsWebSocketsCommunication
 
 # Communication
 
-Once the WebSocket has been set up, data can be sent and received. Commands are implemented using *Command* constructs, and *Responses* can be implemented using *Response* constructs. This is identical to the way in which serial connections work.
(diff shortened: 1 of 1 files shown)


develop/devguide/Connector/ConnectionsWebSocketsCommunication.md
```diff
@@ -4,8 +4,158 @@ uid: ConnectionsWebSocketsCommunication
 
 # Communication
 
-Once the WebSocket has been set up, data can be sent and received. Commands are implemented using Command constructs and Responses can be implemented using Response constructs, just as is the case with serial connections.
+Once the WebSocket has been set up, data can be sent and received. Commands are implemented using *Command* constructs, and *Responses* can be implemented using *Response* constructs. This is identical to the way in which serial connections work.
```
mmphs, line 7:
```suggestion
Once the WebSocket has been set up, data can be sent and received. Commands are implemented using *Command* constructs, and Responses can be implemented using *Response* constructs. This is identical to the way in which serial connections work.
```
Either make Commands italic or make Responses not italic for both to be inline.

mmphs (changes requested):
Technically everything is correct.
There's one remark that should probably be aligned.

MariekeGO (approved):

PR #2487, opened 2024-01-05
Communication Gateway 2.0.2

mmphs:


MariekeGO:
@mmphs, because these two RNs definitely look more like fixes than enhancements to me, I've changed the header to "Fixes". Could you take a look at my changes and let me know if everything is still OK for you?

mmphs (commented):
Thank you for the changes @MariekeGO, no further remarks from my side.

MariekeGO (approved):

PR #2497, opened 2024-01-09
RN 38041

EdithLansens:
Hi @SkylineDennisDr, could you please review this pull request to confirm the accurate inclusion of release note 38041 in the user guide? Thanks in advance!


SkylineDennisDr:
Hi, looks good, although I would maybe not explicitly mention the 330MB as this is just an estimate which depends on the polling times of the specific parameters that are being tracked.

EdithLansens:
Thanks for the feedback @SkylineDennisDr! I have removed the mention of 330 MB in [commit 7457b7d](https://github.com/SkylineCommunications/dataminer-docs/pull/2497/commits/7457b7d2c167b78d2356cedf4293a7d82d848092).

SkylineDennisDr:
Sorry Edith, should have mentioned this from the start, but the "one hour of trend data" that you mention is actually configurable. The default amount of data that we keep for a state icon is 1 hour, but if the user changes that setting, we can keep more or less. Maybe it is better to just mention that, as an extra safeguard, we only calculate state icons (and thus keep trend data) for at most 250 000 trended parameters per dma.

That will be the last remark! :)

EdithLansens:
> Sorry Edith, should have mentioned this from the start, but the "one hour of trend data" that you mention is actually configurable. The default amount of data that we keep for a state icon is 1 hour, but if the user changes that setting, we can keep more or less. Maybe it is better to just mention that, as an extra safeguard, we only calculate state icons (and thus keep trend data) for at most 250 000 trended parameters per dma.
> 
> That will be the last remark! :)

This has been updated in [commit b375e33](https://github.com/SkylineCommunications/dataminer-docs/pull/2497/commits/b375e3345067a11f82d97484b6bc61b540700f0b).

SkylineDennisDr (approved):

user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Working_with_behavioral_anomaly_detection.md
```diff
@@ -8,6 +8,7 @@ uid: Working_with_behavioral_anomaly_detection
 >
 > - This feature requires [Storage as a Service](xref:STaaS) or a [self-hosted Cassandra-compatible database](xref:Supported_system_data_storage_architectures).
 > - Anomaly detection is only available for numeric parameters that are not part of [partial tables](xref:Table_parameters#partial-tables). It is also limited to at most 100,000 parameters per DMA.
+> - From DataMiner 10.4.2/10.5.0 onwards<!--RN 38041-->, trend data is limited to at most 250,000 trended parameters per DMA.
```
MariekeGO, line 11: The current phrasing could be misinterpreted as applying to trend data in general, rather than to the calculation of state icons by DataMiner Analytics. Perhaps you could instead add this info in the notes under "Trend icons on data pages" (e.g. on line 74 below), and rephrase it to e.g. "As a safeguard, from DataMiner 10.4.2/10.5.0 onwards, the number of parameters for which state icons are calculated is limited to at most 250,000 trended parameters per DMA."

PR #2499, opened 2024-01-09
RN 37734 + 36575

EdithLansens:
Hi @LaurensVergote and @JoaoSE-Skyline, could you please review this pull request to confirm the accurate inclusion of release notes 37734 and 36575 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Reference/DataMiner_Extension_Modules.md
+++ b/user-guide/Reference/DataMiner_Extension_Modules.md
@@ -20,7 +20,7 @@ For ease of installation, DxMs may be packaged together in a pack, such as the [
 
 The APIGateway module is the endpoint for the gRPC communication with a DataMiner Agent.
 
-From DataMiner 10.4.2/10.5.0 onwards<!--RN 37734 + 36575-->, DataMiner modules can register with APIGateway. All modules registered with APIGateway are displayed under `/APIGateway/api/version`, showing the following properties:
+From DataMiner 10.4.2/10.5.0 onwards<!--RN 37734 + 36575-->, DataMiner modules can register with APIGateway. All modules registered with APIGateway are displayed under `https://[Your DMA]/APIGateway/api/version`, showing the following properties:
 
 - Name
 - Version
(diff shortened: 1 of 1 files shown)


user-guide/Reference/DataMiner_Extension_Modules.md
```diff
@@ -20,6 +20,12 @@ For ease of installation, DxMs may be packaged together in a pack, such as the [
 
 The APIGateway module is the endpoint for the gRPC communication with a DataMiner Agent.
 
+From DataMiner 10.4.2/10.5.0 onwards<!--RN 37734 + 36575-->, DataMiner modules can register with APIGateway. All modules registered with APIGateway are displayed under `/APIGateway/api/version`, showing the following properties:
+
+- Name
+- Version
```
JoaoSE-Skyline, line 26: @BertVandenberghe I recall you mentioning something about the security aspect of exposing version information.
Should the version be left out of the documentation?

user-guide/Reference/DataMiner_Extension_Modules.md
```diff
@@ -20,6 +20,12 @@ For ease of installation, DxMs may be packaged together in a pack, such as the [
 
 The APIGateway module is the endpoint for the gRPC communication with a DataMiner Agent.
 
+From DataMiner 10.4.2/10.5.0 onwards<!--RN 37734 + 36575-->, DataMiner modules can register with APIGateway. All modules registered with APIGateway are displayed under `/APIGateway/api/version`, showing the following properties:
```
JoaoSE-Skyline, line 23: It is yet to be merged but it is likely it will also become available on 10.4.0MR together with 10.4.2FR

LaurensVergote:
Looks good to me

user-guide/Reference/DataMiner_Extension_Modules.md
BertVandenberghe: No, the version can be left in. The documentation can indicate in which version which features have been added. That's not a concern. An API should not expose the version via an unauthenticated method because that would be considered 'information leakage'. This can be used to query if systems are running outdated versions which could maybe contain vulnerabilities. But documentation can mention which version now includes a new feature, no problem.

user-guide/Reference/DataMiner_Extension_Modules.md
MariekeGO, line 23: Could you clarify that `/APIGateway/api/version` is actually `http(s)://[DMA IP or hostname]/APIGateway/api/version`? Otherwise, some users might not know what we mean here.

MariekeGO (approved):

PR #2503, opened 2024-01-10
RN 38034

EdithLansens:
Hi @tobedeprez, could you please review this pull request to confirm the accurate inclusion of release note 38034 in the user guide? Thanks in advance!


tobedeprez:
Although the remark is more or less correct, I think it is a bit confusing to put it here. The change was about a Cube option 'Automatically remove cleared alarms' (see https://docs.dataminer.services/user-guide/Basic_Functionality/Alarms/Working_with_alarms/Working_with_the_Alarm_Console/Alarm_Console_Settings.html) only. This means that nothing has changed about whether alarms are actually cleared. The only change is that when an open alarm console tab was showing that alarm, it would not get removed automatically from the tab when it was cleared. However, when opening a new tab (not showing history alarms), the alarm would not show up again.

Hence, it is probably better to put the remark on the 'Alarm Console Settings' page mentioned above. Also, reword the first sentence to reflect that it is about removing cleared alarms from the console, and not about whether the alarms actually get cleared.

tobedeprez:
Look OK now from my side.

MariekeGO (approved):

PR #2514, opened 2024-01-11
DaaS and STaaS adjustments

MariekeGO:


user-guide/Advanced_Functionality/Databases/STaaS/STaaS.md
```diff
@@ -68,9 +68,9 @@ For a self-hosted DataMiner System, follow the steps below to set up STaaS.
...
+- **Geo-redundant storage (GRS)** copies your data synchronously three times within a single physical location in the primary region and then also copies your data asynchronously to a single physical location in the secondary region. Only specific regions can be combined in such a setup, e.g. if the primary region is Switzerland North, the secondary region can only be Switzerland West. For an overview of the supported regions, see [Azure paired regions](https://learn.microsoft.com/en-us/azure/reliability/cross-region-replication-azure#azure-paired-regions). If you wish to use DataMiner STaaS with GRS, contact <daas@dataminer.services>.
```
MariekeGO, line 73: @BertVandenberghe, I've used the daas@dataminer.services email address based on your feedback; however, everywhere else on this page we refer to staas@dataminer.services. Is this OK as it is, or should it be the STaaS address here as well?

BertVandenberghe: Sorry, you should then indeed best use staas@dataminer.services here, my bad. Both addresses exist and they will both help you out, but it is more correct to use the staas one here because we are talking about a STaaS config which could be used even without DaaS.

PR #2524, opened 2024-01-12
Update Protocol.Params.Param.Display.Units.md

MichielClepkensSkyline:
Made it more clear that the unit options are not limited to the ones in the validation list.

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Params.Param.Display.Units.md
+++ b/develop/schemadoc/Protocol/Protocol.Params.Param.Display.Units.md
@@ -12,6 +12,8 @@ Contains a unit of measurement.
 
 In the table below, you can find a list of the suggested units to use. These are the units of measurement recognized by the validator in DIS. However, you can also use units that are not mentioned in this list if necessary. For example, a string like "Batteries" can be used to show the number of batteries attached to a backup power supply.
 
+In case you have a suggestion for a unit to be added to the list below, send an email to <DIS@skyline.be>.
+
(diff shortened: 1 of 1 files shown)


develop/schemadoc/Protocol/Protocol.Params.Param.Display.Units.md
```diff
@@ -4,11 +4,13 @@ uid: Protocol.Params.Param.Display.Units
...
 
-Specifies a unit for a parameter value displayed on the user interface.
+Specifies a unit for a parameter value displayed in the user interface (e.g. in a report, in the Alarm Console, etc.).
 
 ## Type
 
-Contains one of the following values:
+Contains a unit of measurement.
+
+In the table below, you can find a list of the suggested units to use. These are the units of measurement recognized by the validator in DIS. However, you can also use units that are not mentioned in this list if necessary. For example, a string like "Batteries" can be used to show the number of batteries attached to a backup power supply.
```
PedroDebevere, line 13:
```suggestion
In the table below, you can find a list of the suggested units to use. These are the units of measurement recognized by the validator in DIS. However, you can also use units that are not mentioned in this list if necessary. For example, a string like "Batteries" can be used to show the number of batteries attached to a backup power supply.

In case you have a suggestion for a unit to be added to the list below, you can send an email to DIS@skyline.be
```

PedroDebevere (approved):

MariekeGO (approved):

PR #2528, opened 2024-01-12
Add DataAPI soft launch option

JoaoSE-Skyline:
@jaklke

### DIFF
--- a/user-guide/Reference/Soft-launch_options/Overview_of_Soft_Launch_Options.md
+++ b/user-guide/Reference/Soft-launch_options/Overview_of_Soft_Launch_Options.md
@@ -118,9 +118,7 @@ This is a Cube-only feature that can only be activated with the argument `ENABLE
 
 ### DataAPI
 
-Enables Data API functionality and scripted connectors.
-
-This displays the Data Sources module in DataMiner Cube and enables the functionality of Data API and Data Sources.
+Enables Data API functionality and scripted connectors. This displays the Data Sources module in DataMiner Cube.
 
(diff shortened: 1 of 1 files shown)


jaklke (approved):

user-guide/Reference/Soft-launch_options/Overview_of_Soft_Launch_Options.md
```diff
@@ -116,6 +116,18 @@ This is a Cube-only feature that can only be activated with the argument `ENABLE
 - **Minimum version**: 9.6.7
 - **Estimated Release version**: To be determined
 
+### DataAPI
+
+Enables Data API functionality and scripted connectors.
+
+This displays the Data Sources module in DataMiner Cube and enables the functionality of Data API and Data Sources.
```
MariekeGO, line 123: It seems to me that the second part of this sentence ("enables the functionality of Data API and Data Sources") means the same as the sentence above it. I'd suggest to either remove this, because there's little point in saying the same thing twice in a row, or to change it to a more detailed explanation of what exactly we mean with "the functionality of Data API".

jaklke: The plan is to introduce specific documentation explaining the functionality of Data API. In the future we could link to these pages for the explanation. For now I propose to write "This displays the Data Sources module in DataMiner Cube". Later this could be rewritten with a link to the documentation.

MariekeGO (approved):

PR #2529, opened 2024-01-12
GQI - Best practices & caveats when developing extensions

skyline-rge:
Added a page with some initial best practices & caveats when developing GQI extensions (Ad hoc data sources and custom operators). This page will be expanded further as best practices are formed and according to feedback and struggles we receive.


WaHaeck:
Looks good to me!

MariekeGO:
It would be great if we could also add a tip with a "see also" reference to this new page on other pages where this might be useful (e.g. on [Configuring an ad hoc data source in a query](https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html)?). Could you check where this might fit and add such a tip there?

PR #2540, opened 2024-01-16
Modern template introduction

PedroDebevere:
Reason for change:
1.	Docfx recommends using this new template: _We recommend using the modern template that matches the look and feel of this site. It supports dark mode, more features, rich customization options_ . ([https://dotnet.github.io/docfx/docs/template.html?tabs=modern](https://dotnet.github.io/docfx/docs/template.html?tabs=modern)).
2.	Things we added previously by customizing the default template are now supported by default in the modern template (e.g. Mermaid support, copy functionality).
3.	Nieuwe things such as MathJax support (allowing to write mathematical equations in markdown) are only provided by the modern template.


MariekeGO:
A couple of things: 

- I noticed that "Propose changes" is now all the way at the bottom of each page instead of in the sidebar on the right, which makes it less user-friendly, especially on long pages. Can this be moved back up into the sidebar on the right?

- The TOC on the right is now automatically fully expanded, which makes it less user-friendly for very long pages. Is this something that can be tweaked?

  image

- When you scroll down on a long page, the header bar now disappears after a while, and you may need to scroll up a bit to see it again, which is not that user-friendly for people who want to use the search. Is this something that can be prevented?

PedroDebevere:
- updated position of proposed changes link to where it was on the default template.
- toc behavior in the modern template is different from the default. In the default template it was showing where you were currently are in the article. Now it's just a TOC. Didn't find a way to change it.
- The header bar should now be fixed at the top and no longer disappear.

BertVandenberghe (approved):
Difficult to review if there is no site available to see the new result, so I hope this has been tested thoroughly...

MariekeGO:
> Difficult to review if there is no site available to see the new result, so I hope this has been tested thoroughly...

I hear you've been able to check a test version generated by Pedro. OK to merge based on that?

MariekeGO (approved):

PR #2547, opened 2024-01-17
Change order hosting options

EdithLansens:
Changed the order of the available hosting options, as requested by @BertVandenberghe.


BertVandenberghe:
Nice! Can we also have the same order in the TOC? I mean the tree structure on the left hand side? Thanks.

PR #2549, opened 2024-01-17
Update Hosting_DaaS_create.png

EdithLansens:
@BertVandenberghe, is it OK to replace the current final image on the [DataMiner as a Service: DataMiner nodes and storage nodes as a service](https://docs.dataminer.services/dataminer-overview/DataMiner_Hosting_and_HA/DaaS_hosting.html) page with this newly recreated image in the updated style? I've recreated the image to match the old one precisely, but with the updated style.


BertVandenberghe (approved):
Yes indeed, well spotted!

PR #2580, opened 2024-01-22
Retro recipes

EdithLansens:
As announced on [Dojo](https://community.dataminer.services/send-us-your-own-retro-recipe/), "submitting a retro recipe" is now an additional way to earn DevOps Points.


EdithLansens:
@SammyDW  FYI, a new option to earn DevOps Points was added in DataMiner Overview.

PR #2605, opened 2024-01-26
Dms obsolete

SimonV-Skyline:


SimonV-Skyline:
This should be double checked by @PedroDebevere.
Different things:
- Interop.SLDms: OK
    - DMS: OK but somehow the warning doesn't get added to the top of the page like it does for Interop.SLDms page. Not sure if you want to further look into improving that but otherwise, it's already better than it was.
    - DMSClass:  OK but somehow the warning doesn't get added to the top of the page like it does for Interop.SLDms page. Not sure if you want to further look into improving that but otherwise, it's already better than it was.
          - Methods
                - GetInfo: for this one, I also tried to add a warning but it didn't work, feel free to either give it another go if you have other ideas, if not, let's remove that overwritte.
                - Notify: Same as for GetInfo but here the uid is a bit different so I gave it another shot but failing as well.
    - IDMS:  OK but somehow the warning doesn't get added to the top of the page like it does for Interop.SLDms page. Not sure if you want to further look into improving that but otherwise, it's already better than it was.

PedroDebevere:
updated overwrite files

PedroDebevere (approved):

spetebroodt (approved):

PR #2610, opened 2024-01-29
Update Stack_Augmented_Operations.md

VeerleLedoux:
-Some new material has been added to the page

-Made consistent with the DataMiner Augmented Operations training Ben has in mind. This training will be worked out by us in the coming weeks, a start has been given by Ben here: https://skylinebe-my.sharepoint.com/:p:/g/personal/ben_vandenberghe_skyline_be/EUFU5yhHc1JPiOhP-FNy3L8BHqEgTcSvH9Dm_puH0FPqfQ?e=VKIaab

### DIFF
--- a/dataminer-overview/DataMiner_Stack_Functions/Stack_Augmented_Operations.md
+++ b/dataminer-overview/DataMiner_Stack_Functions/Stack_Augmented_Operations.md
@@ -10,19 +10,14 @@ uid: Stack_Augmented_Operations
 
 **Make the shift from reactive to proactive management.**
 
-The DataMiner Augmented Operations technology **captures, correlates, and analyzes** all data sources in **real time**. It learns the relationships between metrics and generates alarms when anomalies are detected. This predictive approach improves performance and reduces system downtime.
+DataMiner Augmented Operations continuously and autonomously learn from your data, making DataMiner increasingly intelligent. 
 
-This technology, built into the core of DataMiner, leverages state-of-the-art big data and artificial intelligence (AI) technology. As we are dealing with **increasingly more complex and agile ecosystems that are continuously evolving**, the way things are set up today could be very different as opposed to how they are set up tomorrow.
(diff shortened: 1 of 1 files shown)


MariekeGO:
@SammyDW, could you take a look at this PR with changes for the DataMiner Overview section, and confirm if these are OK for you?

SammyDW (approved):
Read through the changes, some of the Augmented Operations namings got changed or shuffled, and one got added, which I'd assume are all OK since based on inputs from Veerle from AI team, so for me all good to go. But we need to make sure we stay consistent with the lingo on all places, eg https://skyline.be/dataminer/augmented-operation is to be updated. And I believe we should think of removing https://community.dataminer.services/dataminer-augmented-operation/ , this was another approach showing the effectively existing modules back then (reduced set as opposed to what was depicted on the Skyline website from marketing point of view. Since  https://community.dataminer.services/core-features/ as a whole is getting outdated, and being replaced by https://docs.dataminer.services/dataminer-overview/DataMiner_Stack_Functions/Overview_stack_functions.html, should we remove the former from Dojo?

MariekeGO:
> @MariekeGO , @VeerleLedoux , I read through the changes, some of the Augmented Operations namings got changed or shuffled, and one got added, which I'd assume are all OK since based on inputs from Veerle from AI team, so for me all good to go. But we need to make sure we stay consistent with the lingo on all places, eg https://skyline.be/dataminer/augmented-operation is to be updated. And I believe we should think of removing https://community.dataminer.services/dataminer-augmented-operation/ , this was another approach showing the effectively existing modules back then (reduced set as opposed to what was depicted on the Skyline website from marketing point of view. Since https://community.dataminer.services/core-features/ as a whole is getting outdated, and being replaced by https://docs.dataminer.services/dataminer-overview/DataMiner_Stack_Functions/Overview_stack_functions.html, should we remove the former from Dojo?

Since this pull request is already OK as it is, I will already merge it. I will address the comments in a future PR.

MariekeGO (approved):

PR #2669, opened 2024-02-07
Update AdvancedDataMinerDataPersistenceNoSqlCassandra.md due to RN38184

SkylineDennisDr:
The analytics_alarmfocus table was replaced by the ai_alarmfocus table. This contains the same field, except for the fact that the "dn" column was removed.


MariekeGO:
This change has already been implemented by #2672.

PR #2679, opened 2024-02-09
RN 38325

MariekeGO:


baptistepa (approved):
Only a small change but could be more clear towards users as to why we do not support it on DaaS.

PR #2683, opened 2024-02-09
Update Use_Github_Guidelines.md

SimonV-Skyline:


develop/CICD/Skyline Communications/Github/Use_Github_Guidelines.md
MariekeGO, line 1: As the info in parentheses is getting longer and longer, I'm thinking of changing this list to a table, with the columns "Item type" and "Description". Would this be OK for you?

SimonV-Skyline: yes, that's perfectly fine and will indeed be more readable, I would suggest the following column titles though:
1. Syntax or Abbreviation? (containing the abbreviation)
2. Type or description (containing the non-abbreviated description + optionally the extra details)
3. Extra details (optionally containing the extra details)

Since only a few items have extra details, I'm hesitating between having only 2 columns and adding extra details to the type/description or splitting it into a 3rd "Extra details" column. I'll leave that up to you :)

MariekeGO: Two columns is easier to maintain, so I've gone for that. Could you check if everything is OK for you?

MariekeGO (approved):

PR #2710, opened 2024-02-14
Update dataminer.services change log 38771

AlexanderVK-SLC:
DMS connection status visible on dataminer.services

### DIFF
--- a/release-notes/Cloud/DCP_change_log.md
+++ b/release-notes/Cloud/DCP_change_log.md
@@ -13,6 +13,10 @@ The dataminer.services platform gets updated continuously. This change log can h
 
 The versions of a catalog record will now be sorted correctly.
 
+#### 9 February 2024 - Enhancement - DMS cloud connection status now visible on dataminer.services [ID_38771]
+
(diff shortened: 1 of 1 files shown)


release-notes/Cloud/DCP_change_log.md
```diff
@@ -13,6 +13,10 @@ The dataminer.services platform gets updated continuously. This change log can h
 
 The versions of a catalog record will now be sorted correctly.
 
+#### 9 February 2024 - Enhancement - DMS cloud connection status now visible on dataminer.services [ID_38771]
+
+​The cloud connection status is now shown for a DMS on the dataminer.services page, presenting an OK, warning, error or unknown indicator and disabling the app buttons depending on the state of the system.
```
MariekeGO, line 18: Since we should be careful about using the term "cloud" (given that other cloud providers exist, so "cloud connection" is an ambiguous term), I'd propose the following change: 

```suggestion
#### 9 February 2024 - Enhancement - DMS connection status now visible on dataminer.services [ID_38771]

On the dataminer.services page, you can now see the status of the connection of a DMS to dataminer.services. The status is indicated as *OK*, *Warning*, *Error*, or *Unknown*. If the connection is not available, the app buttons will be disabled.
```

MariekeGO (approved):

PR #2715, opened 2024-02-14
Protocol change impact

MariekeGO:

PR #2723, opened 2024-02-15
Protocol change impact

alexjohnson1993:
Here are some changes I made based on the comments


MariekeGO:
Changes have been implemented in #2715 instead, because this PR was made based on an outdated fork.

PR #2741, opened 2024-02-19
Breaking changes cross ref + DCF changes

SimonV-Skyline:


develop/devguide/Connector/ImpactOfProtocolChanges/ProtocolChanges/ChangeDCF.md
```diff
@@ -4,17 +4,15 @@ uid: ChangeDCF
...
-
-This includes adding DCF support to a protocol where this was previously not available or making changes to existing DCF interfaces.
+Changes to existing DCF interfaces and/or connections are typically considered a major change, and as such a major change is required in the version numbering.
 
 ## Impact
 
 ### Adding DCF support to an existing protocol
 
-DCF can add a lot of additional load to a system, and it may introduce new issues while it might not actually be used.
+As such, adding DCF support to an existing protocol is **not** to be seen as a breaking change. However, note that depending on the use-case, adding DCF may or may not have a major impact on performance and resource consumptions and it's only in that aspect that adding DCF can sometimes be considered a breaking change. This is the case when, for example, linking creation of DCF interfaces on a very big and/or a very dynamic table.
```
MariekeGO, line 13:
```suggestion
Adding DCF support to an existing protocol is usually **not** considered to be a breaking change. However, note that depending on the use case, adding DCF may have a major impact on performance and resource consumption, which may cause it to be considered a breaking change after all. This is for example the case when creation of DCF interfaces is linked to a very big and/or a very dynamic table.
```

MariekeGO (approved):

PR #2753, opened 2024-02-21
User-Guide Documentation for Smart Trap Processor

ThomasRich-Skyline:
This commit contains the user-guide materials for the Smart Trap Processor Solution located under the monitoring and management solutions.


user-guide/Standard_Apps/Monitoring_Solutions/Smart_Trap_Processor/Using_Smart_Trap_Processor/Processor_Configuration.md
```diff
@@ -0,0 +1,167 @@
...
+
+# Configuring processing rules
+
+To configure a processor element, you can either use DataMiner Cube or the Smart Trap Processor tool (on the *Processor Configuration* page). The configuration described below is the same regardless of which interface is used.
+
+## Trap IP Sources configuration
+
+Before you define any rules for trap processing, you first have to specify the IP address from which traps will be sent that will be received by the DataMiner Agent hosting the processor element.
+
+To do so, set the *Trap IP Sources* parameter to the IP address corresponding with the source of the SNMP traps sent to the DataMiner System.
```
MariekeGO, line 13: Based on what is mentioned below (line 120), it appears that multiple IP addresses can be specified here. Which separator should be used for this?

MariekeGO (commented):
The text mentions the "Processed Messages" table in many places, but in the UI, I see this is called the "Processed Traps" table instead. Shouldn't this be the "Processed Traps" table in the documentation as well?

user-guide/Standard_Apps/Monitoring_Solutions/Smart_Trap_Processor/Using_Smart_Trap_Processor/Processor_Configuration.md
ThomasRich-Skyline: Hi Marieke, I agree that the naming of the processed table should be consistent.
on the Generic Trap Processor Protocol, the name of the table is "Processed Messages".
On the app, the page where the processed data is observed is called "Processed Traps". The data here comes from the same "Processed Messages" table on the protocol. So they essentially mean the same thing.
I would be ok with sticking with "Processed Traps" across the documentation.

For the multiple IP address input, the separator is a comma.

PR #2766, opened 2024-02-22
Update Stack_SLA_Management.md

SammyDW:
Missing title "SLA Management" - corrected


dataminer-overview/DataMiner_Stack_Functions/Stack_SLA_Management.md
```diff
@@ -8,7 +8,9 @@ uid: Stack_SLA_Management
 
 ![stack.functions.categories](~/dataminer-overview/images/stack_SLA_management.png)
 
-Automate, track, and measure the digital processes that a key relationship depends on.
+## SLA Management
```
MariekeGO, line 11: It's better not to use the exact same header twice in the same document. It looks odd, and it makes it impossible to make a link to the second iteration of the header. Is a different solution possible here?

MariekeGO (approved):

PR #2768, opened 2024-02-22
IDP 1.5.0 release notes

MariekeGO:


MariekeGO:
@jaklke, I included RN 38530 because it is listed as one of the IDP 1.5.0 RNs, but this seems more like a change to 10.4.0/10.4.3 than an IDP change. Wouldn't it be better to include the RN there instead (possibly with a note referring to it here?)? 
Also note that I have not included RN 38784, as this RN is set to "Internal". If this should be included after all, let me know.

jaklke:
Indeed RN 38530 is about 10.4.0/10.4.3. I like your suggestion to add it under the general release notes and refer to it from the IDP RNs.
Like you mention, RN38784 is internal and should not be included.

jaklke (approved):

PR #2780, opened 2024-02-23
Update UD_APIs_Triggering_an_API.md

baptistepa:
Add warning for not using http.


baptistepa:
Related to the following question on DOJO: https://community.dataminer.services/question/user-defined-api-http-vs-https/

MariekeGO (approved):

PR #2738, opened 2024-02-19
Data Aggregator RNs update

MariekeGO:

PR #2822, opened 2024-03-05
Add Data Aggregator DxM RNs for 3.0.X releases

JoaoSE-Skyline:
@wimme @jaklke @PedroDebevere


MariekeGO:
Changes should be added to the change log instead, implemented by #2738.

PR #2843, opened 2024-03-06
DataMiner Overview homepage

EdithLansens:
Updated the DataMiner Overview homepage so that the layout is similar to that of the user guide.


EdithLansens:
Is this OK to merge, @SammyDW?

SammyDW:
@EdithLansens , yes, as confirmed on chat, looks good to me with the new icons, can be merged. Thanks.

PR #2849, opened 2024-03-08
Web client metric types

WoutMH:
New web client metrics introduced with [RN39000](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=39000)


MariekeGO:
We usually don't put this much detailed information about logging in the user guide. Is this something that will be useful for our users, or is this meant more for reference internally? If it's the latter, it might be better to add this on internaldocs instead.

WoutMH:
Moved the info about the metrics. The updated size of the logfile can stay

MariekeGO (approved):

PR #2853, opened 2024-03-11
Update Failover_AWS.md

ThibaultHeylen:
STaaS has nothing to do with failover on AWS. Should not be shown here.


MariekeGO:
As far as I understand it, this procedure is meant for cases where you "use Amazon Web Services (AWS) for your system database", which is a deprecated setup. If this is indeed the case, I'd prefer to keep the disclaimer at the top, because people definitely should not configure a system with a deprecated database setup — though possibly we may need to rephrase it to be more clear.

MariekeGO (approved):

PR #2868, opened 2024-03-13
RN 38811

EdithLansens:


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/Basic_controls/DropdownFeed.md
```diff
@@ -33,3 +38,5 @@ To configure the component:
      > - Prior to DataMiner 10.3.6/10.4.0<!--  RN 35984 -->, this setting is called *Feed Defaults* instead.
 
    - *Clear selection*: Available from DataMiner 10.3.0 [CU13]/10.4.0 [CU1]/10.4.4 onwards<!--RN 38758-->. Allows you to clear the selected values in the component by clicking the X button in the top-right corner of the filter box. This setting is disabled by default.
+
+   - *Data retrieval > Update data*: Available from DataMiner 10.3.0 [CU13]/10.4.0 [CU1]/10.4.4 onwards<!--RN 38811-->. Allows the values displayed in the component to be updated in real time, when the data supports this. This setting is only available when a single query data feed or an index data set was applied. This setting is disabled by default.
```
EdithLansens, line 42: Could you provide an explanation of what the "index data set" entails and how it is used, @GillesBara?

EdithLansens:
Since which DataMiner version do the dropdown, tree, and list components support queries as data input, @GillesBara?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/Basic_controls/DropdownFeed.md
GillesBara: You can drop the entire 'indices' dataset from the data panel on the dropdown as data. When adding an element parameter or (protocol parameter and element) as filter, the dropdown will contain the indices for that parameter. Up to now those indices in the dropdown would always update in realtime, but with this change the updates can also be disabled with the new setting.

GillesBara:
> Since which DataMiner version do the dropdown, tree, and list components support queries as data input, @GillesBara?

Since 10.0.8 you can drop queries on those visuals, which would just list all the different queries. From this change onwards, it will list the resulting values for the query, when there would be only one query.

GillesBara (approved):

MariekeGO (approved):

PR #2871, opened 2024-03-13
Update Data_Aggregator_Migrator.md

miguel-slc:
Additional note when running this migrator tool

### DIFF
--- a/user-guide/Reference/DataMiner_Tools/Data_Aggregator_Migrator.md
+++ b/user-guide/Reference/DataMiner_Tools/Data_Aggregator_Migrator.md
@@ -4,15 +4,12 @@ uid: Data_Aggregator_Migrator
 
 # Data Aggregator Migrator
 
-The Data Aggregator Migrator should be used after upgrading the Data Aggregator module from version 2.x.x to 3.0.0 or higher. This **one-time migration process** is only necessary if the prior installation included [configured GQI queries](xref:Data_Aggregator_queries).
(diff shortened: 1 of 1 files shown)


user-guide/Reference/DataMiner_Tools/Data_Aggregator_Migrator.md
```diff
@@ -11,6 +11,9 @@ The Data Aggregator Migrator should be used after upgrading the Data Aggregator
 > [!TIP]
 > See also: [Data Aggregator](xref:Data_Aggregator_DxM)
 
+> [!NOTE]
+> This tool should be used after upgrading the DataAggregator DxM from 2.x.x to 3.x.x or higher.
```
MariekeGO, line 15: Since this note repeats what is already said in the introduction paragraph, I'm wondering if we couldn't improve that introduction instead (as I prefer not to repeat the same content exactly if we can avoid it). Perhaps we could adjust the text a bit and put some things in bold? For example:
_The Data Aggregator Migrator should be used **after** upgrading the Data Aggregator DxM from version **2.x.x to 3.x.x** or higher._

miguel-slc: Hi @MariekeGO, indeed it makes sense to adjust the text.

MariekeGO: Adjusted accordingly!

MariekeGO (approved):

PR #2902, opened 2024-03-19
First and Second Line Support

EdithLansens:
Is this OK to merge, @SammyDW?


SammyDW (changes requested):
We better place it under the Partner Program, instead of the FAQ about Support Services, the latter being aimed towards end users. This text is about partner responsibilities towards end users, for L1 and L2 support. Refer to my mail with more details about the exact location.

PR #2913, opened 2024-03-20
Add example values to device simulator schema

SkylineThomasCR:
Certain values, like EngineID require a very specific format. By providing an example value, we don't need to describe the specifics of that format.


MariekeGO:
I'd prefer not to add another column to this table if we can avoid it, because it won't look very nice. Also, when users scroll down, they won't see the table header anymore, so a user in a hurry might misinterpret the example value as a default value. Could we just add the examples to the description instead? E.g. for the first line:

```
|Name |1 |Mandatory |An identifier for the simulation agent. Can be an empty value. For example: `MyName`.|
```

MariekeGO (approved):

PR #2919, opened 2024-03-21
Creating your first element on a DaaS system

spetebroodt:


tutorials/images/Tutorials_Getting_Started.svg
EdithLansens, line 1: I'm somewhat unsure about my choice of skylicon. I was first going to use the [light bulb](https://skyline.be/skylicons/duotone/Solution2_Duo_Light.png) icon, but it looked a bit strange with the text balloon  partially covering it. As a second option, I opted for this icon with "learning to walk before you can run" in mind, which fits the vibe of the getting started tutorials (in my eyes). Feel free to suggest [a different skylicon](https://skyline.be/skylicons/) if you find a better option, @spetebroodt and @MariekeGO.

tutorials/Tutorials.md
```diff
@@ -30,23 +30,29 @@ uid: Tutorials
 
 <div class="row"> 
   <div class="column">
-    <a href="/tutorials/IDP_Tutorials.html" title="IDP tutorials" target="_self"><img src="~/tutorials/images/Tutorials_IDP.svg" style="width:100%"></a>
+    <a href="/tutorials/Getting_Started_Tutorials.html" title="Getting started tutorials" target="_self"><img src="~/tutorials/images/Tutorials_Getting_Started.svg" style="width:100%"></a>
```
MariekeGO, line 33: Perhaps we should move this so the getting started icon is the very first one users see? It won't be alphabetical then, but I think it might be more user-friendly that way. (The order in the TOC would then also need to be adjusted.)

PR #2936, opened 2024-03-25
Update Support_services_agreement.md

michelhofmans:
I left out some legal sections. This document should be purely about the terms and conditions for providing support.

### DIFF
--- a/dataminer-overview/FAQ/FAQs_Support_Services.md
+++ b/dataminer-overview/FAQ/FAQs_Support_Services.md
@@ -26,7 +26,7 @@ uid: FAQs_Support_Services
 
   Skyline will provide you with a renewal proposal for the support services at least three months prior to the end date of your support plan. Unless you provide Skyline with a written notice of your wish not to renew the support plan no later than thirty days prior to the end date of the plan, it is automatically renewed for another year, and Skyline will invoice the amount of the renewal proposal. All amounts due to Skyline are payable up front before the provision of any support services. If you terminate the support services early, this does not entitle you to a refund.
 
-  In case the support plan has not been deactivated, and you fail to timely issue pay your invoice(s) or a PO (if required), on the day after the end date of the plan, the service may be stopped along with all related services linked to it. However, you can still continue to consult content and resources on the Dojo Community portal.
+  In case the support plan has not been deactivated, and you fail to pay your invoice(s) or issue a PO (if required) in time, on the day after the end date of the plan, the service may be stopped along with all related services linked to it. However, you can still continue to consult content and resources on the Dojo Community portal.
 
 - **What happens if I do not have active support on a perpetual support plan?**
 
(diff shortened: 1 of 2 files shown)


dataminer-overview/Support Services/Support_services_agreement.md
```diff
@@ -99,35 +97,12 @@ Skyline shall have no obligation of any kind to provide Support for Issues cause
 
 If Skyline determines that an Issue is a User-Generated Issue, Skyline will notify the User as soon as reasonably possible. If the User requests Support for the User-Generated Issue via a confirming email, Skyline shall have the right to invoice the User at Skyline's then current time and materials rates for any such Support provided by Skyline.
 
-## Pricing - Term
```
MariekeGO, line 102: Is the information in this section obsolete? If not, we will have to mention this elsewhere, because I don't think this info on pricing for the support services is mentioned under [Pricing](https://docs.dataminer.services/dataminer-overview/Pricing/Pricing.html) right now. The [FAQs](https://docs.dataminer.services/dataminer-overview/FAQ/FAQs_Support_Services.html) refer to this section here instead, so removing it causes an unresolved cross-reference.

michelhofmans: Thank you for bringing this up, Marieke. The information is not obsolete (yet). The Pricing page doesn't seem like the appropriate place to move this to; I'd rather move it directly to the FAQ page. We could add it as the first bullet point.

MariekeGO: Ok, I'll adjust this.

MariekeGO:
@michelhofmans, I've rephrased the sections about the pricing and data protection and moved them to the support services FAQs. Could you check if everything is still OK?

@SammyDW, tagging you as well so you are aware of this change.

michelhofmans (commented):
made some changes, but not sure if you can see them?

dataminer-overview/FAQ/FAQs_Support_Services.md
```diff
@@ -22,9 +22,11 @@ uid: FAQs_Support_Services
...
+  Skyline will provide you with a renewal proposal for the support services at least three months prior to the end date of your support plan. Unless you provide Skyline with a written notice of your wish not to renew the support plan no later than thirty days prior to the end date of the plan, it is automatically renewed for another year, and Skyline will invoice the amount of the renewal proposal. All amounts due to Skyline are payable up front before the provision of any support services. If you terminate the support services early, this does not entitle you to a refund.
+
+  In case the support plan has not been deactivated, and you fail to timely issue pay your invoice(s) or a PO (if required), on the day after the end date of the plan, the service may be stopped along with all related services linked to it. However, you can still continue to consult content and resources on the Dojo Community portal.
```
MariekeGO, line 29: Something seems off about this sentence. Can I rephrase it like this?

```suggestion
  In case the support plan has not been deactivated, and you fail to pay your invoice(s) or issue a PO (if required) in time, on the day after the end date of the plan, the service may be stopped along with all related services linked to it. However, you can still continue to consult content and resources on the Dojo Community portal.
```

michelhofmans: perfect Marieke, thank you

MariekeGO (approved):

PR #2940, opened 2024-03-26
Added User-Defined APIs Events

VanRobaysWouter:
[RN39117] [DCP205952]


Skyline-ThomasGH (approved):

user-guide/Advanced_Modules/User-Defined_APIs/UD_APIs_Events.md
```diff
@@ -0,0 +1,40 @@
...
+
+```csharp
+// In this example we take the Connection object from the script's Engine object
+var connection = engine.GetUserConnection();
+
+// Create a random set ID that identifies our subscription
+var setId = $"ApiTokenSubscription_{Guid.NewGuid()}"
+
+// Create the filter for the ApiToken events, only enabled tokens should match
+var filter = ApiTokenExposers.IsDisabled.Equal(true);
```
VanRobaysWouter, line 29: Only enabled tokens should match
```suggestion
var filter = ApiTokenExposers.IsDisabled.Equal(false);
```

MariekeGO (approved):

PR #2954, opened 2024-03-26
updates link to web app

juliori85:


MariekeGO (commented):
I've tried the "https://supportpreview.dataminer.services/" URL, but this does not seem to work yet. Could it be that this URL is not available yet to all users?

MariekeGO (approved):

PR #2966, opened 2024-03-28

ThibaultHeylen:
Removed option, as this is deprecated and shouldn't be configured


SkylineCookieMonster (approved):
Note: I'm not an expert in OpenSearch/Elastic. It would be better to have the SME for OpenSearch/Elastic review this one.

ThibaultHeylen:
I discussed this with Simon D, this is an option that is deprecated in Elastic 7.0, so also for opensearch.
Besides this the option is only when you use multiple node's on a single server, eg docker.

MariekeGO (approved):

PR #2986, opened 2024-04-02
Update Paneda_Multiplexer.md

JochenDewachter92:
Wrong firmware version format


MariekeGO:
This change should be implemented in the [dataminer-docs-connectors](https://github.com/SkylineCommunications/dataminer-docs-connectors) repo instead. Changing this old branch on docs will not make the change available to our users.

PR #2997, opened 2024-04-03
added new GQI implementation pages and OFDM/OFDMA information

SkylineGBO:


user-guide/Standard_Apps/EPM/EPM_I-DOCSIS/I-DOCSIS_GQI_Integration.md
```diff
@@ -0,0 +1,60 @@
...
+| DS Port | - | X |
+| US Port | - | X |
+| Node Segment | - | X |
+| Service Group [Fiber Node] | - | X |
+| DS Service Group | - | X |
+| US Service Group | - | X |
+
+> [!NOTE]  
+> When you query data from the EPM connectors, the tables will show as duplicated. The first entry will be the source table and the second entry is the view table. The first entry is what you should use for queries.
+
```
miguel-slc, line 26: Hi, can we add an extra note/warning about querying multiple BEs. It could be the case that the query will go into timeout due to the amount of data queried. For these cases, a ad-hoc data sources could be a better approach.

SkylineGBO: if querying data from a single driver, like the BE driver, native GQI handles it well and will not timeout due to the lazy loading in place. When performing many operations like joins and combining data from various other drivers, that is when native GQI will begin timing out.

SkylineGBO: Added

MariekeGO (approved):

PR #2999, opened 2024-04-04
Update I-DOCSIS_components.md

SkylineGBO:


MariekeGO:
@SkylineGBO, should users always use the latest version of EPM I-DOCSIS? If not, we should add version info (i.e. clarify which component is added in which version) to make sure everything makes sense for those users who still use an older version.

SkylineGBO:
@MariekeGO it is recommended to always use the latest EPM version as we are always adding and improving features and very rarely removing features

MariekeGO (approved):

PR #3004, opened 2024-04-05
KI: CoreGateway issue

MariekeGO:


user-guide/Troubleshooting/Known_issues/KI_CoreGateway_fails_to_create_MessageBroker_session.md
```diff
@@ -0,0 +1,26 @@
...
+CoreGateway 2.14.6
+
+## Cause
+
+CoreGateway version 2.14.6 fails to create a MessageBroker session, causing it to be unable to receive requests from other modules.
+
+## Workaround
+
+1. Manually uninstall the CoreGateway module.
+1. Install version 2.14.5 of the module.
```
MariekeGO, line 18: @baptistepa, how can users install the older version of the module? Can they do so by installing the cloud pack from Dojo?

baptistepa: we would need to provide them with an installer, but I would not post this publicly yet since we are still investigating the issue. We have reason to believe it might be limited to a few systems.

MariekeGO:
Issue turned out to be more isolated than expected, so no issue page is needed.

PR #3008, opened 2024-04-05
Update InnerWorkingsIntroduction.md

akusur-skyline:
RTDisplay(=true) is only required for interactive automation scripts


develop/devguide/Connector/InnerWorkingsIntroduction.md
```diff
@@ -19,7 +19,7 @@ The <xref:InnerWorkingsSLProtocol> process executes the logic defined in a DataM
...
+A protocol typically contains many parameters. Some parameters are used for internal protocol logic, while others hold values that are of interest to the user and that are therefore displayed. The <xref:InnerWorkingsSLElement> process keeps track of parameters that have to be shown to the user and creates alarms. The SLProtocol process will push all parameters with RTDisplay set to true to the SLElement process. RTDisplay needs to be set to true for parameters that are positioned on a page or that need to be available for interactive Automation scripts, Visio drawings, or other elements (running on another DMA).
```
PedroDebevere, line 22:
```suggestion
A protocol typically contains many parameters. Some parameters are used for internal protocol logic, while others hold values that are of interest to the user and that are therefore displayed. The <xref:InnerWorkingsSLElement> process keeps track of parameters that have to be shown to the user and creates alarms. The SLProtocol process will push all parameters with RTDisplay set to true to the SLElement process. RTDisplay needs to be set to true for parameters that are positioned on a page or that need to be available for Automation scripts, Visio drawings, or other elements (running on another DMA).
```

MariekeGO: @PedroDebevere, do you mean that this option is in fact required for all Automation scripts, instead of only for interactive Automation scripts? Because I believe the purpose of this PR was specifically to mention interactive Automation scripts only.

PedroDebevere: yes, in a regular Automation script, RTDisplay also needs to be set to true, otherwise e.g. element.GetParameter("paramName") fails.

PR #3012, opened 2024-04-08
Epm training package

tiagopinaskyline:


MariekeGO:
The content of this PR overlaps in part with the existing [Experience and Performance Management section](https://docs.dataminer.services/user-guide/Advanced_Modules/EPM/EPM.html) in the user guide. As duplicate information is something we really want to avoid, I believe it would be better to integrate the new content in that existing section, instead of adding it under "Reference". 
This means that you can for example leave out the information about "what is EPM" but instead complete the existing page [About DataMiner Experience and Performance Management](https://docs.dataminer.services/user-guide/Advanced_Modules/EPM/About_DMS_Experience_and_Performance_Management.html) with any missing info. You can also add other pages there with general info about EPM as necessary. Information that is specifically related to the training package, rather than to EPM in general, could go under a new dedicated "Getting started with the EPM Integration Training package" subsection. 
For now, I will revert this PR to a draft, because it is not yet ready to merge in its current state.

PR #3034, opened 2024-04-10
Update UIComponentsMatrixSettingCrosspoints.md

TomW-Skyline:


develop/devguide/Connector/UIComponentsMatrixSettingCrosspoints.md
```diff
@@ -16,7 +16,7 @@ Once the command or SNMP set is executed, the flag needs to be changed again so
 
 Summary:
 
-- QAction 1: Triggers on write of matrix, process the data to put it into a buffer parameter in a certain format. Set flag to true if it is not false. (false indicates a set is being executed).
+- QAction 1: Triggers on write of matrix, process the data to put it into a buffer parameter in a certain format. Set flag to true if it is false (false indicates a set is being executed).
```
MariekeGO, line 19: This could actually be improved a bit further, because the sentence isn't quite clear as it is, and there's a comma splice I'd like to get rid of. Is "process the data" an instruction to the reader as to what to do with QAction 1?

PedroDebevere: updated the text. The original sentence "Set flag to true if it is not false." was correct but very confusing. This explanation is describing logic as can be found in the Matrix Basic example protocol (however, there the busy flag is set to 1 to indicate busy state, which seems more logical). The updated text now just mentions busy or not busy state to make it less confusing.

MariekeGO: @PedroDebevere, I've made some changes to make the text easier to read. Could you check if everything is still OK for you?

PedroDebevere (approved):
Changes are OK.

MariekeGO (approved):

PR #3035, opened 2024-04-11
RN 39098

MariekeGO:
@skyline-rge, I've distilled this RN to only the most essential information, as usually we don't go into full detail about the logging. Do you think this is sufficient, or are there things you'd still like me to add?


skyline-rge (approved):
I think this is more than sufficient. I usually add as much functional details in the RN as possible to also help ourselves in the future. The boundary between what is internal and what is not can sometimes be a bit blurry.

PR #3038, opened 2024-04-11
Added TAG Solution documentation

sebastianulloa:


user-guide/Standard_Apps/TAG/Installing_TAG_App.md
```diff
@@ -0,0 +1,81 @@
+---
+uid: Installing_TAG_App
+---
+
+# Installing the DataMiner TAG Management app
+
+## Prerequisites
+
+- DataMiner version 10.4.2 or higher
```
EdithLansens, line 9: @sebastianulloa, what is the minimum DataMiner version required to use the TAG Management application?

user-guide/Standard_Apps/TAG/Installing_TAG_App.md
```diff
@@ -0,0 +1,81 @@
...
+
+1. Select the target DataMiner System and confirm the deployment. The package will be pushed to the DataMiner System.
+
+If your system is **not connected to the cloud**:
+
+1. Look up the [*TAG Management* package](https://catalog.dataminer.services/details/package/6076) in the DataMiner Catalog.
+
+1. Click the *Download* button.
+
+1. In DataMiner Cube, go to *Apps > System Center > Agents > Manage*, and install the package in the same manner as a [DataMiner upgrade](xref:Upgrading_a_DataMiner_Agent_in_System_Center).
```
EdithLansens, line 38: @sebastianulloa, could you double-check this step? I tried this out, and got the following error:

image

user-guide/images/Tag_Management.svg
EdithLansens, line 1: @sebastianulloa, is this icon OK? It will be displayed alongside the other monitoring & management solutions on this page: https://docs.dataminer.services/user-guide/Standard_Apps/Monitoring_Solutions.html

If you had another icon in mind, let me know. See: [Skylicons](https://skyline.be/skylicons/)

user-guide/Standard_Apps/TAG/Installing_TAG_App.md
sebastianulloa: Minimum version has been updated. It is 10.4.5

user-guide/Standard_Apps/TAG/Installing_TAG_App.md
sebastianulloa: Modified steps to install locally as this were copy pasted from PTP package installation and it does not work the same.

user-guide/images/Tag_Management.svg
sebastianulloa: I tried updating the SVG file but I realized I'm not an expert! Do you mind updating the image to this one: https://skyline.be/skylicons/duotone/Multiviewers_Duo_Light.png, it is called "Multiviewers"

user-guide/Standard_Apps/TAG/Installing_TAG_App.md
EdithLansens: Even with these new steps, I cannot install the package after I've downloaded it. I am cloud-connected, so perhaps that is the issue? Could you look into this?
image

JoeyVanhalstSLC:
I would suggest to removing the section about installing the package when not cloud connected. I believe the "download" functionality in the Catalog is only available for Skyline employees. Looking at another example (https://docs.dataminer.services/user-guide/Standard_Apps/Monitoring_Solutions/Asset_Management/Installing_Asset_Management_app.html), it also only explains deploying from the Catalog. I'm also not sure if we need to explain here how to create the MCM/MCS elements? I assume they will already be on the system and we can refer (as we do already) to the connector help of both connectors.

EdithLansens:
> I would suggest to removing the section about installing the package when not cloud connected. I believe the "download" functionality in the Catalog is only available for Skyline employees. Looking at another example (https://docs.dataminer.services/user-guide/Standard_Apps/Monitoring_Solutions/Asset_Management/Installing_Asset_Management_app.html), it also only explains deploying from the Catalog. I'm also not sure if we need to explain here how to create the MCM/MCS elements? I assume they will already be on the system and we can refer (as we do already) to the connector help of both connectors.

I have removed the instructions for downloading the package without a cloud connection and for adding new TAG elements from the documentation. Aligning with other monitoring & management solutions (e.g. the asset management app), we now specify a cloud connection as a prerequisite. Moreover, I agree that the previously included information about adding TAG elements may be redundant. The previous "Creating TAG elements" section has been condensed into a note under "Accessing the TAG management app":

> [!NOTE]
> This application enhances the functionality of existing TAG MCM and MCS elements on your DMA. If these elements have not been configured yet, [add the necessary elements](xref:Adding_elements) using the [appropriate protocols](https://docs.dataminer.services/connector/doc/TAG_Video_Systems_MBC-7000.html).

We now provide links to the general documentation on adding an element and to the connector documentation, which should be sufficient.

If, in the future, we observe that users could benefit from a more thorough explanation, we can always adjust this documentation, @sebastianulloa.

PR #3041, opened 2024-04-12
Package Manifest Schema Reference - Add SetAsActive for Visio's 

MichielOda:


MariekeGO:
@MichielOda, since when is the SetAsActive element supported in DataMiner? That seems to be only thing that's still missing in this PR.

MariekeGO (approved):

PR #3044, opened 2024-04-16
DIS 2.48 release

MichielOda:


spetebroodt:
Hi @MichielOda I've made a few corrections. Can you double-check and approve?

MichielOda:
> Hi @MichielOda I've made a few corrections. Can you double-check and approve?

Looks good. I've updated the change from 'SkylineAPI' to 'Skyline API' as well in DIS, so it matches up. (I'm used to seeing SkylineAPI due to the NuGet)

spetebroodt (approved):

PR #3051, opened 2024-04-17
IDP 1.5.0 CU1

MariekeGO:
@jaklke, can these RNs be published like this, or should I add a preview label? Or would you prefer to wait to merge this PR for now?


jaklke (approved):
Please publish them as preview, because one or more RNs will still be added in the next release.

PR #3053, opened 2024-04-17
Update SAML_using_Entra_ID.md

LuisFreitas-Skyline:
Small tips added
@MichielMN

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
@@ -96,7 +96,7 @@ To set up external authentication, you first need to create an enterprise applic
 
 1. Still on the *Single sign-on* page, look for the *SAML Certificates* section and copy the *App Federation Metadata Url*.
 
-   You will need this later, for the **ipMetadata**, when [configuring DataMiner.xml](#configuring-dataminerxml-to-use-external-authentication).
+   You will need this later, for the *ipMetadata* [configuration in DataMiner.xml](#configuring-dataminerxml-to-use-external-authentication).
 
 ### Creating a DataMiner metadata file
 
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
@@ -96,7 +96,7 @@ To set up external authentication, you first need to create an enterprise applic
 
 1. Still on the *Single sign-on* page, look for the *SAML Certificates* section and copy the *App Federation Metadata Url*.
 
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
```diff
@@ -145,7 +145,7 @@ To set up external authentication, you first need to create an enterprise applic
        </md:SPSSODescriptor>
      ```
 
-1. Replace [ENTITYID] with the IP address or the DNS name of your DataMiner System. This must be the same as the *Entity ID* you specified while setting up the Microsoft Entra ID Enterprise application.
+1. Replace [ENTITYID] with the IP address or the DNS name of your DataMiner System, for example ``https://dataminer.example.com/``. This must be the same as the *Entity ID* you specified while setting up the Microsoft Entra ID Enterprise application.
```
MichielMN, line 148: The example url is already on the line below so I don't know if this would make it any clearer

MariekeGO: In fact, I think mentioning the same example URL as below here introduces confusion instead of resolving it: first we say to replace something with "https://dataminer.example.com/", then we ask to replace "https://dataminer.example.com/" with something else. Perhaps we could use a different example DNS name as an example here to prevent introducing this confusion?

LuisFreitas-Skyline: After rereading this, I came up with two suggestions:

1st => Leave it as it is, because in 5.1 of [Setting up a Microsoft Entra ID Enterprise application](https://docs.dataminer.services/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.html?tabs=tabid-1%2Ctabid-3#setting-up-a-microsoft-entra-id-enterprise-application) we say what the Entity ID must be and give an example.

2nd => Replace [ENTITYID] with the IP address or the DNS name of your DataMiner System. The specified URL(s) must match the Entity ID you specified while setting up the Microsoft Entra ID Enterprise application.

MariekeGO: Can the entity ID be more than one URL then, since you write "The specified URL(s)"?

MichielMN: No, it just needs to be one single string

MariekeGO: In that case, how about this? 
```suggestion
1. Replace [ENTITYID] with the IP address or the DNS name of your DataMiner System. This URL must be the same as the *Entity ID* you specified while setting up the Microsoft Entra ID Enterprise application.
```

MariekeGO (approved):

PR #3060, opened 2024-04-18
hostname not supported in DMS.xml

MariekeGO:
Based on https://community.dataminer.services/question/is-the-use-of-hostnames-supported/


SkylineCookieMonster:
@OnyxBit @SLC-JVH Can you give a concussion if hostnames are/should be supported in DataMiner?
Related question: https://community.dataminer.services/question/is-the-use-of-hostnames-supported/

SkylineCookieMonster:
@OnyxBit @SLC-JVH any update on the topic if hostnames are/should be supported in DataMiner?
Related question: https://community.dataminer.services/question/is-the-use-of-hostnames-supported/

PR #3072, opened 2024-04-23
Lengthy Correlation Rules page

kevin-degroote:


user-guide/Advanced_Modules/Correlation/Lengthy_Correlation_Rules.md
MariekeGO, line 1: Could you include the metadata and results sections for this BPA test, similar to the other BPA tests listed in the documentation (e.g. [Validate connectors](https://docs.dataminer.services/user-guide/Advanced_Functionality/DataMiner_Systems/BPA_tests/BPA_Validate_Connectors.html))?

In addition, you will also need to indicate how users can get this BPA test: is it only available on demand, or is it available by default from a specific DataMiner version onwards?

user-guide/Advanced_Modules/Correlation/Best_Practices_When_Creating_Correlation_Rules.md
```diff
@@ -0,0 +1,26 @@
...
+
+- When lengthy Correlation Rules are running, it can lead to the occurrence of run-time errors (RTEs).
+
+### Possible improvements
+
+- When a Correlation rule executes an Automation script that contains sleep actions, check if it is possible to reduce or even remove these. Whether this is possible depends entirely on the context and the purpose of those sleep actions.
+
+- When you define a Correlation rule to run a script, you can disable the option *Wait for the script to finish before continuing*. If this option is disabled, the rule will not wait before proceeding with other actions. While this may enhance the speed at which actions are scheduled, it is important to note that not all Automation scripts or Correlation rules are suitable for this adjustment. Disabling this feature could potentially lead to other system issues depending on the setup and the intended purpose of the script. Exercise caution when making this decision. It may be worth tweaking the rule for better overall performance, or considering alternative approaches to achieve your objectives.
```
MariekeGO, line 23: Do we have any information we could link to about when it's not a good idea to use this option? Right now, we're still a bit vague about this.

wdm-slc: I would say that turning off the "*Wait for the script to finish before continuing*" option is typically safe.

When could this cause issues?
- If an action on a first alarm needs to have completed before the action on a second alarm can execute. Can't really come up with an example here though, so might just be a theoretical thing.
- If the script being invoked is a lengthy one & correlation is triggering the script too often because of bad configuration. Instead of hogging up the correlation thread, the script requests towards the automation engine would start piling up here.

General rules as I see them:
- Prefer short scripts that execute fast
- Prefer turning off the "*Wait for the script to finish before continuing*" option

MariekeGO: I've made some changes to the PR based on this feedback. Could you check if everything still looks OK for you?

wdm-slc (approved):

MariekeGO (approved):

PR #3075, opened 2024-04-23
feat: added TAG MCM limitations

sebastianulloa:
Added a new page that explains what are the limitations when the app is used with only TAG MCM.


EdithLansens:
@sebastianulloa, thank you for adding this information! Instead of creating a new page dedicated solely to considerations for a TAG MCM setup, I've incorporated this info into the existing documentation pages. Since we strongly recommend using TAG MCS alongside TAG MCM, we should avoid advertising TAG MCM on its own.

EdithLansens (approved):

PR #3091, opened 2024-04-25
Update SRM_1.2.36.md

JensVandewalle:
Since it's compatible with FR 10.3.2 it is compatible with MR 10.4.0

### DIFF
--- a/release-notes/SRM/SRM_1.2.21.md
+++ b/release-notes/SRM/SRM_1.2.21.md
@@ -5,7 +5,7 @@ uid: SRM_1.2.21
 # SRM 1.2.21
 
 > [!NOTE]
-> This version requires that **DataMiner 10.1.10-10897 or higher** is installed. It is not compatible with the DataMiner Main Release track.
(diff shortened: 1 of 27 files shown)


MariekeGO:
@peter-vanpoucke, is this change OK? I thought this version of SRM was only suitable for use with the feature release, but possibly it is indeed OK with the main release from 10.4 onwards?

slc-peter-vanpoucke:
Strictly speaking that's right. The minimum version can change between feature release track versions though, which we'll try to avoid for the main release track. So it's better to pick that one for DM main releases, to avoid surprises.

I'll check with the team if we keep it that way.

slc-peter-vanpoucke (changes requested):
Might be worth rephrasing indeed. The same note should also be on SRM_2.0.1.

MariekeGO:
> Might be worth rephrasing indeed. The same note should also be on SRM_2.0.1.

I've rephrased the note where relevant and added it for 2.0.1 as well. Does this look OK for you?

slc-peter-vanpoucke (approved):

MariekeGO (approved):

PR #3093, opened 2024-04-25
GQI - API reference - GQIDMS GetConnection

skyline-rge:
Feature introduced with [RN39489](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=39489).

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/API_Reference/GQIDMS.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/API_Reference/GQIDMS.md
@@ -25,7 +25,7 @@ The `GQIDMS` class contains the methods listed below. These can be used to reque
 
 ### IConnection GetConnection()
 
-Available from DataMiner 10.4.7/10.5.0 onwards.<!-- RN39489 -->
+Available from DataMiner 10.4.6/10.5.0 onwards.<!-- RN39489 -->
 
 Provides a connection interface from GQI to SLNet. This connection can be used to request data from the DMS or to subscribe on events.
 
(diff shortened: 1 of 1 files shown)


WaHaeck (approved):

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/API_Reference/GQIDMS.md
```diff
@@ -23,6 +23,22 @@ The `GQIDMS` class contains the methods listed below. These can be used to reque
...
+Available from DataMiner 10.4.6/10.5.0 onwards.<!-- RN39489 -->
+
+Provides a connection interface from GQI to SLNet. This connection can be used to request data from the DMS or to subscribe on events.
+
+#### Returns
+
+An [IConnection](xref:Skyline.DataMiner.Net.IConnection) object representing the connection between GQI and SLNet.
+
+> [!TIP]
+> Interacting with the IConnection API directly is complex, but it is compatible to use effortlessly with other DataMiner libraries.
```
MariekeGO, line 37: Could I rephrase this tip like this, to make it a little more clear what we're recommending for the user, and also to avoid the term "effortlessly" (this term is a bit risky to use in a user guide, as users who do experience some trouble are likely to react negatively to this)? Or is this not what you meant?

```suggestion
> While interacting with the IConnection API directly is complex, you can easily use it with other DataMiner libraries.
```

skyline-rge: Feel free to rephrase as much as you want.
I struggled with this tip a bit, since maybe I wanted to squeeze to much info into 1 sentence here:

1. Bad news: due to its complexity, the IConnection interface is not easy to use directly by GQI extensions
2. Justification: This is by design, because the IConnection interface is used to be compatible with other existing DataMiner libraries
3. Good news: Existing DataMiner libraries can now be used via the IConnection interface

Btw, I can see that "effortlessly" might invoke negative reactions, but "easily" can probably too? :)

MariekeGO: You're right about that - frankly, I was struggling with this one too 😊
Maybe we could rephrase the note so it focuses not so much on the complexity - which should soon enough be obvious if someone tries to use it - but instead on how the interface should ideally be used?

skyline-rge: That sounds good!

MariekeGO (approved):

PR #3094, opened 2024-04-25
Updating Remote Log Collection requirements

LouisDJ-SL:


user-guide/Troubleshooting/RemoteLogCollection.md
```diff
@@ -16,13 +16,13 @@ To use this feature, your system needs to meet the following requirements:
 
 - Your system must be connected to dataminer.services. For information on how to connect, see [Connecting your DataMiner System to dataminer.services](xref:Connecting_your_DataMiner_System_to_the_cloud).
 
-- Port **5100** must be open for traffic on the **internal network** for the remote log collection to work. For more information on this endpoint, see [Customizing the dataminer.services endpoint configuration](xref:Custom_cloud_endpoint_configuration).
+- Port **5100** is by default open for traffic on the **internal network** for the remote log collection to work. If this setting was customized, make sure that the port is still enabled. For more information on this endpoint, see [Customizing the dataminer.services endpoint configuration](xref:Custom_cloud_endpoint_configuration).
```
MariekeGO, line 19: Do you mean that by default it must be port 5100 that is open, but that if another port was configured, users need to make sure that that port is still open?

LouisDJ-SL: I mean that when the Cloud Gateway is installed, the port is by default "5100", but a user can change that to e.g. "8452". Also when no extra network settings are made, the port will be open(enabled). But a user can disable this port and this action mostly happens when the port number was also changed.

MariekeGO: Could we rephrase this like this?
```suggestion
- The port used for traffic on the **internal network** must be open for the remote log collection to work. By default, this is port **5100**, but this setting can be customized. For more information on this endpoint, see [Customizing the dataminer.services endpoint configuration](xref:Custom_cloud_endpoint_configuration).
```

MariekeGO: As this is part of a list of requirements, we'll need to make sure it's also phrased as a requirement, so I'd suggest this:
```suggestion
- The port used for traffic on the **internal network** must be open for the remote log collection to work. The default dataminer.services endpoint configuration assigns **port 5100** for internal network traffic. Make sure this port is not closed because of firewall settings, or otherwise make sure a custom port is configured that is enabled for traffic. See [Customizing the dataminer.services endpoint configuration](xref:Custom_cloud_endpoint_configuration).
```

MariekeGO: So actually 5100 always needs to be open, regardless of whether another port is configured as the endpoint, but in the latter case that other port must be open as well?
```suggestion
- Port **5100** must be open for traffic on the **internal network** for the remote log collection to work. If a custom port is configured for the dataminer.services endpoint, this must be open as well. For more information on this endpoint, see [Customizing the dataminer.services endpoint configuration](xref:Custom_cloud_endpoint_configuration).
```

LouisDJ-SL: Yes, I think that now it says everything it should say. I'll commit this.

MariekeGO (approved):

PR #3114, opened 2024-04-30
RN 39370

EdithLansens:


user-guide/Advanced_Functionality/Databases/Optional_database_configurations/Offload_database/How_DataMiner_offloads_data.md
LaurensMT, line 1: The remark about the deprecated MySql.Data.dll is not applicable here. The MySql.Data.dll is present at 3 locations:
-C:\Skyline DataMiner\Files and C:\Skyline DataMiner Files\x64: the DLL located here is used by the core DataMiner processes and is used for example when a MySql database is used as local database, or like in this case is used to be able to offload data to an external MySql database. These DLLs recently got updated to the latest version (see RN39152) and is not considered deprecated.
-C:\Skyline DataMiner\ProtocolScripts: This DLL was used by protocols and automation scripts and is considered as deprecated. This is the one that is checked by the BPA.

LaurensMT (changes requested):
Added comment that the How_DataMiner_offloads_data.md does not need to be changed

MariekeGO (approved):

PR #3135, opened 2024-05-07
High level overview of what GQI is about

SebastiaanSL:


MariekeGO:
@SebastiaanSL, I've reviewed your PR and made some changes:

-  To make the "about" page more accessible and easy to understand, I've rephrased some things with easier language. (The original language was more suitable for marketing, but less for a user guide.)
- I've also added the DataMiner version where GQI was introduced again, as this is important information that should be easy to find for the user.
- I've removed the top-level page with only links, because we want to get rid of all these pages in the long run (they're high-maintenance without much added value).

Could you check if everything is still OK for you?

SebastiaanSL (commented):
Looks good to me!

At what point do version numbers get removed from the documentation? We're currently going towards 10.4.7 and we're still referencing version numbers from almost 4 years ago which is (way) beyond our supported version.

MariekeGO:
> Looks good to me!
> 
> At what point do version numbers get removed from the documentation? We're currently going towards 10.4.7 and we're still referencing version numbers from almost 4 years ago which is (way) beyond our supported version.

Good point, 10.0.x is no longer supported since February, so this no longer needs to be included. I'll adjust it.

MariekeGO (approved):

PR #3151, opened 2024-05-10
Create KI_Hostnames_not_supported_with_SNMP_traps.md

cheeyangK:
Hi,
I was working on a task recently and found out about this issue with SNMP traps not working if hostnames instead of IP were used in an element's polling IP. I thought it would be helpful to include this in the documentation. If there are any questions, feel free to reach out.

Below is the link to the related task:
https://collaboration.dataminer.services/task/238036


MariekeGO:
As this is functionality that is not supported rather than an issue that will be fixed, we will not add this as a known issue. However, it's important that this is instead clearly documented as not supported wherever necessary (e.g. in the connector help).

PR #3176, opened 2024-05-16
Removing mention of clustering with hostnames

XanderWA:


bertbsky:
OK for me. Not sure if the note below it should also be removed, or if we want to keep historical version information in the documentation. In the past adding an agent by hostname was _tolerated_ but with stricter logic in RN32951 that is no longer supported.

MariekeGO:
> OK for me. Not sure if the note below it should also be removed, or if we want to keep historical version information in the documentation. In the past adding an agent by hostname was _tolerated_ but with stricter logic in RN32951 that is no longer supported.

I don't think there's any value in keeping the note, as adding DMAs by hostname is not something we want to promote even in older versions, so it will only cause confusion. As such, I think it would be best if we removed it.

XanderWA:
I agree on removing the note in that case.

MariekeGO (approved):

PR #3192, opened 2024-05-23
Updated training modules and courses

EdithLansens:
@EmmaSaldi, I have updated the training module names and created a table listing all training modules and their courses. This replaces the outdated and somewhat confusing [old list](https://docs.dataminer.services/dataminer-overview/DevOps/DevOps_Enablement/Playbook/Learning_Resources.html). Can you confirm this is OK to publish?


EmmaSaldi (commented):
Hi Edith, I wasn't aware of the existence of this page. Thank you for updating the name of the training courses. I'm not sure if it's necessary to have a table of available training courses as it's all available on Dojo. Wouldn't it be better to add a link to the learning page on Dojo? Otherwise we will need to keep in mind to also update this page when content is added. 

The certification process mentioned here is also not correct. When participant want to start their certification process they need to contact their sales representative or submit the form via the certification page on Dojo. Once the ordering process is finalized, Skyline Academy is in charge of the practicalities such as sending the info on how to start their process and how to make an appointment with their examiner. (see FAQ on Dojo under certifications: 
image

EdithLansens:
@EmmaSaldi, thanks for the input! I agree that it is best to avoid duplicate information. I've replaced the table with a link to the courses page on Dojo.

PR #3210, opened 2024-05-27
Crud Actions on the Ping Monitoring app explained

joncorral:


MariekeGO:
@joncorral, something seems to be wrong with this PR: it contains a lot of changes that do not seem relevant to the explanation of the CRUD actions for the Ping Monitoring app, including changes to the GitHub workflow. Could you check what went wrong and make sure only the changes you need are included? If it's not possible to fix the issue within this PR, you may need to close it and create a new PR containing only the changes you need.

PR #3217, opened 2024-05-28
update docs for DIS

PedroDebevere:


develop/TOOLS/DIS/DIS_menu.md
```diff
@@ -30,7 +30,6 @@ The following sections provide more information on the DIS menu:
 - [Tool Windows \> DIS Validator](xref:DisValidatorToolWindow)
 - [Tool Windows \> DIS Comparer](xref:DisComparerToolWindow)
 - [Settings](xref:DIS_settings)
-- [Check for updates...](#check-for-updates)
```
MichielOda, line 33: As the 'Generate Class Library code' stuff was also kept (with the message in which version it was removed), maybe keep it as well? Or let's remove the class library stuff as well?

MichielOda:
Closed in favor of #3264 due to absence of Pedro.

PR #3222, opened 2024-05-29
Update docs nats TLS configuration

mykebueno:
Updated some information regarding the configuration of TLS for inter-node and DataMiner-node communication.

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Security_NATS.md
+++ b/user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Security_NATS.md
@@ -11,63 +11,69 @@ By default, NATS does **not** employ TLS encryption, leaving communication susce
 > [!NOTE]
 > This applies solely to instances involving a DMS cluster or external DxMs. If you have a single-Agent setup, all communication is confined to the local host, as there is only one NATS node.
 
-1. Request or generate TLS certificates (in PEM format).
+### Request or generate TLS certificates (PEM format)
 
    To generate self-signed certificates, we recommend that you **use our [scripts for generating TLS certificates](https://github.com/SkylineCommunications/generate-tls-certificates)**, available on GitHub. There is a version of the script for Linux and for Windows machines. The script requires two tools: *openssl* and the *Java keytool*. Both of these can run on Linux and Windows.
 
-1. Follow the steps below **for each of the NATS nodes** in the cluster to enable TLS encryption:
+### Configure TLS certificates
 
-   1. Update the **nats-server.config** file to incorporate the TLS section within the cluster segment.
+A **NATS node** is always configured on the system where the DataMiner software is hosted. Follow the steps below for **each of the NATS nodes** in the cluster to enable inter-node TLS encryption:
 
-      This adjustment is necessary to accommodate the specified location of the PEM files. For example:
+1. Ensure the **NATSForceManualConfig** option is enabled. This is crucial as DataMiner does not support automatic TLS configuration on the NATS nodes and will overwrite any additional configurations in the **nats-server.config** file. Refer to [Disabling automatic NATS configuration](xref:SLNetClientTest_disabling_automatic_nats_config) for detailed instructions.
 
-      ```
-      cluster {
-        tls {
-          cert_file: "absolute\\path\\to\\certificate_file\\certificate.pem"
-          key_file:  "absolute\\path\\to\\private-key_file\\private-key.pem"
-          ca_file:   "absolute\\path\\to\\rootCA_file\\rootCA.pem"
-        }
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Security_NATS.md
```diff
@@ -8,100 +8,128 @@ By default, NATS does **not** employ TLS encryption, leaving communication susce
...
 
-   To generate self-signed certificates, we recommend that you **use our [scripts for generating TLS certificates](https://github.com/SkylineCommunications/generate-tls-certificates)**, available on GitHub. There is a version of the script for Linux and for Windows machines. The script requires two tools: *openssl* and the *Java keytool*. Both of these can run on Linux and Windows.
+### Configure the TLS certificates
 
-1. Follow the steps below **for each of the NATS nodes** in the cluster to enable TLS encryption:
+A **NATS node** is always configured on the system where the DataMiner software is hosted. Follow the steps below for **each of the NATS nodes** in the cluster to enable inter-node TLS encryption:
 
-   1. Update the **nats-server.config** file to incorporate the TLS section within the cluster segment.
+1. Ensure the **NATSForceManualConfig** option is enabled (see [Disabling automatic NATS configuration](xref:SLNetClientTest_disabling_automatic_nats_config)).
```
MariekeGO, line 24: @SeppeDejonckheere, could you confirm that this step can be added (also on line 107 below)? Based on the caution under [Disabling automatic NATS configuration](https://docs.dataminer.services/user-guide/Reference/DataMiner_Tools/SLNetClientTest_tool/SLNetClientTest_tool_advanced_procedures/SLNetClientTest_disabling_automatic_nats_config.html), that means users will need to manually maintain several files and ensure the synchronization of credentials in the system. If the step is indeed OK to add, I think it might be best if we also added a caution block at the top of the current page to indicate that this procedure has those consequences.

LaurensVergote: Since the TLS aspect of the configuration isn't done automatically by DataMiner at this point and DataMiner has write control over these config files, it will overwrite them back to the original settings (removing the custom TLS configuration).
At this point in time, it is indeed a choice between automatic configuration and the security aspect of configuring TLS.

user-guide/Advanced_Functionality/Security/Advanced_security_configuration/Security_NATS.md
```diff
@@ -8,100 +8,128 @@ By default, NATS does **not** employ TLS encryption, leaving communication susce
...
 
 ## Enabling DataMiner-to-NATS node TLS communication
 
+To enable DataMiner-to-NATS node TLS communication, first [install the TLS certificate](#install-the-tls-certificate), and then [configure the TLS certificate](#configure-the-tls-certificate).
+
 > [!IMPORTANT]
 > TLS encryption for communication between software and NATS nodes is available starting from DataMiner 10.4.3 [CU0]/10.5.0<!-- RN 38302 --> with support for CA-signed certificates (Certificate Authority-Signed Certificates). While support for self-signed certificates will be introduced later, keep in mind that for now only CA-signed certificates are supported and attempting to use self-signed certificates will result in DataMiner startup failures.
 
-1. Install the certificate in the Windows Trusted Root Certification Authorities Certificate Store, which **requires administrator privileges**.
+### Install the TLS certificate
```
LaurensVergote, line 80: Installing the certificate(s) to the Windows Certificate Store is actually not needed for NATS at all. NATS will not use this Store to retrieve certificate information from.

MariekeGO: Do you mean we should leave out this entire  "Install the TLS certificate" section?

mykebueno: Yapp, that part is totally not needed

SeppeDejonckheere: Doesn't the messagebroker rely on the OS (and thus the certstore) to verify certificates, or does it use a custom certificateVerificationCallback?

mykebueno: Messagebroker doesn't do the validation at all for now. The plan is to have it when the TLS framework is implemented and the certificates are correctly configured. @LaurensVergote can confirm?

SeppeDejonckheere: Right, I wasn't sure if that was still the case

SeppeDejonckheere (approved):

MariekeGO (approved):

PR #3271, opened 2024-06-07
Update Protocol.Triggers.Trigger.Time-value.md

LaurensMT:
Added the remark that the value attribute in combination with a timeout is broken

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Triggers.Trigger.Time-value.md
+++ b/develop/schemadoc/Protocol/Protocol.Triggers.Trigger.Time-value.md
@@ -27,3 +27,7 @@ string
    </Content>
 </Trigger>
 ```
+
+## Remarks
+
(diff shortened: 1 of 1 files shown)


develop/schemadoc/Protocol/Protocol.Triggers.Trigger.Time-value.md
```diff
@@ -27,3 +27,7 @@ string
    </Content>
 </Trigger>
 ```
+
+## Remarks
+
+Above example with the combination of timeout with the value attribute is broken since RN8550, DataMiner version 8.5.2.6. At this moment, the trigger on timeout will always go off when the polling of a parameter results in a timeout, regardless of what is configured in the value attribute.
```
MariekeGO, line 33:
```suggestion
The combination of a timeout with the value attribute shown in the example above no longer works since DataMiner 8.5.2.6 (RN 8550). At present, the trigger on timeout will always go off when the polling of a parameter results in a timeout, regardless of what is configured in the value attribute.
```

MariekeGO (approved):

PR #3287, opened 2024-06-11
State component index limit

MariekeGO:
Based on https://community.dataminer.services/question/dashboard-index-limit/


MariekeGO:
@GillesBara, should I also add this info for the progress bar, gauge, and ring component?

GillesBara (approved):

GillesBara:
> @GillesBara, should I also add this info for the progress bar, gauge, and ring component?

It's indeed the same behavior for these visualizations, when combined with parameter columns.

PR #3291, opened 2024-06-12
Introducing GitOps

saatsky:
Describe GitOps in the context of cloud offering and DevOps practices as CI/CD pipeline as previously referred.


MariekeGO:
@SammyDW, could you verify if this change to the DataMiner overview is OK?

MariekeGO:
We will not add this change yet and close this PR, as this first has to be verified with internal stakeholders, and if it is approved, it will need to be adjusted in more places than just here. When the change gets approved, you can create a new pull request to merge this.

PR #3312, opened 2024-06-17
Update Ad_hoc_Tutorials_GQIDMS.md

EdibSU:


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Tutorials_GQIDMS.md
```diff
@@ -136,3 +136,10 @@ Transform the responses into a [GQIPage](xref:GQI_GQIPage). Each `LoginInfoRespo
 
 > [!NOTE]
 > Ensure `DateTime` values are in UTC. Although the `ConnectTime` value is in UTC, it is not explicitly marked as such. Therefore, this is specified in the example.
+
+> [!NOTE]
+> Don't forget to compile script as [library](xref:Compiling_a_CSharp_code_block_as_a_library) by including following in the Script.Exe tag.
```
MariekeGO, line 141: If this is something the user definitely always has to do, it would be better to add it as another step in the procedure instead of as a note.

EdibSU: It is something user always has to do when working with GQI, it's prerequisite for creating GQI DS. Depending on persons experience it is either just a reminder or additional step to be performed. Your call, if you prefer to have it as extra step I can do that, if you feel like this is fine we can leave it as is.

MariekeGO: As this tutorial is aimed at learners who may not have much experience with GQI, I think adding it as a step would be better than as a reminder.

MariekeGO:
@WaHaeck, could you confirm if this is OK to merge?

WaHaeck:
@MariekeGO It is indeed correct, but only applicable for anyone working with DIS. In this tutorial, DIS was never actually mentioned. The procedure of compiling the script as library is something that should be done for each custom operator and ad hoc data source, and is mentioned here: https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html
Because of that I did not mention it in this tutorial specifically.

MariekeGO:
> @MariekeGO It is indeed correct, but only applicable for anyone working with DIS. In this tutorial, DIS was never actually mentioned. The procedure of compiling the script as library is something that should be done for each custom operator and ad hoc data source, and is mentioned here: https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html Because of that I did not mention it in this tutorial specifically.

I'm not sure I follow completely: Do you mean that if DIS is not used, this extra compiling step is not needed? Or that it's different in that case? I do think that if the tutorial will not work without this additional step, we should make sure that this is clear, so that the tutorial remains user-friendly for inexperienced users as well.

WaHaeck:
> > @MariekeGO It is indeed correct, but only applicable for anyone working with DIS. In this tutorial, DIS was never actually mentioned. The procedure of compiling the script as library is something that should be done for each custom operator and ad hoc data source, and is mentioned here: https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html Because of that I did not mention it in this tutorial specifically.
> 
> I'm not sure I follow completely: Do you mean that if DIS is not used, this extra compiling step is not needed? Or that it's different in that case? I do think that if the tutorial will not work without this additional step, we should make sure that this is clear, so that the tutorial remains user-friendly for inexperienced users as well.

Indeed, if DIS is not used the flow is a bit different: the user would need to 'check' the compile as library option in Cube. We can indeed add it to the tutorial to make it clear.

MariekeGO:
> > > @MariekeGO It is indeed correct, but only applicable for anyone working with DIS. In this tutorial, DIS was never actually mentioned. The procedure of compiling the script as library is something that should be done for each custom operator and ad hoc data source, and is mentioned here: https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html Because of that I did not mention it in this tutorial specifically.
> > 
> > 
> > I'm not sure I follow completely: Do you mean that if DIS is not used, this extra compiling step is not needed? Or that it's different in that case? I do think that if the tutorial will not work without this additional step, we should make sure that this is clear, so that the tutorial remains user-friendly for inexperienced users as well.
> 
> Indeed, if DIS is not used the flow is a bit different: the user would need to 'check' the compile as library option in Cube. We can indeed add it to the tutorial to make it clear.

I've now added this info. Do the changes look OK for you?

WaHaeck (approved):

WaHaeck:
> > > > @MariekeGO It is indeed correct, but only applicable for anyone working with DIS. In this tutorial, DIS was never actually mentioned. The procedure of compiling the script as library is something that should be done for each custom operator and ad hoc data source, and is mentioned here: https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Configuring_an_ad_hoc_data_source_in_a_query.html Because of that I did not mention it in this tutorial specifically.
> > > 
> > > 
> > > I'm not sure I follow completely: Do you mean that if DIS is not used, this extra compiling step is not needed? Or that it's different in that case? I do think that if the tutorial will not work without this additional step, we should make sure that this is clear, so that the tutorial remains user-friendly for inexperienced users as well.
> > 
> > 
> > Indeed, if DIS is not used the flow is a bit different: the user would need to 'check' the compile as library option in Cube. We can indeed add it to the tutorial to make it clear.
> 
> I've now added this info. Do the changes look OK for you?

Looks good, thanks!

MariekeGO (approved):

PR #3320, opened 2024-06-19
Cloud - Added change log for release 06/19

brayunm:


release-notes/Cloud/DCP_change_log.md
MariekeGO, line 1: @brayunm, for the two fixes, could you rephrase the title and description so they focus on the issue rather than the solution? For fixes, it's important that a user searching for the issue can easily spot it, so we need to make sure each fix contains a clear, user-friendly description of the issue, in such a way that the user will recognize it (e.g. in case of the issue with the cookies, the user may not know that the cookies were the problem, so they may not recognize the issue from the current description).

brayunm: Title & descriptions have been updated

MariekeGO (approved):

PR #3321, opened 2024-06-19
DataMiner Object Models (DOM ) - Added reference to the DOM Viewer Application

ruicamacho:


MariekeGO (approved):

MariekeGO:
@ruicamacho, I made some adjustments because a more down-to-earth, minimalist style should be used when adding things in the user guide section. People looking in the user guide want to get their information as quickly and efficiently as possible, so the language we use needs to adjusted accordingly.

PR #3323, opened 2024-06-19
Troubleshooting - Cassandra

spetebroodt:


user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
```diff
@@ -0,0 +1,157 @@
+---
+uid: Troubleshooting_Cassandra
+---
+
+# Troubleshooting – Cassandra
+
+## Context
```
MariekeGO, line 7: Is this "Context" title necessary? Usually, when we add context at the top of a page, we do so directly under the main title, so keeping this title will clash a bit with the rest of the documentation.

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
```diff
@@ -0,0 +1,157 @@
...
+
+Two types of Cassandra storage are used, i.e. two distinct types of Cassandra databases:
+
+- Dedicated clustered storage:, called *Cassandra Cluster* below.
+- Storage per DMA (with or without indexing), called *Cassandra Single* below.
+
+> [!TIP]
+> For more information on available storage options, see [About databases](xref:Databases_about).
+
+When you start troubleshooting, you first need to determine whether the setup uses *Cassandra Cluster* or *Cassandra Single*. There are some key differences between the two that are important for your investigation.
```
MariekeGO, line 19: Can we add a title above this line, e.g. "Determining the type of setup"?

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
```diff
@@ -0,0 +1,157 @@
...
+
+- Elements fail to start and are in status "Error".
+
+- In the Alarm Console, the following alarms are shown:
+
+  - Error alarm with value "General database failure, failed to initialize..."
+  - Error alarm mentioning "DataMiner goes into offload mode"
+
+    > [!NOTE]
+    > This error may also be caused by other databases (e.g. Elasticsearch).
```
MariekeGO, line 48: Maybe we should change this to OpenSearch, since Elasticsearch is no longer recommended?

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
```diff
@@ -0,0 +1,157 @@
...
+
+- The SLDataGateway process is leaking memory.
+
+## Prerequisites and required tools
```
MariekeGO, line 52: To avoid possible confusion, can we change this title to "Prerequisites and required tools for troubleshooting"?

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
```diff
@@ -0,0 +1,157 @@
...
+
+- *Microsoft Platform* elements
```
MariekeGO, line 54: Perhaps we should detail this a bit more? For example:
```suggestion
- A *Microsoft Platform* element for each DMA server
```

MariekeGO (commented):
To increase consistency and make it easier to get an overview of all procedures, I've made some changes to the structure of the troubleshooting procedures section. I've among others changed some titles to be in line with the new page, and moved the other Cassandra procedure so it's a node below the new page:
image

 What do you think of this approach?

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
MiekeDryepondt: It's OK to remove the Context sub title.

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
MiekeDryepondt: Can we both since we still have setups using Elasticsearch. And issues are more likely to occur with something old / no longer recommended. For example:
e.g. OpenSearch, Elasticsearch. Note that the latter is no longer recommended.

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
MiekeDryepondt: Yes please.

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
MiekeDryepondt: OK. good idea. Also add then Trending should be enabled. I couldn't find an overview of what is recommended or min. required. If that is the case we will make an extra page on recommended trending of the microsoft platform for troubleshooting.

user-guide/Troubleshooting/Procedures/Troubleshooting_Cassandra.md
MiekeDryepondt: I'm not sure, the goal of the titles was to keep them the same across all troubleshooting pages and I don't know if we will always have a "Determining the type of setup". 
Perhaps we can add it and if the troubleshooting item is generic across setups, this title section would simply highlight (avoid confusion) that it doesn't matter what type of setup you have.
So: yes, add the title and we can evaluate along the way.

MiekeDryepondt (commented):
Replied to the comments

MariekeGO:
> Replied to the comments

@MiekeDryepondt, I've made the necessary changes based on your feedback. Could you confirm if this is OK to merge now?

MiekeDryepondt:
Perfect, thank you.

MariekeGO (approved):

PR #3324, opened 2024-06-19
Update Overview_of_Soft_Launch_Options.md

ryan-hepworth:


user-guide/Reference/Soft-launch_options/Overview_of_Soft_Launch_Options.md
```diff
@@ -242,7 +242,7 @@ Enables or disables the legacy Reports and Dashboards modules.
 - **Minimum version**: 10.1.10/10.2.0
 
 > [!NOTE]
-> The legacy Reports and Dashboards modules are disabled by default as from DataMiner versions 10.4.0/10.4.1 ([RN 37786](xref:General_Feature_Release_10.4.1#legacy-reports-dashboards-and-annotations-modules-are-now-end-of-life-and-will-be-disabled-by-default-id_37786)). If you want to keep on using these legacy modules, set this soft-launch option to *true*, then run `C:\Skyline DataMiner\Tools\ConfigureIIS.bat` as Administrator, and restart the DataMiner Agent.
+> The legacy Reports and Dashboards modules are disabled by default as from DataMiner versions 10.4.0/10.4.1 ([RN 37786](xref:General_Feature_Release_10.4.1#legacy-reports-dashboards-and-annotations-modules-are-now-end-of-life-and-will-be-disabled-by-default-id_37786)). If you want to keep on using these legacy modules, set this soft-launch option to *true* by adding <LegacyReportsAndDashboards>true</LegacyReportsAndDashboards> to SLNet Tags in C:\Skyline DataMiner\SoftLaunchOptions.xml or create a new file, then run `C:\Skyline DataMiner\Tools\ConfigureIIS.bat` as Administrator, and restart the DataMiner Agent.
```
MariekeGO, line 245: Because the added information fully overlaps with the instructions detailed under [Activating soft-launch options](https://docs.dataminer.services/user-guide/Reference/Soft-launch_options/Activating_Soft_Launch_Options.html), it would be better if we instead added a link to that page. That way we avoid duplicate information (which tends to be overlooked when there are updates, introducing a risk of conflicting information). In addition, as this link would also be relevant for the LegacyAnnotations option above, I'd also add it there.

ryan-hepworth: Ok that makes sense

PR #3334, opened 2024-06-21
RN 30077

MariekeGO:


slc-peter-vanpoucke (approved):
Looks good. 
That change has been there for a while already, but won't hurt to mention the previous value.

PR #3360, opened 2024-06-26
RN 39691+39692

MariekeGO:


VeerleLedoux (changes requested):
I think it would be good to also mention in the following section that parameters which do not have active behavioral anomaly detection will not be included in the trend graph insights results generated by relation learning: "- *Run on trended parameters by default*: Available from DataMiner 10.4.8/10.5.0 onwards. If this option is enabled (default setting), behavioral anomaly detection is enabled by default for trended numeric parameters that are not part of partial tables, but you can override this setting for specific parameters in a trend template. If this option is disabled, behavioral anomaly detection is disabled system-wide, except for trended parameters of which the *Anomalies* setting has explicitly been enabled in the trend template. See [Configuring trend templates](xref:Configuring_trend_templates)."

tobedeprez:
No further comments apart from Veerle's one.

MariekeGO:
> I think it would be good to also mention in the following section that parameters which do not have active behavioral anomaly detection will not be included in the trend graph insights results generated by relation learning: "- _Run on trended parameters by default_: Available from DataMiner 10.4.8/10.5.0 onwards. If this option is enabled (default setting), behavioral anomaly detection is enabled by default for trended numeric parameters that are not part of partial tables, but you can override this setting for specific parameters in a trend template. If this option is disabled, behavioral anomaly detection is disabled system-wide, except for trended parameters of which the _Anomalies_ setting has explicitly been enabled in the trend template. See Configuring trend templates."

I've now tried to add this in a note in such a way that it makes the previous existing sentence about the effect of disabling the feature for relation learning unnecessary. What do you think?

VeerleLedoux (approved):
change looks good, thx

PR #3362, opened 2024-06-26
Update UIComponentsButton.md

rebeccackelsall:
Addition of information on the setter attribute to warn developers not to use it for buttons following this dojo question.  https://community.dataminer.services/question/double-trigger-of-qaction-from-button/answer/127436/

### DIFF
--- a/develop/devguide/Connector/UIComponentsButton.md
+++ b/develop/devguide/Connector/UIComponentsButton.md
@@ -41,6 +41,7 @@ To define a button:
 
 > [!NOTE]
 > Typically, the description tag is left empty when a button is defined.
+> Please be aware that if you are using the button to trigger a QAction if you have setter="true" attribute you may cause your Qaction to get triggered twice. This attribute should not be present on this parameter type. 
 
 ![alt text](../../images/uibutton.png "DataMiner Cube button")
 
--- a/develop/devguide/Connector/UIComponentsButton.md
+++ b/develop/devguide/Connector/UIComponentsButton.md
@@ -40,8 +40,9 @@ To define a button:
 ```
 
 > [!NOTE]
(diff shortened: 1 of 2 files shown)


develop/devguide/Connector/UIComponentsButton.md
```diff
@@ -41,6 +41,7 @@ To define a button:
 
 > [!NOTE]
 > Typically, the description tag is left empty when a button is defined.
+> Please be aware that if you are using the button to trigger a QAction if you have setter="true" attribute you may cause your Qaction to get triggered twice. This attribute should not be present on this parameter type.
```
MariekeGO, line 44: To make sure this info gets spotted straight away, we'd propose to change the description in line 7 as follows instead of adding this info in a note:
`A button is a write parameter with the intent to invoke an action when it is activated. Setting another parameter is not its goal, so it should not have a matching read parameter using the setter attribute.`

FlorisCO: I believe the observed behaviour is still important to document, but that it would be more fitting in the page describing the setter attribute. (uid: Protocol.Params.Param-setter) (ddevelop/schemadoc/Protocol/Protocol.Params.Param-setter.md)
I suggest extending its description as such:
`If no corresponding read parameter is present, the write parameter may execute its triggers or actions a second time`.

PedroDebevere: This behavior only seems to occur if the button is part of a table. Therefore, the sentence "If no corresponding read parameter is present, the write parameter may execute its triggers or actions a second time." could be updated to "If the button is part of a table and no corresponding read parameter is present, the write parameter may execute its triggers or actions a second time." A validator check will be created to spot this during development.

MariekeGO (approved):

PR #3371, opened 2024-06-27
Web component tutorial - KATA 34

SebastiaanSL:


MariekeGO:
@SebastiaanSL, we've reviewed your PR and made some changes. Among others, the tutorial has been moved to the Low-Code Apps section, because it seems to fit better there given that the user is working with an LCA instead of a dashboard in the tutorial. Could you check if everything is still OK for you?

SebastiaanSL:
@MariekeGO, the exercise is carried out in a low-code app, but the component itself isn’t specific to apps and can also be accessed via a dashboard. I think all component tutorials, including those for web components, should be part of the dashboard tutorials since the dashboard serves as the foundation. There’s much more to discuss about the Low-Code Apps (LCA) framework. Therefore, it makes sense to place tutorials on fundamental capabilities that apply to both dashboards and LCA in the dashboards section, and keep LCA tutorials focused on features unique to LCA.

MariekeGO:
> @MariekeGO, the exercise is carried out in a low-code app, but the component itself isn’t specific to apps and can also be accessed via a dashboard. I think all component tutorials, including those for web components, should be part of the dashboard tutorials since the dashboard serves as the foundation. There’s much more to discuss about the Low-Code Apps (LCA) framework. Therefore, it makes sense to place tutorials on fundamental capabilities that apply to both dashboards and LCA in the dashboards section, and keep LCA tutorials focused on features unique to LCA.

OK, I've reverted that change and added a note to clarify why there can be LCA examples in the Dashboards tutorials.

SebastiaanSL (commented):
LGTM

MariekeGO (approved):

PR #3374, opened 2024-06-28
Units and Decimals are no longer a limitation with a new release of DataAPI

JoaoSE-Skyline:


user-guide/Advanced_Modules/Data_Sources/Data_Sources_Limitations.md
```diff
@@ -30,10 +30,10 @@ When working with the [Data API](xref:Data_API) and [scripted connectors](xref:S
 
 - Parameters in auto-generated connectors:
 
-  - Lack units, with no adjustable configuration.
+  - ~~Lack units, with no adjustable configuration.~~ No longer a limitation since [DataAPI 1.1.3](xref:DataAPI_change_log#1-april-2024---new-feature---dataapi-113---new-configuration-endpoint-for-units-and-decimals-id_39016)
```
MariekeGO, line 33: We don't usually use an indication like this in the user guide (it doesn't look nice and isn't future-proof). I'd recommend either leaving out this limitation entirely, or indicating that this is limited to older versions only, e.g. 
```suggestion
  - Only prior to [DataAPI 1.1.3](xref:DataAPI_change_log#1-april-2024---new-feature---dataapi-113---new-configuration-endpoint-for-units-and-decimals-id_39016), lack units, with no adjustable configuration.
```
However, since this is a feature in preview, leaving this limitation out entirely is probably the best approach.

MariekeGO (approved):

PR #3391, opened 2024-07-03
People & Organizations

EdithLansens:
@ruicamacho, could you review this documentation about the People & Organizations app? Thanks in advance!


MariekeGO:
For the "Assigning a skill and experience to a contact", "Assigning areas of expertise and roles", and "Assigning organization categories" sections, I'm wondering if it would be possible to separate these into separate concepts and procedures a bit more. This could make it more user-friendly for users looking for very specific info.
We could for instance make a section "Managing areas of expertise" and a similar one for roles, starting with a brief intro about each pane, followed by a separate procedure for each action on this pane, so also one for assigning an area of expertise or a role.

PR #3392, opened 2024-07-03
Update in DITT Overview

EstebanSkyline:
Elaborating on the main functionality of the DITT


MariekeGO:
@EstebanSkyline, I've simplified your description a little to keep it suitable for the user guide. The kind of elaborate writing you used is more suitable for marketing texts, and would go well in the "DataMiner Overview" section of the documentation, but it's not really suited for the User Guide, which is primarily meant to inform users quickly and efficiently. We need to avoid adding repetitions or unnecessarily elaborate language in there.
Could you take a look and check if anything still needs to be tweaked?

EstebanSkyline:
> @EstebanSkyline, I've simplified your description a little to keep it suitable for the user guide. The kind of elaborate writing you used is more suitable for marketing texts, and would go well in the "DataMiner Overview" section of the documentation, but it's not really suited for the User Guide, which is primarily meant to inform users quickly and efficiently. We need to avoid adding repetitions or unnecessarily elaborate language in there. Could you take a look and check if anything still needs to be tweaked?

rdpslc:
Changes reviewed and OK for release.

MariekeGO (approved):

PR #3393, opened 2024-07-04
Catalog API in user guide

MariekeGO:
@baptistepa & @SkylineKevinMOR, could you take a look at this PR to add the Catalog API to the user guide? This is a first draft, so changes or more extensive info may be needed.

### DIFF
--- a/user-guide/Cloud_Platform/Catalog/Catalog_API.md
+++ b/user-guide/Cloud_Platform/Catalog/Catalog_API.md
@@ -1,0 +1,77 @@
+---
+uid: Catalog_API
+---
+
+# Catalog API
+
+The Catalog API can be used to interact with the DataMiner Catalog, for example to [register a Catalog item](xref:Register_Catalog_Item#using-the-catalog-api).
+
+Below you can find an overview of the methods available in the Catalog API (V2).
+
+- GET api/user-catalog/v2-0/catalogs/search
+
+  Searches for Catalog items.
+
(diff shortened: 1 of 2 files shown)


SkylineKevinMOR (approved):
Looks good to me

baptistepa:
I do not think it is a good idea to document this... This will be yet another place where we need to document the API. ATm we have docs auto generated at https://catalogapi-prod.cca-prod.aks.westeurope.dataminer.services/swagger/index.html?urls.primaryName=User%20Catalog%20API%20v1.0. We could move this to a better URL if we want in the future.

baptistepa:
@AlexanderVK-SLC what do you think?

AlexanderVK-SLC:
The URL can be accessed by everyone, yes, and the swagger is the closest source of truth. However, I would prefer to move it behind api.dataminer.services/.../swagger url before adding it to the docs, because these URLs will not change unlike the direct cluster URL.

It's also a good motivator for us to keep our swagger well documented, knowing that it is consumed by external users directly without UI.

MariekeGO:
> The URL can be accessed by everyone, yes, and the swagger is the closest source of truth. However, I would prefer to move it behind api.dataminer.services/.../swagger url before adding it to the docs, because these URLs will not change unlike the direct cluster URL.
> 
> It's also a good motivator for us to keep our swagger well documented, knowing that it is consumed by external users directly without UI.

In that case, I'd simply close the current PR for now. Could you create a PR to add the link once it's available?

PR #3395, opened 2024-07-04
Update Starting_a_restored_DataMiner_Agent_in_a_cluster.md

AlexanderGorbunovSkyline:
Added a missing step of setting the cluster name on the new agent. See https://docs.dataminer.services/user-guide/Advanced_Functionality/DataMiner_Systems/Adding_a_DMA/Adding_a_regular_DataMiner_Agent.html


MariekeGO:
@AlexanderGorbunovSkyline To prevent duplicate info (which tends to go overlooked during updates), I've replaced the procedure to add the DMA with a link to the identical procedure to add a DMA to a cluster. Could you take a look and confirm if this is OK for you?

AlexanderGorbunovSkyline (commented):
I agree with your suggestion.

MariekeGO (approved):

PR #3402, opened 2024-07-05
Added user guide for DataMiner Objects Tool

VictorScherpereell:


user-guide/Standard_Apps/Monitoring_Solutions/DataMiner_Objects_Tool/Using_DataMiner_Objects_Tool.md
```diff
@@ -0,0 +1,60 @@
+---
+uid: Using_DataMiner_Objects_Tool
+---
+
+# Using the DataMiner Objects Tool
+
+## Low-Code App
```
MariekeGO, line 7: Because so much of the content is blurred in the images under the Low-Code App title here, they don't look very nice and they don't give a very good first impression of what the app looks like. Could you instead show an image of the full app UI, and use numbers to indicate the different areas you want to refer to in the text (the way it's for example done here: https://docs.dataminer.services/user-guide/Standard_Apps/Monitoring_Solutions/Huawei_Manager/Working_with_Huawei_Manager_Dashboard.html)?

MariekeGO (changes requested):

VictorScherpereell:
Hi Marieke, is something still expected from me regarding this pull request?

user-guide/Standard_Apps/Monitoring_Solutions/DataMiner_Objects_Tool/Using_DataMiner_Objects_Tool.md
MariekeGO, line 1: On the DataMiner_Objects_Tool page, you mention that it is possible to delete objects, but I don't see this option in the UI here. Is this indeed possible? If yes, could you add info about this as well?

user-guide/Standard_Apps/Monitoring_Solutions/DataMiner_Objects_Tool/Using_DataMiner_Objects_Tool.md
```diff
@@ -0,0 +1,74 @@
...
+
+- In the scrollable checkbox list at the bottom (4), you can **select individual items** or clear the selection to fine-tune the result according to your preferences.
+
+## Executing actions on selected objects
+
+After you have manually fine-tuned the results, you can execute the following actions for the selected objects:
+
+![Actions](~/user-guide/images/Using_DataMiner_Objects_Tool_Actions.png)
+
+- To add a new multi-line textbox to the dialog containing all selected objects serialized in JSON format, click **Show JSON** (1).
```
MariekeGO, line 72: What would this JSON format typically be used for? It might be useful if we added this here, so users searching for this will come across this specific page.

VictorScherpereell: The JSON is used to see which fields of the object have which value and is mainly used to investigate/debug issues. Can also be used to export the object and import it into any other application. Should I add these 2 main uses for the JSON?

VictorScherpereell: Looking good! Thanks!

MariekeGO (approved):

PR #3429, opened 2024-07-10
Expired SSTables blocking other tables from being dropped

MariekeGO:
Based on input from @joncorral

### DIFF
--- a/user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
+++ b/user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
@@ -4,7 +4,7 @@ uid: Cassandra_expired_blocking_SSTables
 
 # Expired SSTables blocking other tables from being dropped
 
-In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped. This could result in various issues, such as an increase in the disk usage of the Cassandra data partition, or memory leak behavior in the TWCS keyspaces (*trend_data_medium*, *trend_data_short*, and *trend_data_long*).
+In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped. This could result in various issues, such as an increase in the disk usage of the Cassandra data partition or a continuous increase of the size of the TWCS keyspaces on the disk (*trend_data_medium*, *trend_data_short*, and *trend_data_long*).
 
 You can check if you are indeed encountering this issue using the Cassandra tools, for example with the following commands (on Linux):
 
--- a/user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
+++ b/user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
@@ -4,11 +4,25 @@ uid: Cassandra_expired_blocking_SSTables
 
 # Expired SSTables blocking other tables from being dropped
 
-In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped.
+In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped. This could result in various issues, such as an increase in the disk usage of the Cassandra data partition, or memory leak behavior in the TWCS keyspaces (*trend_data_medium*, *trend_data_short*, and *trend_data_long*).
 
-To resolve this issue, you can enable the *unsafe_aggressive_sstable_expiration* option for the Cassandra database. This is something that should only be done to troubleshoot this issue, because it is considered a risky option otherwise. The data in the expired SSTables will be dropped without being checked, but as it consists entirely of tombstone data in this case, this is not a problem.
+You can check if you are indeed encountering this issue using the Cassandra tools, for example with the following commands (on Linux):
 
-To enable this option, follow the steps below.
+```txt
+yum install cassandra-tools
+
+sudo systemctl stop cassandra
+
+sstableexpiredblockers zm1_cc_trend_data_short trend_data_short
+
+for f in *Data.db; do meta=$(sudo sstablemetadata $f); echo -e "Max:" $(date --date=@$(echo "$meta" | grep Maximum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') "Min:" $(date --date=@$(echo "$meta" | grep Minimum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') $(echo "$meta" | grep droppable) ' \t ' $(ls -lh $f | awk '{print $5" "$6" "$7" "$8" "$9}'); done | sort
+```
+
+With the output of the sstableexpiredblockers, you can see which tables are blocking others from being dropped. If the percentage of tombstones in the SSTable is higher than 1, you can assume that all data is tombstone data.
+
+To resolve this issue, we recommend enabling the *unsafe_aggressive_sstable_expiration* option for the Cassandra database. This is something that should only be done to troubleshoot this issue, because it is considered a risky option otherwise. The data in the expired SSTables will be dropped without being checked, but as it consists entirely of tombstone data in this case, this is not a problem.
+
+To enable this option, follow the steps below. These steps are intended for a Linux system, since this is the recommended operating system for a Cassandra cluster.
 
 > [!NOTE]
 > Ideally, this should be done on one node at a time to reduce the downtime for the Cassandra Cluster. However, whether your cluster can cope with one node being down at a time depends on its configuration and architecture. To check this, you can use the [ecyrd.com Cassandra calculator](https://www.ecyrd.com/cassandracalculator/).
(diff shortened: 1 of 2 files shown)


user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,49 @@
...
+uid: Cassandra_expired_blocking_SSTables
+---
+
+# Expired SSTables blocking other tables from being dropped
+
+In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped.
+
+To resolve this issue, you can enable the *unsafe_aggressive_sstable_expiration* option for the Cassandra database. This is something that should only be done to troubleshoot this issue, because it is considered a risky option otherwise. The data in the expired SSTables will be dropped without being checked, but as it consists entirely of tombstone data in this case, this is not a problem.
+
+To enable this option, follow the steps below.
```
MariekeGO, line 11: For which kind of setup is this procedure intended? For Cassandra clusters hosted on Linux?

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,49 @@
+---
+uid: Cassandra_expired_blocking_SSTables
+---
+
+# Expired SSTables blocking other tables from being dropped
+
+In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped.
```
MariekeGO, line 7: Are there any symptoms of this issue that users would recognize, so that they know that this is the issue they are facing?

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,49 @@
...
+   ```
+
+   In most cases, you will need to run this query for the following tables:
+
+   - PREFIXPLACEHOLDER_trend_data_long
+   - PREFIXPLACEHOLDER_trend_data_medium
+   - PREFIXPLACEHOLDER_trend_data_rt
+   - PREFIXPLACEHOLDER_trend_data_short
+
+   To find out what the PREFIXPLACEHOLDER is, navigate to cd /data/cassandra/data.
```
MariekeGO, line 49: This is a bit vague. When you navigate to cd /data/cassandra/data, where is it that you can see what the prefix placeholder is?

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
joncorral: I would expect this tag can be enabled for all Cassandra based systems. Anyhow, this procedure has been tested and tried for Linux systems, and as there is no clear documentation on it online, I would limit it to that.

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
joncorral: Yes there are. I realized by first seeing an increase in the disk usage of the data partition of Cassandra. The TWCS keyspaces (trend_data_medium, trend_data_short and trend_data_long) where showing a memory leak behavior, and they never seemed to drop.

I used the cassandra-tools to double check what was going on. The commands I used for installing and finding out what was the issue were the following

```
yum install cassandra-tools	 

sudo systemctl stop cassandra  

sstableexpiredblockers zm1_cc_trend_data_short trend_data_short   

for f in *Data.db; do meta=$(sudo sstablemetadata $f); echo -e "Max:" $(date --date=@$(echo "$meta" | grep Maximum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') "Min:" $(date --date=@$(echo "$meta" | grep Minimum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') $(echo "$meta" | grep droppable) ' \t ' $(ls -lh $f | awk '{print $5" "$6" "$7" "$8" "$9}'); done | sort 
``` 

With the output of the sstableexpiredblockers and the for below it can be seen which tables are blocking others from being dropped. Basically, if the percetange of tombstones in the SSTable is higher than 1, it can be assumed that all data is tombstone data.

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
joncorral: What I meant there is that by navigating to that folder and checking what is inside that folder the user will be able to see what the name of that prefixplaceholder is. That is defined by them afaik when the installation of cassandra is done, but still, it is there where they can check the names.

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,63 @@
+---
+uid: Cassandra_expired_blocking_SSTables
+---
+
+# Expired SSTables blocking other tables from being dropped
+
+In case the only data SSTables in Cassandra contain is tombstones, and the tables are therefore expired, they could block other tables that are also expired from being dropped. This could result in various issues, such as an increase in the disk usage of the Cassandra data partition, or memory leak behavior in the TWCS keyspaces (*trend_data_medium*, *trend_data_short*, and *trend_data_long*).
+
+You can check if you are indeed encountering this issue using the Cassandra tools, for example with the following commands (on Linux):
```
SkylineCookieMonster, line 9: Please add that that the following commands will stop the Cassandra service and therefore could impact the working of any system using the DB (e.g. DataMiner).

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,63 @@
...
+sudo systemctl stop cassandra
+
+sstableexpiredblockers zm1_cc_trend_data_short trend_data_short
+
+for f in *Data.db; do meta=$(sudo sstablemetadata $f); echo -e "Max:" $(date --date=@$(echo "$meta" | grep Maximum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') "Min:" $(date --date=@$(echo "$meta" | grep Minimum\ time | cut -d" " -f3| cut -c 1-10) '+%m/%d/%Y') $(echo "$meta" | grep droppable) ' \t ' $(ls -lh $f | awk '{print $5" "$6" "$7" "$8" "$9}'); done | sort
+```
+
+With the output of the sstableexpiredblockers, you can see which tables are blocking others from being dropped. If the percentage of tombstones in the SSTable is higher than 1, you can assume that all data is tombstone data.
+
+To resolve this issue, we recommend enabling the *unsafe_aggressive_sstable_expiration* option for the Cassandra database. This is something that should only be done to troubleshoot this issue, because it is considered a risky option otherwise. The data in the expired SSTables will be dropped without being checked, but as it consists entirely of tombstone data in this case, this is not a problem.
```
SkylineCookieMonster, line 23: I don't understand why you would only enable this option for troubleshooting or why it is a risky option otherwise. @joncorral can you elaborate?

joncorral (approved):

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
```diff
@@ -0,0 +1,63 @@
...
+
+   ```txt
+   ALTER TABLE PREFIXPLACEHOLDER_trend_data_medium.trend_data_medium
+   with compaction = {
+   'class' : 'TimeWindowCompactionStrategy',
+   'unsafe_aggressive_sstable_expiration' : true
+   };
+   ```
+
+   In most cases, you will need to run this query for the following tables:
```
SkylineCookieMonster, line 56: For Cassandra Cluster databases used by DataMiner, you will ...

SkylineCookieMonster (changes requested):

user-guide/Troubleshooting/Procedures/Cassandra_expired_blocking_SSTables.md
joncorral: I added that remark based on the Apache Cassandra documentation that says the following:

This is a potentially risky option that can lead to data loss or deleted data re-appearing, going beyond what unchecked_tombstone_compaction does for single sstable compaction. 

That is why I would not use this option if not facing this issue.

SkylineCookieMonster: Then I would say:
'Enabling this option on a table could lead to inconsistencies (data loss or re-appearing) when deletes/updates are done on the data in TWCS tables. In general, deletes and updates should be avoided on tables with TWCS and you should always rely on TTL as this could result in the blocking tables issue.'

joncorral: That looks very clear and understandable. Thanks Michiel

joncorral (approved):

SkylineCookieMonster (approved):

PR #3431, opened 2024-07-10
Adding third-party connectors

MariekeGO:


MariekeGO:
@SammyDW, could you check if this PR is OK? Also, should I add anyone else as reviewer?

PR #3452, opened 2024-07-12
Added KI for the MySQL.Data.dll crash

BrentSLC:
Added KI for MySQL.Data.dll crash to be used as jumping off point for analyzers and other docs

### DIFF
--- a/user-guide/Troubleshooting/Known_issues/KI_MySQL_Unhandled_Exception.md
+++ b/user-guide/Troubleshooting/Known_issues/KI_MySQL_Unhandled_Exception.md
@@ -1,0 +1,44 @@
+---
+uid: KI_MySQL_Unhandled_Exception
+---
+
+# MySQL.Data.dll can cause service crashes
+
+## Affected versions
+
+- DataMiner 10.3.0[CU14] and higher
+- DataMiner 10.4.0[CU2] and higher
+- DataMiner 10.4.5 and higher
+
+## Cause
+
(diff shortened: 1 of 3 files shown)


user-guide/Troubleshooting/Known_issues/KI_MySQL_Unhandled_Exception.md
```diff
@@ -0,0 +1,44 @@
+---
+uid: KI_MySQL_Unhandled_Exception
+---
+
+# MySQL.Data.dll can cause service crashes
```
MariekeGO, line 5: Could you rephrase this title so it is more a description of the issue than of the cause? With the current title, people who go looking for the issue based on what they encounter in their system may not know that this is the page they are looking for.

BrentSLC: Will do.

MariekeGO:
@BrentSLC, I've reviewed your PR and made some changes. Could you check if everything is still OK for you? 
Also, I noticed that RN 40200 seemed to be related to this issue. Is this the RN that will introduce the fix for this?

BrentSLC (commented):
Looks good to me.
Yes RN 40200 is indeed related. That RN should prevent SLDataGateway crashes due to this issue (Connectors can use their own MySQL.Data.dll, so those will remain at risk of the crashes.)

MariekeGO (approved):

PR #3463, opened 2024-07-16
Update DataMiner_Compute_Requirements.md

Joachim-Ally-Skyline:


MariekeGO:
This change is not necessary and changes the meaning and tone of the sentence in a way we want to avoid. The original "[as a rule](https://idioms.thefreedictionary.com/as+a+rule)" is correct and is the phrase we intended to use.

Joachim-Ally-Skyline:
Okay, I understand.

PR #3470, opened 2024-07-17
Catalog - Extending supported types

brayunm:


user-guide/Cloud_Platform/Catalog/About_the_Catalog_module.md
brayunm, line 1: @MariekeGO I currently extended this file with the new types, but it could also be a separate page. What are your thoughts ?

brayunm (commented):
This update should also be linked to a catalog UI update, this is still in progress

user-guide/Cloud_Platform/Catalog/About_the_Catalog_module.md
MariekeGO: Looks fine to me here. I've just added some cross-references, so that users who end up on this page can easily find their way to the sections related to some of the mentioned features.

EdithLansens (approved):

PR #3514, opened 2024-07-30
Update FilteringAlarmsUsingSystemNameType.md

aldrickgardiner:
Add live system example and update a text to be a snippet


EdithLansens (approved):
Please use the following format when adding images: `[link text](URL)`. See [Markdown syntax](https://docs.dataminer.services/contributing/CTB_Markdown_Syntax.html).

PR #3533, opened 2024-08-06
CDM

EdithLansens:
Adding the Common Data Model to the user guide


user-guide/Standard_Apps/MediaOps/Common_Data_Model.md
EdithLansens, line 1: @JoeyVanhalstSLC, we have completed the language check of the CDM. Below, you can find an overview of the changes we've made.

PR #3535, opened 2024-08-06
Adding EPM Kata Tutorial

SkylineGBO:
Please if PMa can review and approve since it will be a Kata Tutorial

### DIFF
--- a/user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
+++ b/user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
@@ -278,3 +278,6 @@ In this final step, you will learn how you can link EPM entities to views in Dat
 All the views will now be linked to their EPM objects and inherit their alarm properties from the EPM entity. Selecting any of the views will show a data section displaying the EPM KPI information.
 
 ![Data section with EPM KPI information](~/user-guide/images/EPM_GS_step_5_3.png)
+
+> [!IMPORTANT]
+> As EPM setups can use quite a lot of element data, if you have completed the tutorial and you no longer need the setup from this tutorial, we highly recommend stopping or removing the elements that were added during the tutorial.


MariekeGO:
@EdithLansens, could you add an icon for the EPM tutorials on the DataMiner tutorials overview page?

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
```diff
@@ -0,0 +1,155 @@
...
+This tutorial will provide a brief introduction on the EPM architecture and how you can use many EPM specific features.
+
+Expected duration: 30 minutes
+
+> [!NOTE]
+> The content and screenshots for this tutorial have been created using DataMiner version 10.4.5.
+
+## Prerequisites
+
+- A DataMiner System that is [connected to dataminer.services](xref:Connecting_your_DataMiner_System_to_the_cloud).
```
MariekeGO, line 16: As the [Topology app](https://docs.dataminer.services/develop/DMI/EPM_Integration_Training/Topology_app_configuration.html) is used, should the CPEIntegration soft-launch option also be enabled on the system? Or does this happen automatically when the package is deployed?

SkylineGBO: Indeed, I forgot to mention that. Though the softlaunch option is not needed if working on a DaaS

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+
+   - The **Topology app** in between the Surveyor and Activity module.
+
+   ![Cube sidebar with Topology app](~/user-guide/images/EPM_GS_step_1.png)
+
+> [!NOTE]
+> If the Topology app does not show up right away, restart the Cube client.
+
+## Step 2: Create a new collector element and rebalance the system
+
```
MariekeGO, line 58: I'd like to add a small intro at the top of this step to make it more clear why we are doing this. Could you give me some info as to why the user should start this tutorial by adding a collector element?

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
```diff
@@ -0,0 +1,192 @@
...
+   ![Creation of new Collector element](~/user-guide/images/EPM_GS_step_2_1.png)
+
+1. Right-click the newly created element, select *Properties*, and copy the DMA ID/element ID of the element.
+
+   ![DMA ID/element ID in element properties](~/user-guide/images/EPM_GS_step_2_2.png)
+
+1. Configure the *EPM Training BE 2* element as the parent manager of the *EPM Training Collector 4* element:
+
+   1. Open the *EPM Training BE 2* (1) element, and go to the *Configuration* visual page (2) and then the *Collectors* tab (3).
```
MariekeGO, line 78: I've installed version 1.0.1-CU3 of the EPM Integration Training package on my DMA (running 10.4.8) to test this. However, I don't see this Collectors tab on the configuration visual page:
image
Any idea what went wrong?

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
SkylineGBO: Adding the collector element to the system is to demonstrate how the EPM architecture has horizontal scaling, where a user can add more collectors to reduce the load on individual elements without adding load to the Frontend

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
SkylineGBO: Not sure, just installed on my local as well (running 10.4.8-) and the pages are showing up. Also inspected the visio and the pages are there. Maybe it's a visual bug? Can you close and open the cube window? Also, did the alarm templates get applied to the FE and BE elements?

MariekeGO: As it seems to be a bug that's specifically related to my system, we've found a way around this by using a different client PC.

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
```diff
@@ -37,7 +37,12 @@ This tutorial consists of the following steps:
    > [!TIP]
    > See also: [Deploying a Catalog item to your system](xref:Deploying_a_catalog_item)
 
-1. Open DataMiner Cube and check whether the following items have been added to your DataMiner Agent:
+1. Connect to the DMS with DataMiner Cube using the argument `ENABLEFEATURE=CPEINTEGRATION`, so that the [CPEIntegration](xref:Overview_of_Soft_Launch_Options#cpeintegration) soft-launch features are enabled.
```
SkylineGBO, line 40: I think the note should be made that this is only needed if not using a DaaS system as the DaaS are created with this option enabled. Also, is the ENABLEFEATURE=CPEINTEGRATION and argument that needs to be passed in to cube or just needs to be added to the softlaunch option file?

SkylineGBO (commented):
All looks good, just a small remark on the CPEIntegration part

user-guide/Advanced_Modules/EPM/Tutorials/EPM_Introduction_Tutorial.md
MariekeGO: At first I thought this was an option in the soft-launch options file, but upon checking in more detail, I noticed that the CPEIntegration option is mentioned as a Cube-only option (see https://docs.dataminer.services/user-guide/Reference/Soft-launch_options/Overview_of_Soft_Launch_Options.html#cpeintegration). This means that it doesn't matter if you use DaaS or not, since it's a client-side option. However, if you're not using this option, and everything works fine for you, it could be that the option is not needed for the tutorial after all.

SkylineGBO: Is there a way to set ENABLEFEATURE=CPEINTEGRATION in Cube without the SoftLaunchOptions file? I tested on a server by removing the soft launch option and the topology app went away, so I think the DaaS have the option on by default since I never messed with any of the local Cube options.

MariekeGO: In that case, the information about the CPEIntegration option on docs must be incorrect, and it's in fact not a Cube-only option. Possibly this got changed at some point and the documentation was not updated. Do you see any difference when the Cube option is or isn't used? I did a quick test myself on my local system, and I get the impression that only the SoftLaunchOptions file now seems to affect which features are displayed.

SkylineGBO: I can't find a way to get the Cube option applied without the softlaunch option, so I think the only way to get it working is with the softlaunch option file.

MariekeGO: OK, in that case I'll adjust the PR, and I'll also update the information about this soft-launch option, because the current information is clearly incorrect.

MariekeGO (approved):

PR #3542, opened 2024-08-07
RN 39495

EdithLansens:
@CatarinaBeirolasVilhena, could you please review this pull request to confirm the accurate inclusion of release note 39495 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
+++ b/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
@@ -216,7 +216,7 @@ The following types of alarm events will not be combined in a consolidated alarm
 
 ### AlarmSettings.PersistParameterLatchState
 
-From DataMiner 10.4.9/10.5.0 onwards<!--RN 39495-->, parameter latch states are no longer persistent by default. This means they will reset after every DataMiner restart.
+From DataMiner 10.4.9/10.5.0 onwards<!--RN 39495-->, parameter latch states are not persistent by default. This means they will reset after every DataMiner restart.
 
 To enable persistent parameter latch states, add the following tags and values in *MaintenanceSettings.xml*:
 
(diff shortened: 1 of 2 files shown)


user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
```diff
@@ -214,6 +214,23 @@ The following types of alarm events will not be combined in a consolidated alarm
...
+
+```xml
+<AlarmSettings>
+   <PersistParameterLatchState>true</PersistParameterLatchState>
+</AlarmSettings>
+```
+
+> [!NOTE]
+>
+> - This setting will only be applied after a DMA restart.
```
MariekeGO, line 231: As changes to MaintenanceSettings.xml always require a DMA restart (as mentioned at the top of this document), it might be better not to mention this again here. Otherwise people might get the impression that some changes need a restart, and some don't.

user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
```diff
@@ -214,6 +214,23 @@ The following types of alarm events will not be combined in a consolidated alarm
 > - [Alarm linking](xref:Alarm_linking)
 > - [Declutter your alarm tree](https://community.dataminer.services/declutter-your-alarm-tree/ "Alarm squashing")
 
+### AlarmSettings.PersistParameterLatchState
+
+From DataMiner 10.4.9/10.5.0 onwards<!--RN 39495-->, parameter latch states are no longer persistent by default. This means they will reset after every DataMiner restart.
```
MariekeGO, line 219: To make this more future-proof, I'd prefer to avoid using "no longer" (especially since at some point the version info will get removed here).
```suggestion
From DataMiner 10.4.9/10.5.0 onwards<!--RN 39495-->, parameter latch states are not persistent by default. This means they will reset after every DataMiner restart.
```

user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
```diff
@@ -214,6 +214,23 @@ The following types of alarm events will not be combined in a consolidated alarm
...
+```xml
+<AlarmSettings>
+   <PersistParameterLatchState>true</PersistParameterLatchState>
+</AlarmSettings>
+```
+
+> [!NOTE]
+>
+> - This setting will only be applied after a DMA restart.
+> - If the `PersistParameterLatchState` option is set to "false" (or if this tag has not been added to the *MaintenanceSettings.xml* file), parameter latch states will no longer be written to or fetched from the database.
```
MariekeGO, line 232:
```suggestion
> -  From DataMiner 10.4.9/10.5.0 onwards, if this tag has not been added to the *MaintenanceSettings.xml* file, or if the `PersistParameterLatchState` option is set to "false", parameter latch states will not be written to or fetched from the database.
```

MariekeGO (changes requested):

MariekeGO (approved):

PR #3547, opened 2024-08-07
DxP & DxMs .NET 8

JordyGit:


release-notes/DxMs/artifactdeployer_change_log.md
```diff
@@ -4,6 +4,10 @@ uid: artifactdeployer_change_log
 
 # ArtifactDeployer change log
 
+#### 8 August 2024 - Enhancement - ArtifactDeployer 1.8.0 - Upgrade to .NET 8 [ID_40441]
+
+DataMiner ArtifactDeployer has been upgraded to .NET 8. **Make sure .NET 8 is installed** before you upgrade to this version. [NET 8 is included since DataMiner 10.X](xref:change log of DM version that includes net8).
```
MariekeGO, line 9: To my knowledge, .NET 8 is not included in DataMiner and needs to be installed manually. However, it is required since DataMiner 10.3.0 [CU12]/10.4.0/10.4.3. (See [Compute requirements](https://docs.dataminer.services/user-guide/Reference/DataMiner_Compute_Requirements.html#microsoft-net) & [RN 37969](https://docs.dataminer.services/release-notes/General/General_Feature_Release_10.4/General_Feature_Release_10.4.3.html#dataminer-requirements-net-80-now-required-id_37969))

JordyGit: I checked with Bert B and it's indeed not included, so makes sense that i couldn't find it in the release notes. An DM upgrade will block if NET 8 is not installed yet.
We can just removed the "[NET 8 is included since DataMiner 10.X](xref:change log of DM version that includes net8)." parts.
Maybe we need to add a section in the table at https://docs.dataminer.services/user-guide/Reference/DataMiner_Compute_Requirements.html#microsoft-net for DxMs?

MariekeGO: > Maybe we need to add a section in the table at https://docs.dataminer.services/user-guide/Reference/DataMiner_Compute_Requirements.html#microsoft-net for DxMs?

That might indeed be an option, though I'm concerned that this might lead to a very complex overview, making this page less user-friendly.
Are users expected to always use the latest version of the DxMs? Because if so, it may not be necessary to list the requirements for past DxMs. In that case, we only need to make sure that the requirements for the latest version are clear. Is this mentioned in the Admin app itself, when users are about to upgrade to a version with different requirements? They are more likely to notice that than a complex overview in the user guide.

JordyGit: When they update in the Admin app a popup will be shown indeed. Also the text on the page where they download the cloud pack will have to be updated (I tagged TW in the email that I've just sent).

MariekeGO (approved):

PR #3560, opened 2024-08-12
Compute requirements terminology adjustment

MariekeGO:


MariekeGO:
@EdithLansens, could you update _dataminer-compute-requirements_V04_Outline.svg_ to mention vCPUs instead of cores?

PR #3584, opened 2024-08-20
Update CPU requirements

MariekeGO:


MariekeGO:
@EdithLansens could you update the image on the compute requirements page so it only mentions the vCPUs for the CPU?

image

EdithLansens:
> @EdithLansens could you update the image on the compute requirements page so it only mentions the vCPUs for the CPU?
> 
> image

The image has been updated as requested! I also added ".NET 8" to the *Frameworks* section since [the text](https://docs.dataminer.services/user-guide/Reference/DataMiner_Compute_Requirements.html#microsoft-net) references this version, but it wasn't included in the image.
image

BertVandenberghe (approved):

PR #3591, opened 2024-08-22
Update Layer_types.md

jarno-lernou:
Add important note to the relations layer.

### DIFF
--- a/user-guide/Advanced_Modules/Maps/Configuring_DataMiner_Maps/Layer_types.md
+++ b/user-guide/Advanced_Modules/Maps/Configuring_DataMiner_Maps/Layer_types.md
@@ -387,6 +387,9 @@ Set the *sourceType* attribute of a layer to "properties" if you want that layer
 
 Set the *sourceType* attribute of a layer to "relations" if you want that layer to display lines between objects that are related via foreign key relationships in a DataMiner element. If, for example, an EPM element has a "Household" and an "Amplifier" table that are linked by means of foreign key relationships, the links between those two tables can be visualized on a map.
 
+> [!IMPORTANT]
+> Do note that the Source Table is the Table having the Foreign Key Column. The Destination table is the table pointed to by the Foreign Key defined in the Source table. Switching the configuration will fail to show the relation.
+
 Example:
 
 ```xml


user-guide/Advanced_Modules/Maps/Configuring_DataMiner_Maps/Layer_types.md
```diff
@@ -387,6 +387,9 @@ Set the *sourceType* attribute of a layer to "properties" if you want that layer
 
 Set the *sourceType* attribute of a layer to "relations" if you want that layer to display lines between objects that are related via foreign key relationships in a DataMiner element. If, for example, an EPM element has a "Household" and an "Amplifier" table that are linked by means of foreign key relationships, the links between those two tables can be visualized on a map.
 
+> [!IMPORTANT]
+> The source table is the table that contains the foreign key column. The destination table is the table pointed to by the foreign key defined in the source table. Switching the configuration will fail to show the relation.
```
MariekeGO, line 391: What do you mean with "Switching the configuration will fail to show the relation."? Do you mean that it will not work if you specify a source table that does not contain the foreign key column, or a destination table that is not pointed to by the foreign key?

jarno-lernou: Hi, I had to investigation for a long time why my configuration was not working. I had a parent (e.g. country) and a child (e.g. province). As the source I had specified the Country and destination Province as it was logical in my mind to specify it like this. This did not work. Instead I had to use the 'Province' as the source table and the 'Country' as destination.

The source table, i.e. Province has a 'FK Column' with a value that is linked to the Primary Key of the destination table (Country). This way you can have multiple provinces linked to one country.

MariekeGO: I've adjusted the description, and I've also modified the order of the paragraphs to make it a bit more logical. Could you check if everything still looks ok for you?

jarno-lernou: Clear and OK.

MariekeGO (approved):

PR #3592, opened 2024-08-22
Added health check installer page

sofianadramiaskyline:
Added the documentation for the new Skyline Health Check package. Included this in the tools folder.


MariekeGO:
Reverted to draft as the solution is not yet ready for publication. @sofianadramiaskyline, could you confirm whether we should close this pull request, or if you want to continue with this same pull request and mark it as ready for review when the solution is ready?

PR #3598, opened 2024-08-23
howto decrypt snmpv3 capture

PedroDebevere:


develop/TOOLS/ThirdPartyTools/WiresharkDecryptTraffic.md
```diff
@@ -0,0 +1,28 @@
+---
+uid: WiresharkDecrypt
+---
+
+# Wireshark stream decryption
```
MariekeGO, line 5: It would be better if the title here and the title in the toc.yml were the same. Which is more appropriate, "Wireshark stream decryption" or "Wireshark package decryption"?

PedroDebevere: Agreed, I would opt for "Wireshark stream decryption"

MariekeGO (approved):

PR #3657, opened 2024-09-06
Update Preparing_the_two_DataMiner_Agents.md

joaobarraca:
Added STaaS reference

### DIFF
--- a/user-guide/Advanced_Functionality/Failover/Configuring_Failover/Preparing_the_two_DataMiner_Agents.md
+++ b/user-guide/Advanced_Functionality/Failover/Configuring_Failover/Preparing_the_two_DataMiner_Agents.md
@@ -15,7 +15,7 @@ Before you start the actual configuration, make sure you have the following:
   > [!NOTE]
   > If you use IP addresses instead of a hostname:
   >
-  > - Make sure [NPCap](https://nmap.org/npcap/) or WinPCap (deprecated) is installed on both DMAs (see [Before you run the installer](xref:Installing_DM_using_the_DM_installer#before-you-run-the-installer)).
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Failover/Configuring_Failover/Preparing_the_two_DataMiner_Agents.md
```diff
@@ -66,6 +66,10 @@ Make sure that packets to and from these ports coming from the virtual IP addres
 
 Each [supported system data storage architecture](xref:Supported_system_data_storage_architectures) has a different way of handling the setup of a Failover system. Below you can find the measures that need to be taken for each of the supported architectures.
 
+> [!IMPORTANT]
+>
+> If your system has been configured to use STaaS, make sure you follow the steps described in [Adding a DMA to a DMS running STaaS](xref:Adding_a_DMA_to_a_DMS_running_STaaS).
```
MariekeGO, line 71: I'm not sure if this will be clear enough, as those steps also involve actually adding the DMA to the cluster, and the backup Agent should not be joined to the cluster yet at this point. I'd also add this under a dedicated subheader, as I see no reason to only put in a note for STaaS while the other possible database setups get a dedicate section. Would the following be OK?

```suggestion
### Storage as a Service

If you want to add the Failover pair to a DataMiner System that uses STaaS, first make sure the primary DMA has been added as detailed under [Adding a DMA to a DMS running STaaS](xref:Adding_a_DMA_to_a_DMS_running_STaaS). Then follow the same steps for the backup DMA, but skip the step where you actually join the DMA to the cluster.
```

joaobarraca: indeed. looks better like this!

PR #3677, opened 2024-09-12
Update Protocol.Params.Param.Measurement.md

RobbieDemaegdt:


EdithLansens:
@PedroDebevere, could you also review this one (identical input but in a different location)?

EdithLansens (approved):

PR #3684, opened 2024-09-13
GQI - Added scaling best practice and example

SebastiaanSL:


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Tutorials_Scaling.md
MariekeGO, line 1: I'm not sure if this really belongs among the tutorials, since it's just an example rather than a step by step guide on how to do something. I'd prefer to move this elsewhere. We could for instance add this example directly on the best practices page, instead of adding a link there. What do you think?

SebastiaanSL: Yeah, as long as users will find it 👍 
I took a similar approach as https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Extensions/Ad_hoc_data/Ad_hoc_Tutorials.html though, so perhaps this one should be moved as well.

MariekeGO: Indeed, that page also doesn't quite seem to be in the right place, but finding a different spot for it is a bit more tricky. One thing we could do is move it up:
image
In that case, an alternative for the current page would also be to move it to the same location, instead of integrating it in the best practices page. Which would you prefer?

SebastiaanSL: I'm not that in favor of moving it up a level. The information above is actually new information, while these 2 tutorials are examples or best practices.

MariekeGO: Perhaps we could add an "Examples" item above "Tutorials" then, and add both items under that?

SebastiaanSL: Sure, I'm fine with that!

MariekeGO:
@SebastiaanSL, I've reviewed the PR and moved the pages as suggested. Does everything look OK for you?

SebastiaanSL (commented):
Looks good!

MariekeGO (approved):

PR #3686, opened 2024-09-16
Crop remote access URL image

AlexanderVK-SLC:


AlexanderVK-SLC:
Not sure this is needed, seems to look OK on the docs but the file has whitespace.

MariekeGO (approved):

PR #3701, opened 2024-09-17
Automatin (I)Engine - depcreate and add methods dealing with alarms

RobbeDGH:
[[RN40773](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=40773)] [[DCP246488](https://collaboration.skyline.be/task/246488)]

### DIFF
--- a/src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
+++ b/src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
@@ -10,7 +10,7 @@ namespace Skyline.DataMiner.Net
     [Serializable]
     public sealed class AlarmTreeID //: DMAObjectRef, IEquatable<AlarmTreeID>
     {
-        // default constructor is required for protobuf
+        // Default constructor is required for protobuf.
         public AlarmTreeID() { }
 
         /// <summary>
--- a/src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
+++ b/src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
@@ -10,7 +10,7 @@ namespace Skyline.DataMiner.Net
     [Serializable]
     public sealed class AlarmTreeID //: DMAObjectRef, IEquatable<AlarmTreeID>
     {
-        // default ctor is required for protobuf
+        // default constructor is required for protobuf
         public AlarmTreeID() { }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
```diff
@@ -0,0 +1,31 @@
...
+using Skyline.DataMiner.Net.Messages.SLDataGateway;
+using System;
+
+namespace Skyline.DataMiner.Net
+{
+    /// <summary>
+    /// Represents a reference to one specific alarm tree.
+    /// </summary>
+    [Serializable]
+    public sealed class AlarmTreeID //: DMAObjectRef, IEquatable<AlarmTreeID>
```
wdm-slc, line 11: do we also need to list the properties like ElementID and RootAlarmID ?

RobbeDGH:
@wdm-slc, to see the resulting docs from this you need to use buildDocs.cmd instead of the raw docfx build command

src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
RobbeDGH: seems like a good idea yeah

RobbeDGH:
Note there is somthing wrong with displaying the AlarmID class, not sure what.
This is not fixed.

Think it might be related to the namespace Skyline.DataMiner.Net.Messages.SLDataGateway being present in multiple dlls

wdm-slc (approved):
lgtm

src/DataMiner/SLNetTypes/Info/IDObjects/AlarmTreeID.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        // default ctor is required for protobuf
```
MariekeGO, line 13: Is this correct, or is "ctor" a typo?

RobbeDGH: It is correct. 'ctor' is often used as a shorthand for 'constructor'.

I copied these parts over from the official source.
Perhaps we should use the full word for public documentation though. 
Up to you

MariekeGO: indeed, I'll adapt this to constructor so it's clear for everyone

MariekeGO (approved):

PR #3705, opened 2024-09-17
Update DCP_change_log.md

bentvandenheede:


release-notes/Cloud/DCP_change_log.md
```diff
@@ -9,6 +9,10 @@ The dataminer.services platform gets updated continuously. This change log can h
 > [!NOTE]
 > Many features on dataminer.services are dependent on DxMs. You can find the change logs for these under [DxM release notes](xref:DxM_RNs_index).
 
+#### 17 September 2024 - Fix - Caching issues on the index file of Cloud web apps [ID 40777]
+
+Adjusted the caching headers for Cloud web apps to improve caching.
```
MariekeGO, line 14: Could you adjust this description so it's clear what went wrong, from the perspective of user? This is necessary so that a user that may have encountered the issue can recognize it in the release notes.

bentvandenheede: I changed the explanation.

MariekeGO (approved):

PR #3708, opened 2024-09-18
User guide updated for RN 40768

MariekeGO:


AlexanderVK-SLC (approved):
Good catch, I'll go through the docs again today to see if other things are out of date now as well.

PR #3709, opened 2024-09-18
Update regarding adding a DMA to a DMS

XanderWA:


MariekeGO:
> The _Join cluster_ button is only needed when initially going from a standalone DMA to a DMS (cluster of one). Cube has been sending the `SetDmsClusterMessage(clustername)` message to the DMA being added to a cluster since DataMiner 8.5 (RN7748).

In that case, it might even be better to not add the note about the cluster name, as this could cause confusion (i.e. people might wonder in which case you do have to set the cluster name, while they actually don't have to do it at all).

XanderWA:
My reasoning behind putting the note there is that people could be confused as they remember the procedure being different in the past. However, I agree with your reasoning for not putting it there too.

user-guide/Advanced_Functionality/DataMiner_Systems/Adding_a_DMA/Adding_a_regular_DataMiner_Agent.md
```diff
@@ -14,18 +14,15 @@ uid: Adding_a_regular_DataMiner_Agent
...
 
 1. In the *Manage* tab, below the list of DMAs in the cluster, click *Add*.
 
 1. Enter the IP address of the DMA you want to add and click *Add*.
 
+> [!NOTE]
+> You should not set the cluster name for each DMA. This will be done automatically upon adding the DMA to the DMS.
+
 > [!NOTE]
 > To add a Failover pair to a DataMiner System, first add the primary DMA as a regular DataMiner Agent, and then configure Failover. See [Preparing the two DataMiner Agents](xref:Preparing_the_two_DataMiner_Agents).
```
MariekeGO, line 27:
```suggestion
> [!NOTE]
>
> - It is not necessary to set the cluster name when you add a DMA to a DMS. The name will be set automatically.
> - To add a Failover pair to a DataMiner System, first add the primary DMA as a regular DataMiner Agent, and then configure Failover. See [Preparing the two DataMiner Agents](xref:Preparing_the_two_DataMiner_Agents).
```

user-guide/Advanced_Functionality/DataMiner_Systems/Setting_up_a_new_DMS/Setting_up_a_new_DMS_in_DataMiner_Cube.md
```diff
@@ -19,20 +19,15 @@ uid: Setting_up_a_new_DMS_in_DataMiner_Cube
...
-
-   1. In the pane on the right, next to *Cluster*, enter the same DMS name as you specified for the master DMA, and then click *Add Cluster* (prior to DataMiner 10.0.13) or *Join cluster* (from DataMiner 10.0.13 onwards).
-
    1. Open DataMiner Cube on the primary DMA, and navigate to the *Agents* > *Manage* tab as before.
 
    1. In the lower left corner, click *Add*.
 
    1. Enter the IP of the DMA you want to add and click *Add*.
 
+> [!NOTE]
```
MariekeGO, line 28: Similar to the comment above, I'd recommend combining this with the note below. We usually try to avoid adding several note blocks close to one another and combine the notes into one block instead.

user-guide/Advanced_Functionality/DataMiner_Systems/Adding_a_DMA/Adding_a_regular_DataMiner_Agent.md
```diff
@@ -14,20 +14,16 @@ uid: Adding_a_regular_DataMiner_Agent
...
 1. On a DMA that is currently already in the cluster, go to the *System Center* module and select the *Agents* page.
 
 1. In the *Manage* tab, below the list of DMAs in the cluster, click *Add*.
 
 1. Enter the IP address of the DMA you want to add and click *Add*.
 
 > [!NOTE]
-> To add a Failover pair to a DataMiner System, first add the primary DMA as a regular DataMiner Agent, and then configure Failover. See [Preparing the two DataMiner Agents](xref:Preparing_the_two_DataMiner_Agents).
+>
+> - You should not set the cluster name for each DMA. This will be done automatically upon adding the DMA to the DMS.
```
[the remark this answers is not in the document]
XanderWA: My change from "It is not necessary to" to "You should not" was done after Bert mentioned to me that "It is not necessary to" still sounds rather optional and you can still set a cluster name if you want to do this.

Yes, this can have a negative effect as this could lead to the wrong agent being interpreted as being "in sync". This will cause data to be lost from the other agent. However, this can only happen if the user doesn't follow the exact procedure. By not setting this this can never happen, thus this is a more robust procedure.

user-guide/Advanced_Functionality/DataMiner_Systems/Setting_up_a_new_DMS/Setting_up_a_new_DMS_in_DataMiner_Cube.md
```diff
@@ -38,4 +30,6 @@ Now all DataMiner Agents that have been added to the DMS will start the initial
 When the DMS has been synchronized, you will be able to consult information about each of the Agents in the *Status* tab of the *Agents* page in *System Center*.
 
 > [!NOTE]
-> It is also possible to migrate elements from one DMA to another within a DMS. For more information, see [Migrating elements in a DataMiner System](xref:Migrating_elements_in_a_DataMiner_System).
+>
+> - You should not set the cluster name for each DMA. This will be done automatically upon adding the DMA to the DMS.
```
MariekeGO, line 34:
```suggestion
> - Only use the *Join cluster* button when you set up the master DMA. When you add other DMAs, the cluster name will be set automatically.
```

Note that it's no longer necessary to mention the previous name of the button as versions prior to 10.0.13 are no longer supported (we'll need to remove that obsolete info from the text too, but that doesn't necessarily have to happen with this PR – it's something we're planning to do across the user guide later).

MariekeGO (approved):

PR #3710, opened 2024-09-18
[STaaS] Updated procedure to add a DMA to a cluster

GuillaumeVadeBe:
Added an extra step to stop both the NAS and NATS services before copying over the nsc folder.


joaobarraca (approved):

MariekeGO:
Should there also be a restart step for these services, or will they restart automatically when DataMiner is restarted?

GuillaumeVadeBe:
@MariekeGO They will automatically start when the DMA starts up

MariekeGO (approved):

PR #3714, opened 2024-09-19
Terminology: "Catalog item" vs "Catalog artifact"

SLC-brechtseys:
An item is what can be found on the Catalog. A Catalog item can contain one or more artifacts. Cleaned up the terminology accordingly.

### DIFF
--- a/develop/CICD/Skyline Communications/Publish to Catalog/Publish_To_Catalog.md
+++ b/develop/CICD/Skyline Communications/Publish to Catalog/Publish_To_Catalog.md
@@ -2,7 +2,7 @@
 uid: publish_to_catalog
 ---
 
-# Publishing items on the DataMiner Catalog
+# Publishing items in the DataMiner Catalog
 
 The [DataMiner Catalog](https://catalog.dataminer.services/) allows users to quickly look up, download, and deploy various DataMiner-related items.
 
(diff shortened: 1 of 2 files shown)


develop/CICD/Skyline Communications/Publish to Catalog/Publish_To_Catalog.md
```diff
@@ -2,11 +2,11 @@
 uid: publish_to_catalog
 ---
 
-# Publishing artifacts to the DataMiner Catalog
+# Publishing items on the DataMiner Catalog
```
MariekeGO, line 5: "on" seems like a weird choice of preposition here. I think "in" would sound more natural:
```suggestion
# Publishing items in the DataMiner Catalog
```

Note also that the corresponding title in the [toc.yml](https://github.com/SkylineCommunications/dataminer-docs/blob/main/develop/toc.yml) should also be adjusted if we adjust the title here.

SLC-brechtseys: I've doubted about it, even asked ChatGPT. It suggested using "in" a catalog if it was an actual paper catalog, but "on" a catalog if it was an online platform.
But I'm OK with either!

MariekeGO (approved):

PR #3729, opened 2024-09-23
Update Pricing_Usage_based_service.md

bfsantos:
adding text on charges to change connectors

### DIFF
--- a/dataminer-overview/Pricing/Pricing_Usage_based_service.md
+++ b/dataminer-overview/Pricing/Pricing_Usage_based_service.md
@@ -36,8 +36,6 @@ By default, DataMiner credits are deducted monthly based on the metered usage (p
 | Annual automatic renewal | Yes | N/A |
 | Annual subscription fee | Yes | N/A |
 | Over-utilization | Charged at the standard Pay-per-use fee | N/A |
-| New connector development | MUA < 200: charged as professional services credits or, optionally, you can purchase the perpetual license for cost control. <br>MUA >= 200: included with no additional costs, subject to a fair use policy |
-| Changes to existing connectors | MUA < 200: charged as professional services credits <br> MUA >= 200: included |
 | Support services | MUA < 200: Community Plan<br>MUA >= 200: Continuity/Evolve | Community Plan |
 
 ### New connector development and changes to existing connectors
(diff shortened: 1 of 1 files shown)


dataminer-overview/Pricing/Pricing_Usage_based_service.md
```diff
@@ -36,9 +36,16 @@ By default, DataMiner credits are deducted monthly based on the metered usage (p
 | Annual automatic renewal | Yes | N/A |
 | Annual subscription fee | Yes | N/A |
 | Over-utilization | Charged at the standard Pay-per-use fee | N/A |
-| New connector development | MUA < 200: charged as professional services credits<br>MUA >= 200: included with no additional costs, subject to a fair use policy | N/A |
+| New connector development | MUA < 200: charged as professional services credits or, optionally, you can purchase the perpetual license for cost control. <br>MUA >= 200: included with no additional costs, subject to a fair use policy |
+| Changes to existing connectors | MUA < 200: charged as professional services credits <br> MUA >= 200: included |
```
MariekeGO, line 40: What should be added in the "Pay-per-Use" column for these two rows? We can also leave the column empty if necessary, but we can't leave it out.

bfsantos: Sorry, that row on the table should be removed. I've added the text under the table

MariekeGO: Do you mean that line 40 should be removed, but line 39 should be kept, or should both of them be removed?

bfsantos: both (sorry for the lack of clarity). They're replaced by the text I've added right after the table

### New connector development and changes to existing connectors

In general, a system running as a usage-based service can use any of the existing connectors available in the [DataMiner Catalog](https://catalog.dataminer.services/) as is, regardless of whether they have active subscriptions or are just on PPU. This also includes new connector versions as they become available.

Organizations with 200+ MUA can request changes to existing connectors as well as new connectors, with no additional costs, **subject to a fair use policy**. However, for organizations under 200 MUA, these services are charged as professional service credits or, optionally, through the purchase of a perpetual license, to lock the price of new connector development.

MariekeGO (approved):

PR #3731, opened 2024-09-24
Add missing weight option

SkylineThomasCR:
Added documentation for the weight option.
Moved example 1 to the weight option section as weight is the only feature it showcases.

### DIFF
--- a/develop/devguide/Connector/Actions/ActionAggregate.md
+++ b/develop/devguide/Connector/Actions/ActionAggregate.md
@@ -377,11 +377,9 @@ In the following example, the result values of the avg extended will be placed i
 ```
 #### weight
 
-Specifies the weight for each row in the aggregation, representing the frequency or number of occurrences of a value.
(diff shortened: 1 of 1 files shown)


LaurensMT:
I find it a bit confusing how a weight is expressed as "_which represents how many times the corresponding value in column 102 occurs_" . Lets say that we want to measure the average code review remarks per person. We want to see this result per floor and in total. We have floor 1 with 10 persons, their individual results are 20,30,40,50,50,50,60,70,70,80. We have floor 2 with 2 persons, their individual results are 20,30. When aggregating this on floor level, this results in 52 for floor 1 with a weight of 10, and 25 for floor 2 with a weight of 2. We can't say here that 52 is 10 times present. 52 is the avg value which is represented by 10 persons. When aggregating from floor to total this results in 47.5 when taking the weight into account (52*10) + (25 *2) / (10 + 2) (similar like on the lowest level taking the individual sum for all persons divided by 12 persons). Without the weight this would be (52+25)/2 = 38.5.

LaurensMT (approved):

MariekeGO (approved):

PR #3749, opened 2024-09-30
Update Overview_Training_certification.md

EmmaSaldi:
Added exam guidelines for Fundamental and Configurator certification processes as requested by some customers. Once you reviewed this, I would like to have the approval of Bert Vandenberghe before publishing it.

### DIFF
--- a/dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
+++ b/dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
@@ -16,79 +16,99 @@ For example, the DataMiner Fundamentals Certification is intended for DataMiner 
 > For more information about the different certifications and their process, please visit [DataMiner Certification](https://community.dataminer.services/learning/certification/).
 
 ## DataMiner Fundamentals exam guide
-This section is designed to help you understand the process, exam format, scoring criteria, and provide you with preparation tips. Our goal is to ensure you have all the information you need to succeed. 
 
-### Certification process 
-As soon as the ordering process is finalized and your certification process can be started, you receive an automated email from Skyline Academy with information about the process and the deadline. 
+Are you preparing to take the DataMiner Fundamentals exam? Below you can find all the information you need, including the certification process, the exam format, the scoring criteria, and tips on how to prepare for the exam.
+
+### Certification process
+
+As soon as the ordering process is finalized and your certification process can be started, you will receive an automated email from Skyline Academy with information about the process and the deadline.
+
+You will also receive an email with your user ID and password to access the training DMA in order to practice and to take the exam.
+
+To book an appointment, you will need to use the booking link that you will find in the email you received with information about the start of the certification process.
 
 > [!NOTE]
-> - The certification proces is valid for 3 months, unless agreed otherwise.
-> - You'll also receive an email with your user ID and password to access the training DMA in order to practice and to take the exam.
-> - To book an appointment you'll need to use the booking link that you will find in the email you received with information about the start of your process.
-> - An appointment can be cancelled or rescheduled up to 96 hours before the start. Please, contact Skyline Academy or your team of examiners if you cannot update the booking appointment online.
+>
+> - The certification process is valid for 3 months, unless otherwise agreed.
+> - An appointment can be canceled or rescheduled up to 96 hours before it starts. If you cannot update the appointment online, please contact Skyline Academy or your team of examiners.
 > - If you ordered two or more certifications, we will only start the next process once you finished the ongoing process.
 
 ### Exam format and grading
-The exam consist of a verbal online exam through Microsoft Teams with one of our official DataMiner trainers. The total duration of the exam is maximum 1 hour and in order to pass the exam you need to score at least 85%.
 
-The examiner is using a predefined exam plan to ask different questions per module of the training. The questions vary from strict theoretical to showing things in the Training DMA. 
+The exam consist of a verbal exam with one of our official DataMiner trainers. It must be taken online using Microsoft Teams and has a maximum duration of 1 hour. To pass the exam, you need to score at least 85%.
+
+The examiner will use a predefined exam plan to ask different questions per module of the training. The questions vary from strictly theoretical questions to showing things on the training DMA.
 
 > [!NOTE]
+>
 > - Turn on your camera during the exam.
 > - Exams are conducted in English.
-> - By default, the training DMA will be used for conducting the exam.
+> - By default, the training DMA will be used to conduct the exam.
 > - You are allowed to use DataMiner Docs during the exam.
-> - If you don't pass the exam, you have one free retake.
+> - If you do not pass the exam, you have one free retake.
 > - Exams are recorded to ensure the integrity of the process and to provide a reference for review if needed.
 
 ### Preparation tips
+
 To prepare effectively for the exam:
+
 - Study materials: review the [recommended course](https://community.dataminer.services/courses/dataminer-operator/) and [online resources](https://docs.dataminer.services/).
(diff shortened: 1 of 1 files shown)


dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
```diff
@@ -14,3 +14,101 @@ For example, the DataMiner Fundamentals Certification is intended for DataMiner
...
+
+As soon as the ordering process is finalized and your certification process can be started, you will receive an automated email from Skyline Academy with information about the process and the deadline.
+
+You will also receive an email with your user ID and password to access the training DMA in order to practice and to take the exam.
+
+To book an appointment, you will need to use the booking link that you will find in the email you received with information about the start of the certification process.
+
+> [!NOTE]
+>
+> - The certification process is valid for 3 months, unless otherwise agreed.
```
MariekeGO, line 32: Do you mean that the maximum duration of the certification process is 3 months, unless otherwise agreed?

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
```diff
@@ -14,3 +14,101 @@ For example, the DataMiner Fundamentals Certification is intended for DataMiner
...
+- Study materials: review the [recommended course](https://community.dataminer.services/courses/dataminer-administrator/) and [online resources](https://docs.dataminer.services/).
+- Practice questions: use the [quiz](https://community.dataminer.services/courses/dataminer-administrator/quizzes/dataminer-administrator/) at the end of the course to practice. The quiz randomly selects 10 questions out of a pool of questions.
+- Schedule check-ins with your examiner via the booking page if you have any questions.
+
+### Examples questions
+
+- What technology is used to manage and execute backup schedules?
+- Is it possible to test queries within Cube? If so, where can you find this feature?
+- How would you create a sample scheduled task with specific details such as name, schedule, time of execution, and actions?
+- How do you define the relationship between objects in Microsoft Visio with DataMiner shapes for elements, views, etc.?
```
MariekeGO, line 113: This seems a bit odd. Is this what you mean?
```suggestion
- How do you define the relationship between shapes in Microsoft Visio and DataMiner objects such as elements, views, etc.?
```

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
```diff
@@ -16,79 +16,99 @@ For example, the DataMiner Fundamentals Certification is intended for DataMiner
...
-As soon as the ordering process is finalized and your certification process can be started, you receive an automated email from Skyline Academy with information about the process and the deadline. 
+Are you preparing to take the DataMiner Fundamentals exam? Below you can find all the information you need, including the certification process, the exam format, the scoring criteria, and tips on how to prepare for the exam.
+
+### Certification process
+
+As soon as the ordering process is finalized and your certification process can be started, you will receive an automated email from Skyline Academy with information about the process and the deadline.
+
+You will also receive an email with your user ID and password to access the training DMA in order to practice and to take the exam.
+
+To book an appointment, you will need to use the booking link that you will find in the email you received with information about the start of the certification process.
```
EmmaSaldi, line 28: I also want to add the info that candidates receive monthly reminders by email so don't forget about the process. 
20, 10, 5 days before the deadline they also receive emails to inform them that their process will expire soon. Once expired they also receive an email. 
This information is the same for the configurator process

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
EmmaSaldi: The candidate has max. 3 months to finalize the process (pass the exam incl. retakes). Unless agreed otherwise because some organizations want their employees to pass the exam within 1 month after the process started.

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
```diff
@@ -16,79 +16,99 @@ For example, the DataMiner Fundamentals Certification is intended for DataMiner
...
```
EmmaSaldi, line 29: I also want to add that candidates receive automatic reminders via email so they don't forget about their process.
20, 10, 5 days before the expiration date of their process they also receive an email to inform them that their process will expire soon. Once expired they are also notified via email. 
This info is the same for the configurator process.

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
EmmaSaldi: Will need to check with one of the trainers. 
The questions  they use in the exam plan is:
How do you define the relationship between objects in MS Visio with DataMiner shapes for elements, views, etc....

Answer: 
You use shape data in visio. The labels and value fields need to be used.

MariekeGO: If that is the answer, I think the way I rephrased it is indeed better, because strictly speaking there is no such thing as "DataMiner shapes", but instead there are Visio shapes referring to DataMiner objects.

EmmaSaldi (commented):
Ok for me - thank you

dataminer-overview/Training_courses_and_certification/Overview_Training_certification.md
EmmaSaldi: thanks - I've updated the question in the exam plan as well

EmmaSaldi (commented):
Ok

BertVandenberghe (changes requested):
Two small remarks:
- I have the impression a large part of the text is repeated, wouldn't it be better to make a common part and just make a distinction between the 2 trainings in the preparation and question section only?
- I also miss a bit context for the questions. I think it would be better to add about which module a question is because sometimes the question itself does not always mention about what it is talking and therefore a bit confusing.

I believe the second bullet point needs some improvement before we can release this,

MariekeGO:
@EmmaSaldi, would it be OK for you if I change the text so only one "## DataMiner Fundamentals and Configurator exam guide" title is added, where I differentiate between the two wherever necessary in the text?
Also, could you let me know which module each question applies to?

EmmaSaldi:
Agree with Bert; Thanks for updating. 

Regarding the questions, if it's okay I would like to select other questions because we have two different content (recordings on Dojo with the old slides and open training and onsite training with the new slides) so I selected questions that are related to both instead of only old videos). Once the recordings on Dojo have been updated I can send updated questions. This will be end 2024 - Q1 2025.

Fundamentals:
[INTRO] List and explain at least two characteristics of the DataMiner Platform.
[GENERAL USER INTERFACE] Show me where I can find the surveyor with all the elements and open an element. 
[ELEMENT DISPLAY DATA] Can you lock an element and explain what this mean? 
[TRENDING] Where can I see trend prediction?
[CUBE SIDEBAR] Create a new property on your element and show it in the surveyor.
[ALARMS] Where can you see how your alarm console looked like yesterday?
[VISUAL OVERVIEW] What are Visio drawings used for in DataMiner?
[ELEMENT OVERVIEW] By default, the hidden elements are not shown in the element overview. How can you enable that option?
[DASHBOARDS & LOW-CODE APPS] What is the difference between a dashboard and a low-code app?

Configurator:
[STREAM VIEWER] Open the Stream viewer on your IRD element.
[ELEMENTS] Show how you can add a read-only element or make an existing element read-only.
[PROTOCOLS] Name at list one method to add a new protocol version. 
[ALARM TEMPLATES] What is the purpose of hysteresis on alarm templates, and what is the difference between HYST OFF and HYST ON?
[TREND TEMPLATES] Describe the difference between real-time trending and average trending.
[SECURITY] How do you give additional rights to a specific user from a user group?
[VIEWS] What's the goal of views?
[SERVICES] Create a new service, assign a service protocol and add one element with all parameters.

EmmaSaldi (commented):
Looks good to me.

BertVandenberghe (approved):
Perfect, thanks!

PR #3759, opened 2024-10-02
Add Tutorial for Kata#43: Grouping alarms like a pro

SkylineDennisDr:
Kata number 43: Grouping alarms like a pro was recorded on Tuesday 01/10/24. This commit contains the accompanying tutorial.


SkylineDennisDr:
Thanks for the changes and rephrasing various things. It's a lot nicer this way.

I had a few small remarks:
1) In step 5 of the tutorial, there is an xml code snippet. Above the snippet we mention 3 steps, but under the snippet, the counting starts from 1 again. Should the count continue at 4 here?
2) In step 6, the final exercise, I noticed from the 6-step procedure that I was a bit unclear on the procedure. For users having access to the configuration.xml file, they can just change that xml file, save it and restart SLAnalytics. For people not having access to this file, they should modify the KataExercise string in the automation script and run that script. For them, it is not (!) needed to save the xml and restart SLAnalytics as the script will do this for them.
3) On the new "Customizing alarm grouping rules", you added 3 sections: on service based grouping, grouping based on relation learning and grouping based on location. The other grouping rules don't really need to be mentioned because their configuration works similarly to one of the mentioned 3. Maybe we should mention this in the text or else it might look weird that we're zooming in on 3 rules and not on the others? Or am I seeing ghosts?

MariekeGO:
> Thanks for the changes and rephrasing various things. It's a lot nicer this way.
> 
> I had a few small remarks:
> 
> 1. In step 5 of the tutorial, there is an xml code snippet. Above the snippet we mention 3 steps, but under the snippet, the counting starts from 1 again. Should the count continue at 4 here?
> 2. In step 6, the final exercise, I noticed from the 6-step procedure that I was a bit unclear on the procedure. For users having access to the configuration.xml file, they can just change that xml file, save it and restart SLAnalytics. For people not having access to this file, they should modify the KataExercise string in the automation script and run that script. For them, it is not (!) needed to save the xml and restart SLAnalytics as the script will do this for them.
> 3. On the new "Customizing alarm grouping rules", you added 3 sections: on service based grouping, grouping based on relation learning and grouping based on location. The other grouping rules don't really need to be mentioned because their configuration works similarly to one of the mentioned 3. Maybe we should mention this in the text or else it might look weird that we're zooming in on 3 rules and not on the others? Or am I seeing ghosts?

I'm not done with the review yet, actually 😊. I'll take your comments into account when I continue (need to do some other things first now).

user-guide/Basic_Functionality/Alarms/Working_with_alarms/Advanced_analytics_features/Tutorials/Incident_Tracking_Configuration_Tutorial.md
```diff
@@ -0,0 +1,248 @@
...
+   > [!IMPORTANT]
+   > Use the following email format to send us your screenshot:
+   >
+   > - Subject: Tutorial - Fine-tune incident tracking in your system
+   > - To: [ai@skyline.be](mailto:ai@skyline.be)
+   > - Body:
+   >   - Dojo account: Clearly mention the email address you use to sign into your Dojo account, especially if you are using a different email address to send this email.
+   >   - Attachment: Add a screenshot of the amplifiers you have set in alarm and of the corresponding alarm group.
+   >
+   > Upon successful validation, you will be awarded 75 DevOps Points.
```
MariekeGO, line 242: Could you adjust this step so that it's more clear what exactly the user should do? Right now, you're kind of dropping them into the deep end and having them figure things out themselves, but this isn't really what we want with these tutorials. As a user following a tutorial, I wouldn't want to be forced to go read another section of the help to be able to continue; instead I'd want to have the basic steps, with a link to another section in case I want to dive deeper (the way I did it for step 5). Could you also clarify why it's necessary to add another custom element property, instead of using the *parentNode* property from the previous step?

user-guide/Basic_Functionality/Alarms/Working_with_alarms/Advanced_analytics_features/Tutorials/Incident_Tracking_Configuration_Tutorial.md
```diff
@@ -0,0 +1,313 @@
...
+
+1. Under *Automatic Incident Tracking*, check which ID is mentioned in the box *Leader DataMiner ID*.
+
+   This is the ID of the DataMiner Agent within your DMS that is in charge of running the incident tracking algorithm for the entire DMS.
+
+1. On the server running the leader DMA, go to the folder `C:\Skyline DataMiner\Analytics`.
+
+   Within this folder, you will find the *configuration.xml* file.
+
+1. Manually take a copy of your current *configuration.xml* file so that you can put it back when you finish this tutorial.
```
MariekeGO, line 63: Why should users take this manual copy? In step 7, it says a script will revert the configuration and the manual copy is not mentioned. Is this intended as a backup?

SkylineDennisDr: The catalogue package contains 4 scripts: the last one being the reset script. Whenever you run script 1, 2 or 3 it creates a backup of your configuration.xml file if such a backup doesn't already exist. This backup is then used by script 4 to revert to the original setup. So, if you always use the automation scripts and don't do any manual changes to the configuration.xml, you don't need to worry about taking backups.

However, if you make any manual changes to the configuration.xml file before running any of the scripts, then the original configuration as you had it is gone (ie there is no backup of the original setup). Script 4 will then be unable to revert everything...

MariekeGO: I've updated the text to make sure this is clear. Could you take a look to check if everything is still OK for you?

SkylineDennisDr:
Looks good! One small remark is maybe that at the end (step 7), we mention two ways of resetting everything: manually or using the script. Maybe we could mention that "using the script" is an alternative method for resetting everything in case you didn't do any manual changes. Now it sounds a little bit like you should not manually restore the configuration.xml in case you didn't do any manual changes, despite the fact that it is also perfectly possible to do so.

No further comments apart from that :) Looks nice!

MariekeGO (approved):

PR #3767, opened 2024-10-07
Add Flows documentation

Robin-Debel:


MariekeGO (commented):
Could you move your images to the _user-guide/images/_ folder and update your image syntax so it refers to that folder instead of the _develop/images/_ folder?
In addition, could you also add DataMiner version information so that, for users who are using a different DataMiner version, it is clear why they don't see the options you describe and which version they need to upgrade to in order to see them.
For now, I will revert this pull request to a draft. When it is fully ready, could you mark it as "Ready for review"?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Dashboards_app/Tutorials/Controls_Feeds_Query.md
MariekeGO, line 1: As long as the disclaimer at the top of the tutorial mentions that the tutorial content matches and earlier version (e.g. 10.4.1 in this case), we need to make sure that the text indeed applies for people using that version (though we can add info about other versions as a nice-to-know). I've now adjusted the PR so this is OK again. Alternatively, we could also change the disclaimer at the top to a more recent version, but then we'll need to make sure all screenshots are updated to that version as well.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Flows/Configuring_a_flow.md
MariekeGO, line 1: I've adjusted the procedure on this page to be more suitable for a regular user guide procedure, because the original style was more like a tutorial.

MariekeGO (commented):
@Robin-Debel, I've reviewed your PR and made some changes. Could you take a look and confirm if everything still looks OK for you?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Flows/Configuring_a_flow.md
Robin-Debel: Looks good to me

MariekeGO (approved):

PR #3781, opened 2024-10-09
Update Accessing_DMS_remotely_with_Cube.md

bertbsky:
ConnectionSettings.txt is not accessed by the remote Cube, this is only used for direct access to the agent. Also see RN35779

### DIFF
--- a/user-guide/Cloud_Platform/RemoteAccess/Accessing_DMS_remotely_with_Cube.md
+++ b/user-guide/Cloud_Platform/RemoteAccess/Accessing_DMS_remotely_with_Cube.md
@@ -10,7 +10,6 @@ From DataMiner 10.3.0/10.3.2 onwards, you can use DataMiner Cube to access your 
 
 - The DataMiner System uses DataMiner 10.3.0/10.3.2 or higher and is [connected to dataminer.services](xref:Connecting_your_DataMiner_System_to_the_cloud).
 - The DataMiner System has been [configured to use HTTPS](xref:Setting_up_HTTPS_on_a_DMA).
-- The DataMiner Cube connection settings have been configured to use gRPC (see [ConnectionSettings.txt](xref:ConnectionSettings_txt)).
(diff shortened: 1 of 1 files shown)


user-guide/Cloud_Platform/RemoteAccess/Accessing_DMS_remotely_with_Cube.md
```diff
@@ -10,7 +10,6 @@ From DataMiner 10.3.0/10.3.2 onwards, you can use DataMiner Cube to access your
 
 - The DataMiner System uses DataMiner 10.3.0/10.3.2 or higher and is [connected to dataminer.services](xref:Connecting_your_DataMiner_System_to_the_cloud).
 - The DataMiner System has been [configured to use HTTPS](xref:Setting_up_HTTPS_on_a_DMA).
-- The DataMiner Cube connection settings have been configured to use gRPC (see [ConnectionSettings.txt](xref:ConnectionSettings_txt)).
```
MariekeGO, line 13: As RN 35779 is introduced in 10.3.0 [CU2]/10.3.5, does this mean that for earlier DataMiner versions the line you removed does still apply?

bertbsky: Yes, the ConnectionSettings.txt requirement would still apply to 10.3.0 CU0-CU1 and 10.3.2-10.3.4 versions.

MariekeGO: In that case, instead of removing this line, I'd propose modifying it with the necessary version info. After all, the user guide must be valid for users of all currently supported DataMiner versions, so users of these older versions also need to find accurate information for their setups. Something like this:

```
- Prior to DataMiner 10.3.0 [CU2] and 10.3.5, the DataMiner Cube connection settings must be configured to use gRPC (see [ConnectionSettings.txt](xref:ConnectionSettings_txt)). This is no longer required in later DataMiner versions.
```

MariekeGO (approved):

PR #3783, opened 2024-10-10
Update WinDbg.md

PedroDebevere:


develop/TOOLS/ThirdPartyTools/WinDbg.md
```diff
@@ -8,18 +8,123 @@ uid: WinDbg
 
 ![WinDbg](~/develop/images/WinDbg.png)<br>*WinDbg*
 
-## Useful commands
+## Commands
 
-The following table contains an overview of some useful commands for investigating a dump.
+The following table contains an overview of some useful meta-commands.
```
MariekeGO, line 13: I suppose this should be "commands" rather than "meta-commands"? I'm also not sure if we should speak of "some useful commands" if there's only one listed – unless we're planning to add more commands soon, it might be better to rephrase this (same below).

PedroDebevere: correct, should be commands. It's indeed the intention that more commands will get added to these tables.

MariekeGO (approved):

PR #3790, opened 2024-10-11
RN 40658

EdithLansens:
@wmelangeSkyline, could you please review this pull request to confirm the accurate inclusion of release note 40658 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
+++ b/user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
@@ -4,9 +4,9 @@ uid: Adding_time_scoped_related_parameters_to_a_trend_graph
 
 # Adding time-scoped related parameters to a trend graph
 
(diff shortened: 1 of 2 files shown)


wmelangeSkyline (approved):

user-guide/Basic_Functionality/Trending/Advanced_analytics_trending/Relation_learning/Adding_time_scoped_related_parameters_to_a_trend_graph.md
```diff
@@ -4,9 +4,9 @@ uid: Adding_time_scoped_related_parameters_to_a_trend_graph
...
 
-You can for instance use this in case a parameter (e.g. the total available memory of a server) behaves oddly during a particular time range (e.g. a downward spike), in order to find out if other parameters of the same device also showed unusual behavior during the same time range.
+You can for instance use this in case a parameter (e.g. the total available memory of a server) behaves oddly during a particular time range (e.g. a downward spike), in order to find out if other parameters, either from the same device or from devices within the same service, also showed unusual behavior during the same time range.
```
MariekeGO, line 9: Maybe we should change "device" to "element" here, so we don't exclude elements that don't monitor an actual physical device.
```suggestion
You can for instance use this in case a parameter (e.g. the total available memory of a server) behaves oddly during a particular time range (e.g. a downward spike), in order to find out if other parameters, either from the same element or from elements within the same service, also showed unusual behavior during the same time range.
```

MariekeGO (approved):

PR #3792, opened 2024-10-11
Update Installing_the_Keysight_Agilent_IO_Libraries.md

SkylineThomasCR:
Keysight installation requires a restart of the machine to take full effect. Otherwise, SLPort won't find the dll even though the file is present in system32.

Add an alternative solution to fix the Keysight and NATS port problem.


MariekeGO (commented):
I've adjusted your PR so that the server reboot and DMA restart are included in the relevant procedures, to ensure that users who skip the text at the top don't forget these steps. Could you check if everything still looks OK for you?

SkylineThomasCR:
Thanks for the great improvement. Everything is still correct.

MariekeGO (approved):

PR #3795, opened 2024-10-11
RN 40372

EdithLansens:
@RobbeDGH, could you please review this pull request to confirm the accurate inclusion of release note 40372 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Query_Data_Sources/Get_alarms.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Query_Data_Sources/Get_alarms.md
@@ -4,4 +4,14 @@ uid: Get_alarms
 
 # Get alarms
 
-Available from DataMiner 10.2.0/10.1.9 onwards. The *Get alarms* data source retrieves the alarms in the DataMiner System. Several columns, such as *Element Name*, *Parameter Description*, *Value* and *Time*, are included by default. Others can be added with a [*Select* operation](xref:GQI_Select).
+The *Get alarms* data source retrieves the alarms in the DataMiner System. Several columns, such as *Element Name*, *Parameter Description*, *Value* and *Time*, are included by default. Other columns, such as *Alarm ID* and *Root Alarm ID* can be added with a [*Select* operation](xref:GQI_Select).
+
+The content of the *Alarm ID* and *Root Alarm ID* columns may differ depending on your DataMiner version:
+
+| Column | From DataMiner 10.4.11/10.5.0 onwards<!--RN 40372--> | Prior to DataMiner 10.4.11/10.5.0 |
+|--|--|--|
+| Alarm ID | DMAID/EID/RootAlarmID/AlarmID, e.g. `537/45/2184182/2184188` | HostingDMAID/AlarmID, e.g. `537/2184188` |
+| Root Alarm ID | DMAID/EID/RootAlarmID, e.g. `537/45/2184182` | HostingDMAID/RootAlarmID, e.g. `537/2184182` |
+
+> [!NOTE]
+> "DMAID" is the DataMiner ID of the DataMiner Agent on which the alarm was generated.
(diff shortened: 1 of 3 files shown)


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/Query_Data_Sources/Get_alarms.md
```diff
@@ -4,4 +4,14 @@ uid: Get_alarms
...
+
+The content of the *Alarm ID* and *Root Alarm ID* columns may differ depending on your DataMiner version:
+
+| Column | From DataMiner 10.4.11/10.5.0 onwards<!--RN 40372--> | Prior to DataMiner 10.4.11/10.5.0 |
+|--|--|--|
+| Alarm ID | DMAID/EID/RootAlarmID/AlarmID, e.g. `537/45/2184182/2184188` | HostingDMAID/AlarmID, e.g. `537/2184188` |
+| Root Alarm ID | DMAID/EID/RootAlarmID, e.g. `537/45/2184182` | HostingDMAID/RootAlarmID, e.g. `537/2184182` |
+
+> [!NOTE]
+> "DMAID" is the DataMiner ID of the DataMiner Agent on which the alarm was generated.
```
RobbeDGH, line 17: This is wrong. 

DMAID is the dataminer id of the agent that created the element.
HostingDMAID is the dataminer id of the agent that is hosting the element at the time of alarm creation.

Maybe its best to add both explanations in the note for those unfamiliar with the difference (it is tricky!)

Most of the time these values are the same, except in 2 cases:
An element is exported from one agent and imported on another.
An element is swarmed from one agent to another.

EdithLansens:
```suggestion
> "DMAID" refers to the DataMiner ID of the DataMiner Agent where the alarm was created. "HostingDMAID" refers to the DataMiner ID of the DataMiner Agent hosting the element. Most of the time, these two values will be the same, but occasionally, they may differ, e.g. when an element is exported from one Agent and imported onto another Agent.
```

Is this more accurate, @RobbeDGH? I haven't mentioned swarming as this isn't out yet.

RobbeDGH: No this is still not _always_ true.

Either we have to put

"HostingDMAID" refers to the DataMiner ID of the DataMiner Agent where the alarm was created. 

or

"DMAID" refers to the DataMiner ID of the DataMiner Agent where the **element** was created. 

or both of course.

e.g. imagine following scenario

DMA with id 15 creates an element => the dmaid of the element is 15 and gets a fresh elementid, lets say 145. The full id is now 15/145.
The DMAID here is 15.
If this element creates an alarm it gets hostingdmaid 15 as well => eg lets say 15/789456

When the element is exported and imported on another agent, lets say agent with id 21 and then creates an alarm, the alarm gets ids 21/56895623. (just picked random alarmid)
So in this case the hosting id is 21 while dmaid is 15.

Is this more clear? Let me know if I explained it clearly!
Otherwise I can give you a live example on my DMA with the differences

EdithLansens: Thanks for the explanation! Is this better?

```suggestion
> "DMAID" refers to the DataMiner ID of the DataMiner Agent where the element was originally created. "HostingDMAID" refers to the DataMiner ID of the DataMiner Agent currently hosting the element and managing its alarms. Most of the time, these two values will be the same, but they may differ, for example, when an element is exported from one Agent and imported onto another Agent. In this case, the element retains the original DMAID, but the HostingDMAID will reflect the new Agent's ID.
```

RobbeDGH: Yes this is much better.

Can we add this suggest change? I don't see a button.
Should I push this update?

EdithLansens: Great, I'll commit the suggestion!

PR #3823, opened 2024-10-17
DCP Changelog 17-10-2024

SkylineKevinMOR:


release-notes/Cloud/DCP_change_log.md
```diff
@@ -9,6 +9,10 @@ The dataminer.services platform gets updated continuously. This change log can h
 > [!NOTE]
 > Many features on dataminer.services are dependent on DxMs. You can find the change logs for these under [DxM release notes](xref:DxM_RNs_index).
 
+#### 17 October 2024 - Fix - Catalog - Updated heading styles [ID 41145]
+
+Headings in the Catalog will now be rendered correctly.
```
MariekeGO, line 14: Could you adjust this description so it describes what the issue was? This way, it will be consistent with the way we usually write RNs for fixes, and users who have encountered the issue will be able to recognize it more easily.

SkylineKevinMOR: Wasn't really an issue, we mainly just adjusted the font to the correct font and improved the tickness a bit so the heading weren't as thin. Not sure if it should be an enhancement instead.

MariekeGO: If it wasn't actually an issue, could you adjust this to "Enhancement" and also rephrase the description so it doesn't seem to refer to an issue?

MariekeGO (approved):

PR #3854, opened 2024-10-30
Update Trending_techniques.md

BrentSLC:
Small addition on the difference in avg-values between numeric and non-numeric values


user-guide/Basic_Functionality/Trending/Trending_techniques.md
```diff
@@ -28,4 +28,4 @@ Logging of average values only, usually across longer time spans than those used
 - Last year, with 1-hour averages (max. about 8760 data points).
 
 > [!NOTE]
-> In case of average trending, the DMS not only stores the average value for a particular time slot, but also the minimum and maximum value detected in that same time slot.
+> In case of average trending, the DMS not only stores the average value for a particular time slot, for numeric values it also stores the minimun and maximum value detected in that same time slot. In case of non-numeric parameters (such as [Discrete, analog, and hybrid parameters](xref:https://docs.dataminer.services/user-guide/Basic_Functionality/Parameters/Types_of_parameters/Discrete_analog_and_hybrid_parameters.html)) the DMS stores the value that occured the most, the value that was active the longest, and the percentage of time this value was active during the timeslot.
```
MariekeGO, line 31: To create a cross-reference within the documentation, you should use the UID instead of the URL: 
```suggestion
> In case of average trending, the DMS not only stores the average value for a particular time slot. For numeric values it also stores the minimum and maximum value detected in that same time slot. For non-numeric parameters (such as [discrete and hybrid parameters](xref:Discrete_analog_and_hybrid_parameters)) the DMS stores the value that occurred the most, the value that was active the longest, and the percentage of time this value was active during the time slot.
```

MariekeGO (approved):

PR #3860, opened 2024-10-30
Connector documentation update

MariekeGO:


develop/devguide/Connector/ConnectorHelp/Connector_help_pages.md
```diff
@@ -59,13 +59,9 @@ In the **first paragraph** below the title, add a short paragraph explaining the
...
 
-In the **About** section, you will need to add a number of tables, depending on the connector. You can find these tables in the [template](xref:Connector_help_template).
+**OBSOLETE**
 
-- For all connectors except exported child connectors, add the **Version Info** table.
-
-- For all connectors except virtual connectors, add the **Product Info** table.
-
-- For **all connectors**, add the **System Info** table. For exported connectors or connectors that cannot have exported connectors, you can leave out the *Exported Components* column in this table.
+Older connector documentation will contain an *About* section with tables with version info, product info, and system info, but this is no longer required.
```
SimonV-Skyline, line 64: Shouldn't we add the reasonining why it is no longer required?
the reasoning is that all that info should now be included directly within the protocol xml via the various VersionHistory tags
image

SimonV-Skyline: BTW, such info from those tag is going to be read by our pipelines which will generate a summary of it and push it to the Catalog.

MariekeGO: Shall I add this with a cross-reference like this?
```suggestion
Older connector documentation will contain an *About* section with tables with version info, product info, and system info, but this is no longer required as this information is now included directly within the protocol.xml via the various [VersionHistory](xref:Protocol.VersionHistory) tags.
```

PR #3863, opened 2024-10-30
deprecate unattended installer

juliori85:
this feature is no longer supported

### DIFF
--- a/user-guide/Getting_started/Installing_a_DMA/Unattended_DM_installation.md
+++ b/user-guide/Getting_started/Installing_a_DMA/Unattended_DM_installation.md
@@ -4,6 +4,9 @@ uid: Unattended_DM_installation
 
 # Unattended DataMiner installation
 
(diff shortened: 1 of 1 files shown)


user-guide/Getting_started/Installing_a_DMA/Unattended_DM_installation.md
```diff
@@ -4,6 +4,9 @@ uid: Unattended_DM_installation
 
 # Unattended DataMiner installation
 
+> [!IMPORTANT]
+> (Deprecated) This feature is not supported.
```
MariekeGO, line 8: @juliori85, I assume unattended installation is only possible with the older, deprecated installers? In that case, to make this more clear, I'd rephrase this as follows:
 ```suggestion
> This feature is deprecated. It is no longer available starting from version 10.4 of the DataMiner Installer.
```

MariekeGO (approved):

PR #3865, opened 2024-10-31
added info on suppressing and postponing validator results

PedroDebevere:


develop/TOOLS/DIS/ToolWindows/DisValidatorToolWindow.md
```diff
@@ -6,14 +6,60 @@ uid: DisValidatorToolWindow
...
+- Code: Displays the unique code of the validator check.
+- Line: Specifies the line number where the detected issue.
+- Column: Specifies the column number where the issue was detected.
+- DVE: Displays the name of the DVE protocol in case this validation result belongs to a DVE protocol.
 
 ![DIS Validator tool window](~/develop/images/DisValidatorToolWindow.png)
+
+## Suppressing or postponing a validation result
+
+In case a validation result is not valid, it can be suppressed. To suppress a validation result, select the item from the list, right click to open the context menu and select *Suppress...* from the context menu. In the *Suppress Validator Result* window that appears, provide a reason why the result has been suppressed.
```
SimonV-Skyline, line 38: It would be great if we could here explain that we allow suppression on any result for flexibility but that, in theory, only a validation result defined as "Uncertain" should be suppressed. Indeed such "Uncertain" checks require a human to double check them as the automated validation can't be sure. If one would need to suppress a "Certain" result, it probalby means the check is buggy and either should be improved, either should be defined as "Uncertain" and in that case, the proper way to go is to create a validator issue task and postpone it.
The only other (valid? debatable?) use-case where you would suppress a valid and "Certain" remarks is if you consider it's not important enough.

develop/TOOLS/DIS/ToolWindows/DisValidatorToolWindow.md
```diff
@@ -6,14 +6,60 @@ uid: DisValidatorToolWindow
...
+> As of DIS v2.41, when you open a main DVE protocol and click *Validate*, DIS will not only validate the main protocol but also all its exported child DVE protocols.
+
+The validator results are listed in a treegrid which groups the validation results by their severity. The treegrid has the following columns:
 
-As of DIS v2.41, when you open a main DVE protocol and click *Validate*, DIS will not only validate the main protocol but also all its exported child DVE protocols.
+- Description: Description: Description of the issue
+- Information icon: This icon is shown in case more information is available about this validator check. Clicking the information icon will open a pop-up window which provided more details about the validator check.
+- Autofix icon: If this icon is displayed, an auto-fix is available. To execute the fix, from the context menu, select either *Fix* > *This error* or *Fix* > *All errors of this type*
+- State: Shows the state of this validation result (Active, Postponed or Suppressed). 
+- Certainty: Specifies whether the validator is certain or uncertain about the validity of the detected issue.
```
SimonV-Skyline, line 26: Could be good here to refer to the fact that "Uncertain" results should be double checked by a developer and either fixed, either suppressed (with a link to the section about suppression lower on this page)

develop/TOOLS/DIS/ToolWindows/DisValidatorToolWindow.md
MariekeGO: Can external users also create a validator issue task, or should they use the "Send feedback" feature instead?

SimonV-Skyline: DIS now has a "Report Issue..." that brings end-users towards the create UOSA ticket web page.
Personally, I hate it because on there, you then have to select which project, etc which is not very user-friendly to me. But yeah, I gave that feedback to Platform and they told me no improvements were going to be made there and that was still the official way to go :)

MariekeGO: Right now, we only mention the Send Feedback option in the documentation, so I'll have to add this as well. How could we distinguish between the two options? Perhaps we can say that if you don't know which project to select, you can use Send Feedback instead?

SimonV-Skyline: Maybe a bit vauge but I would say use "Report Issue..." for reporting issues and use "Send Feedback..." for anything else :) @PedroDebevere, what do you think?

PedroDebevere: Note that the Report issue menu item is currently only available for SLC users only. For other uses only the send feedback menu item is shown. We can add this as an additional note that the report issue is currently not available for other users and that they should use the Send Feedback functionality.

SimonV-Skyline (approved):
all good to me now

MariekeGO (approved):

PR #3885, opened 2024-11-05
update installer docs

juliori85:


user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
```diff
@@ -49,6 +49,9 @@ If you are using DataMiner Installer v10.4, follow the steps below to install Da
 
    Once the installation is complete, the configuration window will be displayed.
 
+   > [!IMPORTANT]
+   > If you intend to restore a backup coming from another machine because of e.g. a hardware migration or during disaster recovery, skip the configuration below and follow the steps under [Restoring a DataMiner Agent](xref:Restoring_a_DataMiner_Agent).
```
MariekeGO, line 53: I think we'll need to be more specific here, because this creates a loop reference: here we refer the user to restoring a DataMiner Agent, but as soon as they get to the first page there (preparing the destination server), they are referred back to this page.

MariekeGO (approved):

PR #3900, opened 2024-11-08
RN 41034

EdithLansens:
@WoutMH, could you please review this pull request to confirm the accurate inclusion of release note 41034 in the user guide? Thanks in advance!


user-guide/images/Version_History.png
WoutMH, line 1: I see that the colored circles are not centered in the screenshot. Did you zoom in while taking it? On some zoom levels the browser has to do some rounding which can cause this. Zooming in slightly more/less can fix this

EdithLansens: @WoutMH, I double-checked, and I didn't zoom in while taking the screenshot—it was set to 100%. I tried zooming out to 80%, and now the circles appear more centered.

WoutMH: The image is indeed correct now. Zooming isn't the only thing that can produce these rounding errors, just the most frequently occurring in my experience. It is also related to: screen size, window size, browser, browser version, browser extensions, ... which are all out of control of our software.

MariekeGO (approved):

PR #3903, opened 2024-11-08
RN 41235 - GetNext note

spetebroodt:


LaurensMT (changes requested):
In the Capture observation there is mentioned "Additional get next requests are performed until the OID in the response exceeds the table OID range." Before RN 41235 this was correct, but since this RN it will perform get next requests until the values of the defined column parameters are retrieved, or until the table OID range is exceeded in case there were more columns defined than there are column values.

spetebroodt:
@LaurensMT I rewrote the capture observation. Is this correct?

LaurensMT (approved):

PR #3920, opened 2024-11-13
Update Indexing_Database.md

AlexanderGorbunovSkyline:
Somehow it was not mentioned anywhere in the Docs that Indexing database is used to store alarm data. Hereby I'd like to fix that.

### DIFF
--- a/user-guide/Advanced_Functionality/Databases/Indexing_database/Indexing_Database.md
+++ b/user-guide/Advanced_Functionality/Databases/Indexing_database/Indexing_Database.md
@@ -20,6 +20,8 @@ An indexing database is required for many DataMiner features, including:
 
 - [Specific GQI data sources](xref:Query_data_sources)
 
+- [Active and historical alarms and information events](xref:About_alarms) (required for [Dedicated clustered storage](xref:Configuring_dedicated_clustered_storage), optional for [Separate Cassandra storage](xref:Separate_Cassandra_setup_without_Elasticsearch))
+
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Databases/Indexing_database/Indexing_Database.md
```diff
@@ -20,6 +20,8 @@ An indexing database is required for many DataMiner features, including:
 
 - [Specific GQI data sources](xref:Query_data_sources)
 
+- [Active and historical alarms and information events](xref:About_alarms) (required for [Dedicated clustered storage](xref:Configuring_dedicated_clustered_storage), optional for [Separate Cassandra storage](xref:Separate_Cassandra_setup_without_Elasticsearch))
```
MariekeGO, line 23: I'm not sure if this is the right place to put this, as this list enumerates the features that are only available with indexing, but active and historical alarms and information events are also available without - they just get stored elsewhere then. I think that to add the info on what is stored in the indexing database, we should update the [DataMiner data persistence](https://docs.dataminer.services/develop/devguide/Connector/AdvancedDataMinerDataPersistence.html) section, because this is now far from up to date. It includes only the [database structure](https://docs.dataminer.services/develop/devguide/Connector/AdvancedDataMinerDataPersistenceNoSqlCassandra.html#database-structure) for legacy setups with a Cassandra  DB per DMA only. However, as users are not supposed to interact with the database directly, updating this section currently has very low priority. Until we can update this fully, perhaps we could add this info as a note under "[Information events](https://docs.dataminer.services/develop/devguide/Connector/AdvancedDataMinerDataPersistenceNoSqlCassandra.html#information-events)" there?

MariekeGO (approved):

PR #4003, opened 2024-11-29
RN 41328

EdithLansens:
@Tom-Van-Bruaene, could you please review this pull request to confirm the accurate inclusion of release note 41328 in the user guide? Thanks in advance!

I've largely copied information added for [RN 40372](https://github.com/SkylineCommunications/dataminer-docs/pull/3795) and [RN 41113](https://github.com/SkylineCommunications/dataminer-docs/pull/3899) as this felt like a very similar change. However, if I've included too much or incorrect information because of this assumption, let me know.

### DIFF
--- a/user-guide/Basic_Functionality/Alarms/Working_with_alarms/Working_with_the_Alarm_Console/Alarm_Console_Columns.md
+++ b/user-guide/Basic_Functionality/Alarms/Working_with_alarms/Working_with_the_Alarm_Console/Alarm_Console_Columns.md
@@ -110,6 +110,14 @@ For more information on changing the column layout, see [Changing the column lay
 
 - **Alarm ID**: The unique ID that identifies the alarm record. This ID is for instance used to link alarms to trouble tickets.
 
+  > [!NOTE]
+  > For SLAs, the content of the *Alarm ID* column may differ depending on your DataMiner version:
+  >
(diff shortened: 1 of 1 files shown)


user-guide/Basic_Functionality/Alarms/Working_with_alarms/Working_with_the_Alarm_Console/Alarm_Console_Columns.md
MariekeGO, line 1: @wdm-slc, to my knowledge the Alarm ID will be a different format for most objects by now. Possibly we do not need to detail this here. What do you think?

wdm-slc: Cube alarm console always shows the full ids nowadays. There's no distinction between alarms from SLAs or from other elements.

These changes were made during the 10.4.x feature release cycles as part of a series of updates that made Cube compatible with the full ids (I could find that alarmid was updated in RN38815, rootalarmid will have been around the same time).

AlarmID: dmaid/eid/rootalarmid/alarmid
RootAlarmID: dmaid/eid/rootalarmid

I also believe that these latest Cube versions show these types of references even when connected to older systems.

PR #4015, opened 2024-12-03
update: redundant polling page with descriptions of how Dataminer switches between interfaces (SNMP)

ChianHA:
Based on locally simulated tests with 2 SNMP simulation files running on 2 interfaces, the findings of Dataminer's redundant polling feature's behaviour has been recorded in the page.

This PR covers the following cases for SNMP:
- On timeout of the currently polling interface
- On polling invalid SNMP parameters (**This is a corner case that usually won't be encountered if parameters are verified correctly. I discovered it while testing so I thought it would be good to note it down somewhere at least. Pending review if it should be included in the docs.**)


develop/devguide/Connector/ConnectionsRedundantPolling.md
```diff
@@ -15,10 +15,26 @@ In the protocol, you need to define the 2 connections and specify that redundant
 The option "redundantPolling" in the communicationOptions attribute is required to activate the redundant polling mechanism.
 
 > [!NOTE]
-> Redundant polling is supported with protocols that define exactly two connections of the same type. The following is supported: two SNMP, two serial, two smart-serial or two HTTP connections.
+> Redundant polling is supported with protocols that define exactly two connections of the same type. The following is supported: two SNMP, two serial, two smart-serial, or two HTTP connections.
 >
 > Note that for smart-serial connections, in order to get a timeout (enabling the redundant polling connections to switch) the pair must contain a response. Alternatively, you can implement logic to set the communication state using the <xref:NT_CHANGE_COMMUNICATION_STATE> Notify call.
 
+## Switching implementation
```
MariekeGO, line 22: Are the cases described in this section the only cases when switching will ever be implemented, or could there be others?

SimonV-Skyline (commented):
State more explicitly that: 
Currently, the element timeout time doesn't play a role in the switching behaviour. A single group timing out after retries is enough to trigger the switch.

Also, can you add conclusion to following tests
- if a group on which the connection to be used is explicitly define in connector will influence or not the switch?
- if a parameter where the IP/port to be used is defined via the ipid attribute will influence the switch or not?
- if a parameter is polled using 1 and the other connection from QAction will influence the switch or not?

MariekeGO (commented):
I've done a basic review of this PR already, but I think it's still a little confusing as it is.
The current "Switching implementation" heading suggest that the section below it contains info on all the different ways you can implement this. But then the information of the section seems more of a description of some of the situations where the switching can occur. And then the "Other polling scenarios" section also contains a situation where the switching can occur, which makes the structure all the more confusing. All in all, it's not quite clear what we want to achieve with the new sections. Do we want to inform users of how to implement this feature? Then we should describe the implementation in more detail. Do we want to give examples of the implementation? Then we should make it clear that these are some examples of when switching does or does not occur.

PR #4032, opened 2024-12-05
Update to match new version (range) states & semver strategy

SLC-brechtseys:
Semantic versioning is not longer enforced (but still recommended). Some version (range) states got changed.


MariekeGO (approved):

SLC-brechtseys (commented):
ok

PR #4033, opened 2024-12-05
Update ConnectionsSnmpAlteringAVariable.md

tamaramarkesic:
Enumeration typo fixed


MariekeGO:
Numbering happens automatically – it doesn't matter if the number is "1." or "2." – and on the whole we prefer using "1." everywhere because it's lower-maintenance. However, to get a continuous numbered list, the code examples in the list must be indented, and this is what went wrong here.

MariekeGO (approved):

PR #4059, opened 2024-12-12
New solution section on front page

MariekeGO:


EdithLansens (commented):
@MariekeGO, I’d avoid capitalizing "solutions" or "framework" since they’re not specific to DataMiner, so it doesn’t feel necessary. For example, we don’t capitalize "modules" in "DataMiner modules," which seems like a similar kind of term.

image

Apart from these small suggestions, everything looks great!

PR #4060, opened 2024-12-12
Update Creating_a_DaaS_system_FAQ.md

baptistepattyn:


MariekeGO:
@baptistepattyn, I've updated your PR with references to the daas@dataminer.services address to be more specific as to where people can get assistance. Could you check if everything is still OK for you?

MariekeGO (approved):

PR #4074, opened 2024-12-18
Added table with item types & descriptions

SLC-brechtseys:
Re-added the item types & descriptions. Some will be added later.


user-guide/Cloud_Platform/Catalog/About_the_Catalog_module.md
```diff
@@ -25,6 +25,23 @@ There are several ways to access the Catalog module:
...
+| Automation | Contains a general-purpose [DataMiner Automation script](xref:automation) to execute automation routines, triggered either on demand by users, based on a time schedule, or based on specific events that occur. Deploying the component will make it available on the target DataMiner System, where it can be used as is, or where you can further adjust it to meet your own specific needs. |
+| User-defined API | Contains a [user-defined API](xref:UD_APIs), which adds a new endpoint to your DataMiner System tailored to a specific use case, and which enables third-party applications to consume selective data from your DataMiner System. Deploying the component will make it available on the target DataMiner System, where it can be used as is, or where you can further adjust it to meet your own specific needs. You will then have to make that new user-defined API accessible by generating and sharing a token with the intended party. |
```
MariekeGO, line 39: I see that we use title casing for all the types except for this one. Wouldn't it be better to also use it here for the sake of consistency?
```suggestion
| User-Defined API | Contains a [user-defined API](xref:UD_APIs), which adds a new endpoint to your DataMiner System tailored to a specific use case, and which enables third-party applications to consume selective data from your DataMiner System. Deploying the component will make it available on the target DataMiner System, where it can be used as is, or where you can further adjust it to meet your own specific needs. You will then have to make that new user-defined API accessible by generating and sharing a token with the intended party. |
```

MariekeGO (approved):

PR #4116, opened 2025-01-07
Add information about the DxM logs in DataMiner Cube

AndreRodriguesSkyline:


user-guide/Troubleshooting/Logging/Consulting_the_DataMiner_logs/DataMiner_logging.md
```diff
@@ -101,3 +103,9 @@ Some items in the list are of particular note:
 >
 > - For each of these log files, a corresponding TXT file is located in the `C:\Skyline DataMiner\Logging\` folder. The name of the TXT file is often the same as the name mentioned in Cube, but with an "SL" prefix. However, this is not always the case. If you select a log file in Cube, the corresponding TXT file name will be displayed at the top.
 > - Some of the mentioned log files refer to advanced DataMiner modules that are not part of the DataMiner system by default.
+
+## DataMiner Extension Modules (DxM) logs
+
+The DataMiner Extension Modules logs are identified by their module name with the suffix (DxM).
+
+To adjust these logs, manually edit the *appsettings.json* or *appsettings.custom.json* file located in the installation directory of the respective DxM.
```
MariekeGO, line 111: I think a bit more info is needed here about when to use which of these two files. Based on what is written elsewhere in the documentation, I get the impression that any changes to *appsettings.json* get overwritten when there is an update, but I'm not sure if this is always the case, or if there is other information that might also need to be added here.

AndreRodriguesSkyline: Thanks for the remarks! You're right. It's better to recommend using an appsettings.custom.json to prevent the settings from being overwritten.

MariekeGO (approved):

PR #4118, opened 2025-01-08
RN 41495

EdithLansens:
@AaronVanVyve, could you please review this pull request to confirm the accurate inclusion of release note 41495 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Low-Code_Apps_configuration/LowCodeApps_event_config.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Low-Code_Apps_configuration/LowCodeApps_event_config.md
@@ -45,7 +45,9 @@ To configure an event to launch a script:
 1. Optionally, click the *Show settings* button to configure [script execution options](xref:Script_execution_options).
 
 > [!NOTE]
-> Data can often contain multiple values, and therefore the values will be wrapped in a JSON array. To be consistent, this will also be the case when only a single value is provided.
+>
+> - Data can often contain multiple values, and therefore the values will be wrapped in a JSON array. To be consistent, this will also be the case when only a single value is provided.
+> - From DataMiner 10.4.0 [CU11]/10.5.2 onwards<!--RN 41495-->, when an interactive Automation script is launched using the *Launch a script* event and the parameter *useNewIASInputComponents=true* has been added to the URL of the low-code app, the latest version of the interactive Automation script UI is displayed. If the parameter is set to "false" or omitted from the URL, the low-code app will use the old UI to display the interactive Automation script.
 
 > [!IMPORTANT]
 > From DataMiner 10.3.0 [CU14]/10.4.0 [CU2]/10.4.5 onwards<!-- RN 39027 -->, linking a script parameter to empty data (or to an empty feed, prior to DataMiner 10.3.0 [CU21]/10.4.0 [CU9]/10.4.12<!--RN 41141-->) will fill it with an empty array. The dialog to manually enter a parameter will no longer be shown when the action is launched. This change can break existing implementations when it is not handled by the script.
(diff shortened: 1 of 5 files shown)


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Low-Code_Apps_configuration/LowCodeApps_event_config.md
```diff
@@ -45,7 +45,9 @@ To configure an event to launch a script:
 1. Optionally, click the *Show settings* button to configure [script execution options](xref:Script_execution_options).
 
 > [!NOTE]
-> Data can often contain multiple values, and therefore the values will be wrapped in a JSON array. To be consistent, this will also be the case when only a single value is provided.
+>
+> - Data can often contain multiple values, and therefore the values will be wrapped in a JSON array. To be consistent, this will also be the case when only a single value is provided.
+> - From DataMiner 10.4.0 [CU11]/10.5.2 onwards<!--RN 41495-->, when an interactive Automation script is launched using the *Launch a script* event and the parameter *useNewIASInputComponents=true* has been added to the URL of the low-code app, the latest version of the interactive Automation script UI is displayed. If the parameter is set to "false" or omitted from the URL, the low-code app will use the old UI to display the interactive Automation script.
```
jaklke, line 50: I would not mention the option useNewIASInputComponents=true, because this page is about configuring an LCA. The option can be used when consuming the LCA. Whether the new IAS components are used or not, is not defined by a LCA configuration.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/Other/Interactive_automation_script.md
```diff
@@ -12,6 +12,7 @@ This component allows you to visualize any interactive Automation script (IAS) w
 >
 > - This component is only available in the Low-Code Apps module.
 > - Scripts will time out after they have run for 15 minutes. This means that after a script has been shown in the component for 15 minutes, an error will be shown.
+> - From DataMiner 10.4.0 [CU11]/10.5.2 onwards<!--RN 41495-->, when the parameter *useNewIASInputComponents=true* has been added to the URL of the low-code app, the latest version of the interactive Automation script UI is displayed. If the parameter is set to "false" or omitted from the URL, the low-code app will use the old UI to display the script.
```
jaklke, line 15: I would not mention the option useNewIASInputComponents=true, because this page is about configuring an LCA. The option can be used when consuming the LCA. Whether the new IAS components are used or not, is not defined by a LCA configuration.

MariekeGO: Did you have a specific location in mind for this? We're having trouble pinpointing the best place to put this information, because currently the documentation does not contain info for end users on how to use specific dashboards/low-code apps. 
One option would be to add a new page at the top level for Dashboards and Low-Code Apps, listing the possible URL arguments you can use to access a dashboard or app. A disadvantage of that would be that it's a very small, very specific topic at top level — but I don't really see another place where it could fit right now.
Another option we thought was that possibly this argument could be considered a kind of soft-launch option, so we could add it to the soft-launch options documentation in the same way as the Cube-only arguments (though the info on activating soft-launch options will need to be updated to include the web apps then).
What do you think?

jaklke: the URL attribute can only be used for Low-Code Apps (not for dashboards), so the best fit would probably be where the LCA URL arguments are described : https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Executing_actions_via_URL.html
However, this new attribute is not an action, so the title and URL are not correct.

EdithLansens: @jaklke, I've implemented your feedback in [commit dcea034](https://github.com/SkylineCommunications/dataminer-docs/pull/4118/commits/dcea03421aa28bd653762dd1fd9f07c41982fd2d). I've changed the title of the page to "Using app URL actions and configurations". Let me know if these changes look OK.

EdithLansens: Last-minute change: The new title is "Configuring app actions and behavior via the URL".

jaklke: @EdithLansens, thanks for considering my feedback!

MariekeGO (approved):

PR #4119, opened 2025-01-09
DC251684: Update RN and DataMiner version to consult DxM logs

AndreRodriguesSkyline:


MariekeGO:
Could you change the italics on this page back to using asterisks instead of underscores? We use asterisks throughout the documentation, and I'd prefer not to introduce an inconsistency here. Possibly you are using a VS Code extension that changed these automatically? In that case I'd recommend switching to the [Learn Authoring Pack](https://marketplace.visualstudio.com/items?itemName=docsmsft.docs-authoring-pack) instead, which is most suited for use with DocFX-flavored markdown.

AndreRodriguesSkyline:
Sorry!! Seems that some VS extension is changing the asterisks to underscores. I've fixed it!

MariekeGO (approved):

AndreRodriguesSkyline:
Sure!! Thanks!!

PR #4121, opened 2025-01-09
RN 41511

EdithLansens:
@BramDevlaminck-Skyline, could you please review this pull request to confirm the accurate inclusion of release note 41511 in the developer documentation? Thanks in advance!

### DIFF
--- a/develop/api/NotifyTypes/NT_FILL_ARRAY_WITH_COLUMN.md
+++ b/develop/api/NotifyTypes/NT_FILL_ARRAY_WITH_COLUMN.md
@@ -76,7 +76,7 @@ protocol.NotifyProtocol(220 /*NT_FILL_ARRAY_WITH_COLUMN*/, columnInfo, values);
   <ColumnOption idx="0" pid="1001" type="retrieved" options="" />
   ```
 
-- The column data should always be wrapped in object arrays for the data to be correctly interpreted by the SLProtocol process.
+- The column data should always be wrapped in object arrays for the data to be correctly interpreted by the SLProtocol process. If an invalid type, like a string array, is passed, the function will fail and throw an error, which will automatically be logged in the element's log file. Prior to DataMiner 10.4.0 [CU11]/10.5.2<!--RN 41511-->, the function fails silently in such cases, without logging or throwing an error.
 
 - It is possible to specify a datetime that will be applied to all values passed in the parameter set.<!-- RN 21482 -->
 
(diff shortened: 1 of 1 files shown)


develop/api/NotifyTypes/NT_FILL_ARRAY_WITH_COLUMN.md
```diff
@@ -76,7 +76,7 @@ protocol.NotifyProtocol(220 /*NT_FILL_ARRAY_WITH_COLUMN*/, columnInfo, values);
   <ColumnOption idx="0" pid="1001" type="retrieved" options="" />
   ```
 
-- The column data should always be wrapped in object arrays for the data to be correctly interpreted by the SLProtocol process.
+- The column data should always be wrapped in object arrays for the data to be correctly interpreted by the SLProtocol process. If an invalid type, like a string array, is passed, the function will fail and throw an error, which will automatically be logged in the element's log file. Prior to DataMiner 10.4.0 [CU11]/10.5.2<!--RN 41511-->, the function fails silently in such cases, without logging or throwing an error.
```
BramDevlaminck-Skyline, line 79:
```md
- The column data should always be wrapped in an object array or string array for the data to be correctly interpreted by the SLProtocol process. If any other type is passed, the function will fail and throw an error, which will automatically be logged in the element's log file. Prior to DataMiner 10.4.0 [CU11]/10.5.2<!--RN 41511-->, the function fails silently in cases where something else than an object array was passed, without logging or throwing an error.
````

**Review by BramDevlaminck-Skyline — CHANGES_REQUESTED** [review]
I suggested a new sentence.
The difference is that from now on we **also support a string array as input** (next to object arrays which was the only option previously).
Additionally, **any other type that is passed will trigger an error**.
Previously, the only valid option was using an object array, everything else (including using a string array) failed silently.

File: develop/api/NotifyTypes/NT_FILL_ARRAY_WITH_COLUMN.md (hunk 1 above)
Comment (EdithLansens) [reply develop/api/NotifyTypes/NT_FILL_ARRAY_WITH_COLUMN.md:79]: Thanks for the correction, @BramDevlaminck-Skyline. I've updated the text based on your input.

```suggestion
- The column data should always be wrapped in object or string arrays for the data to be correctly interpreted by the SLProtocol process. If a type other than an object or string array is passed, the function will fail and throw an error, which will automatically be logged in the element's log file. Prior to DataMiner 10.4.0 [CU11]/10.5.2<!--RN 41511-->, the function fails silently if anything other than an object array is passed, without logging or throwing an error.
```

BramDevlaminck-Skyline (approved):
👍

MariekeGO (approved):

PR #4129, opened 2025-01-10
Introducing the GQI DxM

wimme:


release-notes/Web_apps/Web_apps_Feature_Release_10.5/Web_apps_Feature_Release_10.5.2.md
```diff
@@ -30,21 +30,9 @@ Up to now, all GQI-related operations were executed by the SLHelper process, whi
 - System load by GQI operations can now be balanced among the different DataMiner Agents in a DMS.
 
 > [!IMPORTANT]
-> Until further notice, the DataMiner web apps will continue to use the SLHelper process to execute GQI-related operations. If you want them to use the new GQI DxM instead, update the *C:\\Skyline DataMiner\\Webpages\\web.config* file accordingly.
+> Until further notice, the DataMiner web apps will continue to use the SLHelper process to execute GQI-related operations. If you want them to use the new GQI DxM instead, update the *C:\\Skyline DataMiner\\Webpages\\API\\Web.config* file accordingly.
 
-##### Specifying when idle child processes should be terminated
```
MariekeGO, line 35: If at all possible, I would prefer to keep this information in the release notes as well. This because when a user looks at the release notes for an older DataMiner version, those are supposed to show the info as it was at the time of publication; but if we replace the info with a link to a section that will change over time, it will become very difficult for the user to still know exactly what the feature is like in that particular version.

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.md
```diff
@@ -0,0 +1,87 @@
...
+
+# GQI DxM
+
+The GQI DxM is a [DataMiner Extension Module (DxM)](xref:DataMinerExtensionModules) that hosts the GQI core engine, enabling efficient data retrieval through [GQI queries](xref:About_GQI).
+
+Running GQI as an extension module has several advantages:
+
+- **Independent updates**: Using a [web-only upgrade package](xref:Upgrading_Downgrading_Webapps), GQI can be updated independently from the main DataMiner core software, eliminating the need for a core software restart during updates.
+- **Load balancing**: With the DxM, GQI operations can distribute system load more effectively across multiple DataMiner Agents within a DataMiner System (DMS).
+- **Resource visibility**: Enhanced monitoring capabilities provide greater insight into the resource consumption of individual extensions.
```
MariekeGO, line 13: It isn't clear to me here what the "enhanced monitoring capabilities" are. Do you mean that it's easier to monitor the exact resource consumption of GQI when it's a separate extension?

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.md
```diff
@@ -0,0 +1,87 @@
...
+- **Improved dependency resolving**: Using the DxM simplifies and optimizes the resolution of DLL dependencies, preventing conflicts.
+- **Performance enhancements**: The GQI DxM leverages the latest advancements in .NET 8 for improved speed and efficiency, uses proto-first communication, prevents SLNet throttling, etc.
+
+## Installation
+
+The GQI DxM is supported from DataMiner 10.5.0 onwards. It gets installed or updated automatically from DataMiner version 10.5.2/10.5.0 onwards. The GQI DxM also gets updated when installing a web-only upgrade package.<!-- RN 41811 -->
```
MariekeGO, line 19: We try to always mention both feature and main release, so that it's clear for people on either track which minimum version applies for them. OK to update this like this?
```suggestion
The GQI DxM is supported from DataMiner 10.5.0/10.5.2 onwards, and automatically gets installed and updated by upgrade packages from that version onwards. It also gets updated when you install a web-only upgrade package.<!-- RN 41811 --> 
```

wimme: The GQI DxM is also supported on 10.5.1 although it does not get installed automatically. That's why I put it this way. But since there's no public download of the GQI DxM available for 10.5.1, it makes sense to indeed say it's supported from 10.5.2 onwards for the FR.

release-notes/Web_apps/Web_apps_Feature_Release_10.5/Web_apps_Feature_Release_10.5.2.md
wimme: Ok I'll put it back

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.md
wimme: Not as GQI as a whole. GQI supports running your own extensions. Each extension will now run within its own process. This is visible in task manager, which gives more visibility which extension is taking most resources.

MariekeGO (approved):

PR #4145, opened 2025-01-14
RN 41561

EdithLansens:
@Robin-Debel, could you please review this pull request to confirm the accurate inclusion of release note 41561 in the user guide?

As this release note introduces the URL data category to the Low-Code Apps module, I’ve moved the URL data page from the Dashboards app section to the Dashboards and Low-Code Apps section. Additionally, I’ve reworked the page. Let me know if you think this wasn’t the right approach.

image

Thanks in advance!


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Configuring_components/Dynamically_Referencing_Data_in_Text.md
```diff
@@ -80,11 +80,18 @@ To create URL references inside textual settings, use the following syntax:
...
+- If you want to pass a number as a query parameter to a URL, you could use the following syntax:
+
+  ```txt
+  {URL."Query parameters".Numbers.Value}
+  ```
+
+- If you want to pass a hostname and IP port to a URL, you could use the following syntax:
+
+  ```txt
+  {URL.DMAIP.Value}
```
EdithLansens, line 94: @Robin-Debel, I’ve added the IntelliSense example you included in the original release note. Could you confirm if this description is correct?

Additionally, we mention above that the syntax {URL."Category name".type.Value} is used to create URL references inside textual settings. However, in this example, it seems the category name is missing. Could you clarify if this is intentional? Thanks!

Robin-Debel: This is indeed intentional. I believe the {URL."Category name".type.Value} will still be the most used way of configuring a syntax, but additional 'static vars' will be created so it won't be like that always anymore

MariekeGO (approved):

PR #4150, opened 2025-01-15
Update DOM_Generating_code_with_the_DOM_editor.md

Ivehe:
I don't believe the Visual Studio solution troubleshooting step should be in this tutorial. It's off-topic.


MariekeGO:
@Ivehe, we added this info to this tutorial just in case a user encounters the same issue as we encountered while testing the tutorial, as otherwise this will lead to questions and loss of time investigating the issue. Do you have a different suggestion as to how we may be able to tackle this?

Ivehe (commented):
Looks good!

MariekeGO (approved):

PR #4152, opened 2025-01-15
changelog 15 01 -1

KoenAnnys:


release-notes/Cloud/DCP_change_log_2024.md
MariekeGO, line 1: Earlier, we decided not to change the name of this page in order to avoid breaking links, since the change is really only visible under the hood so the risk isn't worth the reward. @JordyGit, can we implement this name change after all now, or would it be better to revert this particular change?

KoenAnnys: i wasnt aware of ealier decision tbh. if we expect breaking links i'm good to revert it.

MariekeGO (approved):

PR #4170, opened 2025-01-21
Update Prerequisites.md

ThomasW-Skyline:
Update DIS prerequisites to indicate issues with DataMiner Templates

### DIFF
--- a/develop/TOOLS/DIS/Installing_and_configuring/Prerequisites.md
+++ b/develop/TOOLS/DIS/Installing_and_configuring/Prerequisites.md
@@ -34,12 +34,11 @@ https://*.skyline.be/*
 > [!IMPORTANT]
 > DIS works with Visual Studio Community, Visual Studio Professional, and Visual Studio Enterprise. Visual Studio Code and Visual Studio for Mac are not supported.
 
-## DataMiner Templates
+### nuget.org configuration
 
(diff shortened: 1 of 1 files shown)


develop/TOOLS/DIS/Installing_and_configuring/Prerequisites.md
```diff
@@ -34,6 +34,13 @@ https://*.skyline.be/*
 > [!IMPORTANT]
 > DIS works with Visual Studio Community, Visual Studio Professional, and Visual Studio Enterprise. Visual Studio Code and Visual Studio for Mac are not supported.
 
+## DataMiner Templates
```
MariekeGO, line 37: The way this is listed now, it creates the impression that DataMiner templates are a prerequisite, but do I understand it correctly that actually configuring and enabling nuget.org is the prerequisite here, required in order to be able to install DataMiner templates?

ThomasW-Skyline: Yes, indeed your assumptions are fully right

MariekeGO: I've adjusted the PR to prevent any possible confusion. Could you take a look and let me know if this looks OK to you?

PR #4177, opened 2025-01-22
Update SLProtocol.cs

SkylineThomasCR:
Mention that method GetConnectivityConnections returns null when the specified element is stopped or unavailable.

@PedroDebevere can this be added to the xmldoc shipped with the devpacks?

### DIFF
--- a/src/DataMiner/SLManagedScripting/SLProtocol.cs
+++ b/src/DataMiner/SLManagedScripting/SLProtocol.cs
@@ -3288,7 +3288,7 @@ namespace Skyline.DataMiner.Scripting
 		/// <param name="iEId">The element ID of the element.</param>
 		/// <param name="nameFilter">The name filter.</param>
 		/// <param name="exported"><c>true</c> to include connections of DVE children; otherwise, <c>false</c>.</param>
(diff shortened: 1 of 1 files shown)


src/DataMiner/SLManagedScripting/SLProtocol.cs
```diff
@@ -3288,7 +3288,7 @@ public interface SLProtocol
 		/// <param name="iEId">The element ID of the element.</param>
 		/// <param name="nameFilter">The name filter.</param>
 		/// <param name="exported"><c>true</c> to include connections of DVE children; otherwise, <c>false</c>.</param>
-		/// <returns>The DCF connections that are known by the specified element and that match the specified name filter.</returns>
+		/// <returns>The DCF connections that are known by the specified element and that match the specified name filter. Can return <see langword="null"/> when the specified element is stopped or unavailable.</returns>
```
MariekeGO, line 3291: Does it always return null when the element is stopped/unavailable, or only in some cases? If it always does this, I'd suggest to rephrase this as follows (also below):

```suggestion
		/// <returns>The DCF connections that are known by the specified element and that match the specified name filter. Returns <see langword="null"/> when the specified element is stopped or unavailable.</returns>
```

SkylineThomasCR: It will always return null. Thanks for the suggestion.

MariekeGO (approved):

PR #4179, opened 2025-01-22
Update Connect_to_cloud_requirements.md

alexpresland:
While integrating we needed to add dataminerservices.b2clogin.com as well, as this is used for authentication


MariekeGO:
> This is indeed true but the DataMiner agent does not need access to this URL to make the connection work. This URL is needed for the login towards our platform to setup the connection. In theory this URL can be copied and pasted on to another machine to complete the setup process so it is not mandatory to have access to that endpoint. @MariekeGO would it be better to make this a note instead to advice users that this endpoint will be used once during the setup of the connection?

Indeed, I'll adjust this so we instead mention this in a note.

baptistepattyn (approved):

MariekeGO (approved):

PR #4181, opened 2025-01-23
User Guide - DOM - Tutorial - Dom Viewer

ruicamacho:


MariekeGO:
@ruicamacho Given the nature of this content, it doesn't really fit in the tutorial format. Giving a quick tour of the UI is something we typically do on a how-to page, e.g. [Working with the EVS Cerebrum application](https://docs.dataminer.services/solutions/product_solutions/EVS_Cerebrum/Using_EVS_Cerebrum.html). As you probably noticed yourself, the tutorial format with its strict order of steps just doesn't fit with this kind of content.
As such, I think it would be better if we changed this to a regular how-to page. This may also be the perfect chance for us to properly add the DOM Viewer solution to the [Solutions](https://docs.dataminer.services/solutions/solution_index.html) section of the documentation. Right now it's only mentioned in the [DataMiner Overview](https://docs.dataminer.services/dataminer-overview/DataMiner_Solutions/DOM_Viewer/domviewer_about.html), but I'm not sure if that content even really belongs there. In the overview section, I think users will expect more generic bird's-eye info on the general capabilities of DataMiner, rather than the specific help page for one solution.
My recommendation would therefore be to make a DOM Viewer section where appropriate under [DataMiner Solutions](https://docs.dataminer.services/solutions/solution_index.html), and to move both the content from the page in the DataMiner Overview (or most of it) and the content of the new page in your pull request over there. Let me know if you agree or if you want to discuss this further! If you don't have much time, we can take care of everything for you.

ruicamacho:
Hey Marieke.
Thank you for the feedback. 
I think I will defer to your better judgement and implement your proposed changes.
I have moved the existing "Dom Viewer (About)" and the new "Using Dom Viewer" pages from DataMiner Overview to Solutions under DevTools.
Hopefully you'll agree. 
If not let me know what additional changes you wish me to do.

MariekeGO (approved):

PR #4204, opened 2025-01-29
GQI DxM - Configuration - Remove unnecessary logging configuration

skyline-rge:
See [RN42075: GQI - Configuration - Remove obsolete configuration for the .NET Core logging provider](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=42075).

### DIFF
--- a/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_Logging.md
+++ b/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_Logging.md
@@ -66,12 +66,6 @@ Add the following configuration:
 
 ```json
 {
-  "Logging": {
-    "LogLevel": {
-      "Default": "Debug",
-      "Microsoft": "Information"
-    }
-  },
   "Serilog": {
     "MinimumLevel": {
       "Default": "Debug",


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_Logging.md
MariekeGO, line 1: Is this change compatible with earlier DataMiner versions from before this change was introduced? If not, we will have to mention the info for older versions as well, because the help must be compatible with any supported DataMiner version.

MariekeGO: At present, the [documentation ](https://docs.dataminer.services/user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.html)states that the DxM is "supported from DataMiner 10.5.0/10.5.2 onwards", so users will get the impression that it's actually been released already—but I suppose it's actually only available as an opt-in feature right now? Will it be fully released in 10.5.0 as well? In that case, since users are supposed to always use the latest feature release and the main release does not change, we could indeed keep from adding the extra version info here. However, it will be important to properly document this change in the release notes so that users who are already using the feature will be informed (which makes me think the RN should not be set to "Internal" as is the case now).

skyline-rge: The GQI DxM was included in the upgrades from 10.5.0/10.5.2 onwards, which means it is _technically_ supported in the sense that someone could enable it and it would work. I think that's why this was already added like that in the docs.
However, it was only meant to be used internally; we don't support it publicly yet in the sense that we don't want any external users using it yet and there are no guarantees for backward compatibility for GQI DxM specific features.

The 10.5.3 release would be the first _public_ release (still opt-in).
Does that make sense?

Regardless, this particular change is both forward and backward compatible so it wouldn't hurt either way to use the old configuration on the new release or vice-versa.

MariekeGO: That makes it fine to include this change here as it is, but I do believe we will need to include the release note in the public release notes. Our [communication for 10.5.2](https://docs.dataminer.services/release-notes/Web_apps/Web_apps_Feature_Release_10.5/Web_apps_Feature_Release_10.5.2.html#dataminer-upgrade-packages-now-include-the-new-gqi-dxm-id-39145-id-41097-id-41811) gave no indication whatsoever that this feature was not supposed to be used yet, so some external people may be using this already. Going forward, we will have to be more careful to ensure that things that should not be used yet are also not yet published, or in the very least that if we publish them, we make it clear that they are preview features for which no backwards compatibility is guaranteed.

MariekeGO (approved):

PR #4207, opened 2025-01-30
Update KI_Upgrade_to_1051_stuck.md

MariekeGO:


wobog:
Only saw this after creating #4211 however the designation "mostly" is actually misleading. It occurs on a single agent as soon as the user is part of multiple groups. Which is not related to being in a cluster.

MariekeGO:
> Only saw this after creating #4211 however the designation "mostly" is actually misleading. It occurs on a single agent as soon as the user is part of multiple groups. Which is not related to being in a cluster.

Thanks for the info! I'll close this PR and merge yours instead 😊

PR #4211, opened 2025-01-30
Update KI_Upgrade_to_1051_stuck.md

wobog:
Issue can also occur on a single agent


MariekeGO:
This was already mentioned in #4207; but there the sentence is adjusted instead of left out. Which of the options do you think is best?

wobog:
I would leave it out as the issue is unrelated to being in a cluster and can occur in both situations (whether or not the trigger is the same I don't know)

MariekeGO (approved):

PR #4248, opened 2025-02-06
Permission flags: allow to manage custom presets

lisdjong:
There is new functionality added in cube (see RN: https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=41656).

### DIFF
--- a/user-guide/Advanced_Functionality/Security/Basic_security_configuration/Managing_user_groups/Configuring_a_user_group.md
+++ b/user-guide/Advanced_Functionality/Security/Basic_security_configuration/Managing_user_groups/Configuring_a_user_group.md
@@ -35,7 +35,7 @@ To configure a user group:
 
 1. To change the user permissions for a group, go to the *Permissions* tab, select the *Rights* subtab and select the appropriate checkboxes.
 
-   Alternatively, instead of selecting the different separate permissions, you can select a preset in the drop-down list at the top of the card. The following presets are available:
+   Alternatively, instead of selecting the different separate permissions, you can select a preset at the right of the card. The following presets are available:
 
    - **No rights**: Deselects all rights.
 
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Functionality/Security/Basic_security_configuration/Managing_user_groups/Configuring_a_user_group.md
```diff
@@ -35,27 +35,54 @@ To configure a user group:
...
+
+- **Administrators**: Selects all rights.
+
+In addition to these default presets, from DataMiner 10.4.0 [CU12]/10.5.0 [CU1]/10.5.3 onwards, custom presets are also supported. You can manage these with the buttons at the top of the pane on the right:
+
+| Button | Description |
+|--|--|
+| ![Reset button](~/user-guide/images/Security_preset_reset.png) | Reverts all changes made to the rights of the currently selected user group. |
+| ![Export button](~/user-guide/images/Security_preset_export.png) | Exports the currently selected rights/preset to a preset file (in JSON format). |
+| ![Import button](~/user-guide/images/Security_preset_import.png) | Imports a preset file. |
```
MariekeGO, line 71: When you import a preset file, does it only get added to the list, or is it also applied to the selected group?

user-guide/Advanced_Functionality/Security/Basic_security_configuration/Managing_user_groups/Configuring_a_user_group.md
```diff
@@ -35,27 +35,54 @@ To configure a user group:
...
+
+To **add a new preset**, first configure the rights you want to add to the preset, then click the *Save rights as preset* button and specify the name for the preset.
+
+Updating an existing custom preset can be done as follows:
+
+- To rename a preset, select the preset, delete it, and then save it with another name.
+- To change the rights configured for a preset, save it as a new preset using the same name.
+
+> [!NOTE]
+> Any changes to the presets will not automatically be reflected in other places.
```
MariekeGO, line 82: I'm afraid this might not be clear enough and could lead to confusion. Could I rephrase this as follows, or should I interpret this differently?
```suggestion
> Changes to presets do not get pushed from the server, which means that when you make changes to a preset, other client instances will not automatically be updated.
```

MariekeGO (commented):
@lisdjong, I've reviewed your pull request and made some changes. Among others, I've moved the presets info to a separate section to keep the procedure user-friendly, and I've also added RN 40803 (since this is related to the same topic). I have two more questions left; could you take a look?

user-guide/Advanced_Functionality/Security/Basic_security_configuration/Managing_user_groups/Configuring_a_user_group.md
lisdjong: You can see it as follows:
- Group1 has rights selected that is exactly the same as is saved in PresetX, so this preset is selected.
- Group2 has the exact same rights, so also PresetX is selected.
- When changing the rights in Group1 and then saving it as PresetX, PresetX is updated.
- Group1 has these changes still selected, so PresetX is again selected.
- When going to Group2, it will see that its selected rights are not the same as PresetX because does not listen to changes to PresetX. It just keeps its rights, it does not keep track of the rights saved in PresetX. PresetX will also not be selected.
It also works this way in other client instances. They also do not listen to the preset, they only keep track of the selected rights and compares it to the rights of the presets that it got from the server.

MariekeGO: So if you make changes to a preset, these are only automatically applied to the currently selected group, and other groups using that preset will only be updated when a manual change is applied to them? And do the other clients need to reconnect to the server to get the updated preset (e.g. to be able to apply it to other groups)?

lisdjong: The preset is only saved (and since you save it in that group, it is already applied). Other groups will only be updated if you specifically go to a group and select the preset.
Other clients will need to at least close and reopen the system center card to get the updated presets.

I also want to emphasize that applied means that the rights are only selected. You still need to push the Apply button if you want the rights for a group to be saved.

MariekeGO: I've made a couple more changes based on your feedback. Could you check if everything still looks OK for you?

lisdjong: Yes, the changes are OK.

MariekeGO (approved):

PR #4254, opened 2025-02-07
.NET 8.0 link update

MariekeGO:


MariekeGO:
> Can we rephrase "download the latest ASP.NET Core Runtime Hosting Bundle from..." to something like "download the latest Hosting Bundle under ASP.NET Core Runtime from..."? Just to make clear you need that hosting bundle, I made that mistake before if this is all in one sentence...

OK, done!

BertVandenberghe (approved):

PR #4277, opened 2025-02-11
Addition of GitHub support note to Package Creation docs

rafaelabranchesskyline:
Added GitHub Automation script repository support note to Application package creation documentation


MariekeGO:
This section is currently being fully rewritten by @janstaelensskyline (see #4278), so this change will not be necessary.

PR #4286, opened 2025-02-14
[NATS] BPA/Prerequisite change for NATS communication

lustpieter:
RN 42206


MariekeGO (commented):
I noticed that the version info of this RN on intranet is still set to TBD. Will this still be included in 10.5.4, or will the release versions in this PR need to be updated?

MariekeGO (approved):

PR #4288, opened 2025-02-16
Update Sync_Check.md

joaquimad:
Add Examples of the Syncinfo output file

### DIFF
--- a/user-guide/Reference/DataMiner_Tools/Sync_Check.md
+++ b/user-guide/Reference/DataMiner_Tools/Sync_Check.md
@@ -18,11 +18,11 @@ To use this script:
 
 1. Extract the zip file you downloaded.
 
-2. Import the script *Script_SyncCheck.xml* on a DMA in the DataMiner System you are troubleshooting. The script will become available in the *Skyline-TechSupport* folder in the Automation module.
+1. Import the script *Script_SyncCheck.xml* on a DMA in the DataMiner System you are troubleshooting. The script will become available in the *Skyline-TechSupport* folder in the Automation module.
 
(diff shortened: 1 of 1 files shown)


user-guide/Reference/DataMiner_Tools/Sync_Check.md
```diff
@@ -64,5 +64,75 @@ The script checks the following things:
 
   - Inconsistent folder name (different name or capitalization)
 
+  ## Examples of the Syncinfo output file
```
MariekeGO, line 67: Is there a specific reason why you used the numbering format "1.1", "1.2", etc. in the section below this? We don't use this anywhere on docs, so I'd prefer not to do so here if at all possible.

user-guide/Reference/DataMiner_Tools/Sync_Check.md
```diff
@@ -64,5 +64,118 @@ The script checks the following things:
 
   - Inconsistent folder name (different name or capitalization)
 
+## SyncInfo output file
```
MariekeGO, line 67: I've adjusted this new section to better match our usual format. Note also that the old format included some things that would not be rendered correctly, so I'd recommend taking a look at [our markdown guide](https://docs.dataminer.services/contributing/CTB_Markdown_Syntax.html) for future contributions.
Meanwhile, could you take a look at the updated pull request and let me know if you see anything that still needs to be changed?

joaquimad: This new version sounds good to me. 
And thanks for the markdown guide link, will have a look at it in future cases.

MariekeGO (approved):

PR #4296, opened 2025-02-17
Update DOM_Generating_code_with_the_DOM_editor.md

joaquimad:
Project template name update

### DIFF
--- a/user-guide/Advanced_Modules/DOM/tutorials/DOM_Generating_code_with_the_DOM_editor.md
+++ b/user-guide/Advanced_Modules/DOM/tutorials/DOM_Generating_code_with_the_DOM_editor.md
@@ -78,11 +78,10 @@ To create an Automation script solution:
 
 1. In Visual Studio, select *File > New > Project*.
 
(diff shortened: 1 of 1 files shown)


user-guide/Advanced_Modules/DOM/tutorials/DOM_Generating_code_with_the_DOM_editor.md
MariekeGO, line 1: Wouldn't it be better if we just replaced *DataMiner Automation Script Solution (Skyline Communications)* with *DataMiner Automation Script Project (Skyline Communications)* in line 81? Is there a specific reason why this should be in a separate comment instead?

joaquimad: Thought about that, but in the Kata video, the "DataMiner Automation Script Solution (Skyline Communications)" template is mentioned. So, I think it is best to have it like that so people can understand that the one mentioned in the Kata video is no longer available. Do you agree?

MariekeGO (approved):

PR #4298, opened 2025-02-18
Update STaaS documentation to always direct to support for the endpoints

ruben12321:
Update docs to always direct the user to staas@dataminer.services as the IPs mentioned before aren't enough. The storage queue and eventhub need different ranges and hostnames


LouisSkyline (approved):

MariekeGO (approved):

user-guide/Advanced_Functionality/Databases/STaaS/STaaS.md
MariekeGO, line 1: @ruben12321, I noticed that two of these endpoints are also mentioned on the [STaaS troubleshooting page](https://docs.dataminer.services/user-guide/Troubleshooting/Procedures/STaaS_Troubleshooting.html#verify-your-dma-has-a-working-internet-connection). Should that page also be updated?

PR #4301, opened 2025-02-19
Update UIComponentsMatrixDiscreetInfo.md

jaklke:
There can only be one parameter of type 'discreet info' in the protocol.


develop/schemadoc/Protocol/Protocol.Params.Param.Type.md
MariekeGO, line 1: @jaklke, I've also added your info on this page, as I believe it is also relevant here. Could you take a look and let me know if this looks OK for you?

MariekeGO (approved):

PR #4328, opened 2025-02-26
Use 'SwarmingHelper' instead of the request messages in the booking swarming docs

Seppe-Vanhee:
- Use the 'SwarmingHelper' in the code examples as this is the official public API
- Removed the example with async progress events for now since this is not yet supported in the helper.
- Updated the element swarming docs to use 'engine.GetUserConnection()' since usage of 'Engine.SLNetRaw' directly is discouraged.


JorgeFilipeDias (approved):
Looks good to me

MariekeGO (approved):

PR #54, opened 2022-03-07
Updates for RNs 10.2.0 - 10.2.3 + other updates

PedroDebevere:

PR #4348, opened 2025-03-04
Update DOM_Using_DOM_Viewer.md

sielschoonbaert:


MariekeGO:
There already is a tip referring to Kata #54 right above the place where the new tip of the current PR is added, making this additional tip unnecessary.

PR #4349, opened 2025-03-04
added protocol.extension util nuget

PedroDebevere:


MariekeGO (changes requested):
I tested this new addition with a test build, and noticed something that isn't OK on the /develop/api/types/Skyline.DataMiner.Utils.Protocol.Extension.ProtocolExtension.html page:
image
Could you update this so the release version is displayed correctly?

MariekeGO:
I will close this PR for now; a new one will be made when the issue is fixed.

PR #4366, opened 2025-03-06
DCP changelog 06-03-2025

SkylineKevinMOR:


release-notes/Cloud/DCP_change_log_2025.md
```diff
@@ -9,6 +9,14 @@ The dataminer.services platform gets updated continuously. This change log can h
 > [!NOTE]
 > Many features on dataminer.services are dependent on DxMs. You can find the change logs for these under [DxM release notes](xref:DxM_RNs_index).
 
+### 6 March 2025 - Fix - Catalog - Deploy button state when not authenticated [ID 42439]
+
+When users are not authenticed, the state of the deploy button will now be applied correctly if their issues with the version you want to deploy.
```
MariekeGO, line 14: For a fix, the description should clearly describe what the issue is, rather than what the situation now is. Users looking in the fixes will be searching for what they encountered, so we need to make sure this can easily be spotted. Could you rephrase both this item and the item below so it's a clearly recognizable description of the issue?

MariekeGO (changes requested):

release-notes/Cloud/DCP_change_log_2025.md
```diff
@@ -9,6 +9,29 @@ The dataminer.services platform gets updated continuously. This change log can h
 > [!NOTE]
 > Many features on dataminer.services are dependent on DxMs. You can find the change logs for these under [DxM release notes](xref:DxM_RNs_index).
 
+### 6 March 2025 - Fix - Catalog - Deploy button state when not authenticated [ID 42439]
+
+The deploy button of a catalog version will now be disabled if there are any issues with the version.
+Before the button would always be enabled if the user was not authenticated.
+Note: the button can remain disabled after logging in when the user does not have a working cloud connected DMS
```
MariekeGO, line 16: I'm not sure I fully understand this. Do you mean that the Deploy button for an item version with issues is only disabled in case the user is not authenticated? Or do you mean that it should have been disabled regardless of whether the user was authenticated or not, but this wasn't the case yet? If it's the latter, what does the cloud connection have to do with the button remaining disabled?

release-notes/Cloud/DCP_change_log_2025.md
```diff
@@ -9,6 +9,29 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+### 6 March 2025 - Fix - Catalog - Increased polling duration for deployment status [ID 42405]
+
+When deploying an item from the Catalog, the duration for checking the status of the active deployment has been increased to around 5 minutes, before this was only 10 seconds.
+If the deployment would still be pending after a certain amount of polling tries, it will result in a timeout state, before it would remain on pending after 10 seconds.
```
MariekeGO, line 21: As this is a fix, it must be made more clear what the issue is the user experienced. Would this be a correct description?
```suggestion
When an item was deployed from the Catalog, it could occur that the status of the deployment was not updated correctly, showing that the deployment was still pending even though this was no longer the case. To resolve this issue, the duration for checking the status of the active deployment has been increased to around 5 minutes, while before this was only 10 seconds. If the deployment is still pending after a certain amount of polling tries, this will result in a timeout state.
```

release-notes/Cloud/DCP_change_log_2025.md
brayunm: It's indeed the latter, but when a user is not authenticated yet we cannot check everything yet.

Once a user is logged in we can check more things e.g. if the user has a cloud connected system that has a working cloud connection. If we detect any issues here, the deploy button would be disabled again (while in the unauthenticated view it was enabled).

I have hard time explaining this myself, if you need more information I can demo it for you during a call.

release-notes/Cloud/DCP_change_log_2025.md
brayunm: This is indeed correct

release-notes/Cloud/DCP_change_log_2025.md
EdithLansens:
```suggestion
The *Deploy* button for a Catalog version was always enabled when the user was not authenticated, even if the version had issues. This occurred because certain checks could not be performed before logging in. This issue has now been resolved. The *Deploy* button will be correctly disabled if there are any issues with the version, regardless of authentication status. If a user is not logged in, the button may still appear enabled initially, but once they log in, additional checks will be performed. If any issues are detected, such as the absence of a working cloud-connected DMS, the *Deploy* button will remain disabled.
```

@brayunm, based on your explanation, would this be a correct description?

brayunm: Yes, thanks :)

EdithLansens:
Is this pull request OK to merge, @SkylineKevinMOR and @brayunm?

release-notes/Cloud/DCP_change_log_2025.md
```diff
@@ -9,10 +9,36 @@ The dataminer.services platform gets updated continuously. This change log can h
 > [!NOTE]
 > Many features on dataminer.services are dependent on DxMs. You can find the change logs for these under [DxM release notes](xref:DxM_RNs_index).
 
+### 11 March 2025 - Enhancement - Catalog - Ranges displayed for all items [ID 42426]
+
+The Catalog now displays version ranges for all item types. Previously, ranges were only displayed for connectors. A range is created when a version ID follows a semantic format (e.g. x.x.x.x or x.x.x). Other versions are grouped under *other*.
+
+### 11 March 2025 - Enhancement - Catalog API - Public endpoint to retrieve recommended versions
```
MariekeGO, line 16: Could you add the intranet RN number here?

release-notes/Cloud/DCP_change_log_2025.md
```diff
@@ -9,10 +9,36 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+The Public Catalog API now also has an endpoint that can be used to get the recommended versions of any Catalog item.
```
MariekeGO, line 18: Could you (briefly) clarify how this can be used?

release-notes/Cloud/DCP_change_log_2025.md
brayunm: I don't have a RN for this one, I can still make one if this is needed

release-notes/Cloud/DCP_change_log_2025.md
MariekeGO: @SkylineKevinMOR, could you adjust the description so it will be clear to the user what has changed and what the impact of this is for them?

brayunm:
```suggestion

All of our different API's (key, user & public) now have a shared logic to determine recommended versions for a catalog item.
Before it could occur that you would see different recommended versions based on the API that was called.
```

brayunm:
The described changes have been deployed @EdithLansens & @MariekeGO

MariekeGO (approved):

PR #4412, opened 2025-03-13
Update Protocol.Params.Param.Display.Range.md

LuisFreitas-Skyline:
@SamuelRDuarte @fsmeneses @HansSolos

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Params.Param.Display.Range.md
+++ b/develop/schemadoc/Protocol/Protocol.Params.Param.Display.Range.md
@@ -31,4 +31,4 @@ Defines the parameter value range.
 > - The range information is used by the Analytics component of DataMiner for more accurate trend/alarm prediction.
 
 > [!IMPORTANT]
-> When using the [time](xref:Protocol.Params.Param.Measurement.Type-options) option, for Cube to correctly display the tooltip and slider within the expected ranges, both the low and high intervals must be specified if a range is required—or neither should be.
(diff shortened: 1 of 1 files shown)


develop/schemadoc/Protocol/Protocol.Params.Param.Display.Range.md
```diff
@@ -29,3 +29,6 @@ Defines the parameter value range.
 > - The range information is used when defining alarm thresholds.
 > - The range information is used by UI components such as parameters of measurement type "analog".
 > - The range information is used by the Analytics component of DataMiner for more accurate trend/alarm prediction.
+
+> [!IMPORTANT]
+> When using the [time](xref:Protocol.Params.Param.Measurement.Type-options) option, for Cube to correctly display the tooltip and slider within the expected ranges, both the low and high intervals must be specified if a range is required—or neither should be.
```
MariekeGO, line 34: This isn't entirely clear. Do I interpret it correctly like this?
```suggestion
> When you use the [time](xref:Protocol.Params.Param.Measurement.Type-options) option, you should either specify both the low and high intervals if a range is required, or specify neither. Otherwise, Cube will not be able to correctly display the tooltip and slider within the expected ranges.
```

HansSolos: The goal is to inform to the user that Range should always fill in both Low and High and avoid filling in only one of them. Reason:  In almost all references in Cube where a Range is used, the  readonly on the parameter info object called HasRange property is used to verify if a Range is configured. This property checks if a Low is filled in and a High is filled in. If not, then the Range will not be used.

LuisFreitas-Skyline: @MariekeGO I'm fine with your suggestion.

MariekeGO (approved):

PR #4413, opened 2025-03-13
Swarming - update startup stats

RobbeDGH:
Redid startup performance tests.
- The timings were for element start up only (all elements) but not the whole agent
- The timings were on prem agent with on prem db vs staas. Redid them for daas + staas (swarming vs non swarming)


RobbeDGH:
@wdm-slc fyi

MariekeGO (approved):

PR #4422, opened 2025-03-14
Update EnableSwarmingTutorial.md

EmmaSaldi:
Added link to today's kata session - Couldn't find this tutorial under tutorials - not sure if we need to add it there as well.


MariekeGO:
@EmmaSaldi, it's included under the [Advanced admin tutorials](https://docs.dataminer.services/tutorials/Advanced_Admin_Tutorials.html), because in principle only administrators should do such far-reaching changes as enabling Swarming.

PR #4425, opened 2025-03-14
RN 42226

EdithLansens:
@Robin-Debel, could you please review this pull request to confirm the accurate inclusion of release note 42226 in the user guide? Thanks in advance!


user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Maps.md
```diff
@@ -44,7 +44,11 @@ On the *Component > Layout* tab, you can configure the following settings under
 
 - **Identifier**/**Latitude**/**Longitude**: For each dimension, a dropdown box is shown where the column used for that dimension can be changed. You can only select valid options for each dimension. For markers, you need to select string-valued columns for the ID and numeric-valued columns for the latitude and longitude. For lines, you need to select a latitude and longitude for both the source and the destination.
 
-- **Template**: Only available when the query is configured as a set of markers. Allows you to change the appearance of the markers. When you click *Edit*, an editor window opens where you can make changes to the template.
+- **Template**: Only available when the query is configured as a set of markers. Allows you to change the appearance of the markers.
+
+  - When you click *Edit*, an editor window opens where you can make changes to the template.
+
+  - From DataMiner 10.4.0 [CU13]/10.5.0 [CU1]/10.5.4 onwards<!--RN 42226-->, when you click *Browse templates*, you can reuse previously saved templates for components in the same dashboard or low-code app.
```
Robin-Debel, line 51: Previously saved is not entirely true. Any known template for each component can be seen. But not the history all templates that were ever on that component. When a template will be visible and when not anymore is a bit technical and hard to explain. I wouldn't it mention that deeply. But would change 'previously saved templates' to 'saved templates'

Robin-Debel (changes requested):
Might need to mention somewhere that we do not replace any linked columns. So when ur columns have different ids these will not resolve. This is a small detail but if we don't mention it anywhere this might come in as an issue while *for now* it is by design

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Visualizations/Available_visualizations/General/Maps.md
EdithLansens:
```suggestion
  - From DataMiner 10.4.0 [CU13]/10.5.0 [CU1]/10.5.4 onwards<!--RN 42226-->, when you click *Browse templates*, you can reuse saved templates for components in the same dashboard or low-code app.
```

user-guide/Advanced_Modules/Dashboards_and_Low_Code_Apps/Configuring_components/Template_Editor.md
```diff
@@ -411,3 +411,15 @@ If you have already configured a template for a component of the same type in th
 1. Select *Reuse*.
 
 ***
+
+In a text layer of a template, you can enter a column name surrounded by curly brackets (e.g. {CustomerName}) to insert the corresponding cell value into your text. However, when you reuse a template, these linked columns will not be automatically updated. If the component you are applying the template to has columns with different name tags, you will need to manually update the column name in the template to ensure the correct data is displayed.
```
Robin-Debel, line 415: it's not due to name tags. its actually technically the query id. If you have 2 columns that are called 'Name' they will not always map on each other. This is however something we plan to implement in the future

PR #4451, opened 2025-03-17
DCP263994: Added page Handling_Async_Code

JorgeFilipeDias:
Added page Handling_Async_Code to explain how to run async libraries from automation scripts


user-guide/Advanced_Modules/toc.yml
MariekeGO, line 1: I'm not entirely sure if this is the best location to add this info. It seems more like a best practice, so I'd be more inclined to add it on the [Automation best practices](https://docs.dataminer.services/develop/devguide/Automation/Automation_best_practices.html) page. What do you think?

slc-peter-vanpoucke: When a script is incorrect marked as async an error will now be returned, with an aka link referring to the Handling_Async_Code page. To avoid confusion, it might be better to add that new page.

Adding a link to [Automation best practices](https://docs.dataminer.services/develop/devguide/Automation/Automation_best_practices.html) page might suffice perhaps?

MariekeGO: Would it be a solution if I made it a separate page under that best practices page? 
My concern with the current suggested location for the page is that this should only contain information related to the Automation module in Cube, while this new page goes well beyond that and is really more suitable for the Developer documentation section.

slc-peter-vanpoucke: It's fine for me to move that page there.

MariekeGO (approved):

PR #4458, opened 2025-03-19
Update Migrating_the_general_database_to_Cassandra.md

robin-devos-skyline:
After receiving feedback on "Cassandra_General_DB_Failure.md", appending the tombstone threshold values for Windows here. Given "Cassandra per DMA" is even less recommended than clustered Cassandra, I just provided the essentials without full procedure.


user-guide/Advanced_Functionality/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_Cassandra.md
MariekeGO, line 1: As the default procedure here also downloads and installs Cassandra, setting the tombstones *before* you start the migration will generally be impossible. Can I move this to "After the migration"?

robin-devos-skyline: You're correct, please move to "After the migration"
I thought the migration to a clustered cassandra was using the same migrator tool..

MariekeGO (approved):

PR #4473, opened 2025-03-20
Update Collecting_data_to_report_an_issue_to_TechSupport.md

MicBroSky:
Added keywords related to DELT Export


MariekeGO:
These same words are already mentioned in the text itself, so there is no use in adding them in the keywords as well.

PR #4487, opened 2025-03-24
Add additional FAQs for DaaS

wobog:


MariekeGO:
@wobog, I've reviewed your PR and combined two pairs of questions that were very closely related, as I think this will make them more user-friendly. Could you take another look and let me know if this looks OK for you?

wobog:
Looks good combined. Thanks.

I do have a request. After reading 'Where is the data for my DaaS system hosted?' question and rereading my initial commit, I believe we should also add that 'if you want your data to be hosted in another region, please contact daas@dataminer.services'
I realize I forgot to add that.

MariekeGO (approved):

PR #4563, opened 2025-04-02
RN 42326

EdithLansens:
@lisdjong, could you please review this pull request to confirm the accurate inclusion of release note 42326 in the user guide? Thanks in advance!

### DIFF
--- a/develop/devguide/Connector/Howto/How_to_use_history_sets_on_a_protocol_parameter.md
+++ b/develop/devguide/Connector/Howto/How_to_use_history_sets_on_a_protocol_parameter.md
@@ -40,7 +40,9 @@ In order to make sure the sets are correctly stored and linked with the provided
 ```
 
 > [!NOTE]
-> When this attribute is not present, it is possible that the data is not stored correctly, and an incorrect result can be shown when the historical data is accessed, e.g. via a trend graph.
+>
+> - When this attribute is not present, it is possible that the data is not stored correctly, and an incorrect result can be shown when the historical data is accessed, e.g. via a trend graph.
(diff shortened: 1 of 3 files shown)


develop/devguide/Connector/Howto/How_to_use_history_sets_on_a_protocol_parameter.md
```diff
@@ -40,7 +40,9 @@ In order to make sure the sets are correctly stored and linked with the provided
 ```
 
 > [!NOTE]
-> When this attribute is not present, it is possible that the data is not stored correctly, and an incorrect result can be shown when the historical data is accessed, e.g. via a trend graph.
+>
+> - When this attribute is not present, it is possible that the data is not stored correctly, and an incorrect result can be shown when the historical data is accessed, e.g. via a trend graph.
+> - If you set `historySet` to "true" for a parameter that uses the [smart baseline option](xref:Configuring_dynamic_alarm_thresholds), the option will no longer have effect.
```
lisdjong, line 45: I suppose you mean the smart baseline options with 'the option', but maybe make it more clear that historySet is not 'the option'? For me it was not directly 100% clear.

EdithLansens:
```suggestion
> - If you set `historySet` to "true" for a parameter that uses the [smart baseline option](xref:Configuring_dynamic_alarm_thresholds), the smart baseline option will no longer have effect.
```

You're right. "The option" might be a bit ambiguous.

lisdjong (approved):

MariekeGO (approved):

PR #4441, opened 2025-03-17
GQI - IGQIOptimizeableDataSource life cycle for ad hoc data sources

skyline-rge:
Updated and added docs for the new IGQIOptimizeableDataSource life cycle for ad hoc data sources in the GQI DxM.
Related RN: [GQI DxM - Extensions - Allow ad hoc data sources to optimize operators](https://intranet.skyline.be/DataMiner/Lists/Release%20Notes/DispForm2.aspx?ID=42528)

PR #4581, opened 2025-04-07
RN 42468

EdithLansens:
@skyline-rge, could you please review this pull request to confirm the accurate inclusion of release note 42468 in the user guide? Thanks in advance!


skyline-rge (commented):
I'm not sure if we should at this here, because from now on, **all** new features and improvements for GQI will only be applied to the GQI DxM.
For consistency we would then have to keep adding notes here?

Until now, we just extended the existing GQI docs while mentioning that is only available when using the GQI DxM, e.g.:

> available from DataMiner 10.5.0 [CU2]/10.5.5 onwards when using the GQI DxM

For example in: [PR: GQI - IGQIOptimizeableDataSource life cycle for ad hoc data sources](https://github.com/SkylineCommunications/dataminer-docs/pull/4441/files#diff-34ecf80168cf95ed1677dc1f9b0eee5bba5c55eea12b6c56773fde91e41320c0)

skyline-rge (approved):
Huge improved, thanks!
+ a good starting point for us to add info in the future.

MariekeGO (approved):

PR #4593, opened 2025-04-08
Update DB_xml.md

LuisFreitas-Skyline:
Up for discussion because of DCP266793
If the type="MySQL" is not there, there will be errors during startup while parsing the db.xml 

Using the UI to configure it, this will not be an issue but if via file change (db.xml) could be misleading. 
@DieterDegrande


DieterDegrande (approved):
Is acceptable for me to put in the docs.

MariekeGO (approved):

PR #4600, opened 2025-04-10
remove overview parts

SkylineCookieMonster:


SkylineCookieMonster:
@MariekeGO  I would remove these parts from the overview. Cost & Billing is not part of the MediaOps package and is not scheduled on the roadmap. All other apps are linked to live and is also not available in the MediaOps package yet. This is probably only be available towards the end of this year.

MariekeGO (approved):

PR #4601, opened 2025-04-10
Update Troubleshooting_SLDMS_exe.md

LuisFreitas-Skyline:
Removed "A DMA restart is the only way to clear this RTE." There are scenarios where this may be "normal"
FYI, @bertbsky

### DIFF
--- a/user-guide/Troubleshooting/Troubleshooting_Flowcharts/Troubleshooting_Process_Identification/Communication_processes/Troubleshooting_SLDMS_exe.md
+++ b/user-guide/Troubleshooting/Troubleshooting_Flowcharts/Troubleshooting_Process_Identification/Communication_processes/Troubleshooting_SLDMS_exe.md
@@ -179,6 +179,8 @@ Check the entries near the time of the issue, find a suspicious thread and follo
 
 - A notification is a request that a process can send to SLDMS to do something, e.g. in case a file has changed. Usually SLNet sends many notifications to SLDMS.
 
+- If this RTE needs to be cleared, which is not always the case since sometimes an RTE can mean that the system is processing a large amount of data, you will need to restart the DMA. If you want to resolve the RTE without a DMA restart, you will need a memory dump.
+
 - This RTE could result in a memory leak. The NotificationThread slows down, causing notifications to stack up and not get handled, which can result in the SLDMS process going out of memory (memory leak). You might see this in the log file as "No more threads can be created".
 
 - There are also other RTEs associated with SLDMS, e.g. the ConnectionThread RTE, the SNMPThread RTE, etc. However, the NotificationThread RTE is the most frequently seen RTE.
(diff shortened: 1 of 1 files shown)


MariekeGO:
> Removed "A DMA restart is the only way to clear this RTE." There are scenarios where this may be "normal" FYI, @bertbsky

What do you mean with "There are scenarios where this may be normal"? Do you mean that there are cases where it is normal that this RTE occurs? In that case, might there still be cases where it is not normal and a user might want to clear the RTE after all? Because if so, it might be better to not just remove this info but instead add more information.

LuisFreitas-Skyline:
You raise a valid point, and I also prefer adding more information rather than simply removing it.

However, due to the way DataMiner is built, an RTE doesn't always mean that a process is stuck - it can also mean that the system is processing a large chunk of data without actively informing the user. This was the case in the use case that prompted me to remove the information.

Context:
TDF added two new DMAs to a heavily loaded DMS (TNT).

[DCP266375](https://collaboration.dataminer.services/task/266375)
_Developer Notes (DDG):
Connection and Notification threads are waiting on m_csClusters lock, which is held by the thread completing the synchronization. RTEs during DMS sync — especially during the initial addition of a DMA to a cluster — are to be expected, particularly in large clusters._

[DCP266376](https://collaboration.dataminer.services/task/266376)
_This sync started at 10:54, and by 11:50 it still hadn't finished. An RTE in SLDMS had been present since 2025-04-04 10:53:57. We have a memory dump of SLDMS — is it possible to determine why the RTE was there or where the sync was stuck?_

While the sync was still ongoing, the DMA was restarted due to the RTE in SLDMS.

As mentioned earlier, I would like to document these kinds of corner cases as well. Unfortunately, we might be missing some others. Given how an RTE in SLDMS is presented in our documentation, users could assume that a restart is the only fix - rather than first checking if the RTE is the expected behavior in that context.

I'm adding @DieterDegrande and @XanderWA because they were involved in the above tasks and to get their input here.

MariekeGO:
> You raise a valid point, and I also prefer adding more information rather than simply removing it.
> 
> However, due to the way DataMiner is built, an RTE doesn't always mean that a process is stuck - it can also mean that the system is processing a large chunk of data without actively informing the user. This was the case in the use case that prompted me to remove the information.
> 
> Context: TDF added two new DMAs to a heavily loaded DMS (TNT).
> 
> [DCP266375](https://collaboration.dataminer.services/task/266375) _Developer Notes (DDG): Connection and Notification threads are waiting on m_csClusters lock, which is held by the thread completing the synchronization. RTEs during DMS sync — especially during the initial addition of a DMA to a cluster — are to be expected, particularly in large clusters._
> 
> [DCP266376](https://collaboration.dataminer.services/task/266376) _This sync started at 10:54, and by 11:50 it still hadn't finished. An RTE in SLDMS had been present since 2025-04-04 10:53:57. We have a memory dump of SLDMS — is it possible to determine why the RTE was there or where the sync was stuck?_
> 
> While the sync was still ongoing, the DMA was restarted due to the RTE in SLDMS.
> 
> As mentioned earlier, I would like to document these kinds of corner cases as well. Unfortunately, we might be missing some others. Given how an RTE in SLDMS is presented in our documentation, users could assume that a restart is the only fix - rather than first checking if the RTE is the expected behavior in that context.
> 
> I'm adding @DieterDegrande and @XanderWA because they were involved in the above tasks and to get their input here.

Perhaps instead of removing the line, we could replace it with this?

```
- If this RTE needs to be cleared, which is not always the case since sometimes an RTE can mean that the system is processing a large amount of data, you will need to restart the DMA. If you want to resolve the RTE without a DMA restart, you will need a memory dump.
```

LuisFreitas-Skyline:
I agree with that.

MariekeGO (approved):

PR #4605, opened 2025-04-14
Updated Adding DMA to DMS: recommended to only add one agent at a time

XanderWA:


user-guide/Advanced_Functionality/DataMiner_Systems/Adding_a_DataMiner_Agent_to_a_DataMiner_System.md
```diff
@@ -21,7 +21,9 @@ Please also note the following:
 - In most cases, when you add a DataMiner Agent to a DataMiner System, all other DataMiner Agents in the DataMiner System will connect to it using its primary IP address. However, in some cases, you may need to configure the connection strings by hand. See [Connection strings](xref:Connection_strings).
 
 > [!WARNING]
-> If you add a DataMiner Agent to a DataMiner System, please make sure that it is a **new DataMiner Agent** that has not yet been put into use and that uses the **same software version** as the other Agents in the system.
+>
+> - If you add a DataMiner Agent to a DataMiner System, please make sure that it is a **new DataMiner Agent** that has not yet been put into use and that uses the **same software version** as the other Agents in the system.
+> - It's recommended to only add one DataMiner Agent at a time and to wait until it got synchronized with the cluster (indicated by the information event with parameter description "Synchronization finished") before adding another agent.
```
MariekeGO, line 26: You have now added this to a warning block, which would mean that ignoring this would have dire consequences, possibly causing the entire system to malfunction. However, from the way it is phrased as just a recommendation, I don't get the impression the consequences would be that bad. Should this indeed be part of the warning, or should it be moved to a separate note? What happens if people do not add the DMAs one at a time?

PR #4618, opened 2025-04-16
IDP content from Dojo moved

MariekeGO:
In this PR, I'm moving the information from https://community.dataminer.services/dataminer-apps/dataminer-infrastructure-discovery-and-provisioning-app/ to docs so we have all IDP information in one place. However, I left out the information about Asset Manager, as this module is [being retired](https://docs.dataminer.services/user-guide/Reference/Software_support_life_cycles.html#dataminer-functionality-evolution-and-retirement).

@jaklke, since we've changed the parent section for DataMiner IDP to "DataMiner frameworks", I was also wondering if it wouldn't be better to change the name of the IDP section to "DataMiner IDP" instead of "DataMiner IDP app". What do you think?


jaklke (approved):
I would no longer link to the product sheet as it is outdated in terms of screenshots, functionality and order info.

PR #4656, opened 2025-04-24
installer + disk failover info

wobog:
Add section on configuration in case of a failover pair

### DIFF
--- a/user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
+++ b/user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
@@ -18,7 +18,9 @@ You can follow the steps below or watch this short video, which shows you how to
 <p><br></p>
 
 > [!IMPORTANT]
-> Avoid using duplicates of existing VMs to install a new DataMiner machine. Using cloned VMs can cause certain configurations from the previous DataMiner machine to linger and cause conflicts in the system.
+>
+> - **Avoid using duplicates of existing VMs** to install a new DataMiner machine. Using cloned VMs can cause certain configurations from the previous DataMiner machine to linger and cause conflicts in the system.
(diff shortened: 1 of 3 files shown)


user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
```diff
@@ -18,7 +18,9 @@ You can follow the steps below or watch this short video, which shows you how to
 <p><br></p>
 
 > [!IMPORTANT]
-> Avoid using duplicates of existing VMs to install a new DataMiner machine. Using cloned VMs can cause certain configurations from the previous DataMiner machine to linger and cause conflicts in the system.
+>
+> - **Avoid using duplicates of existing VMs** to install a new DataMiner machine. Using cloned VMs can cause certain configurations from the previous DataMiner machine to linger and cause conflicts in the system.
+> - To restore a **backup**, install a **Failover** Agent, or install an **offline** Agent, you will need to follow a **different procedure** than shown in the video above. Please read the instructions below carefully to make sure you follow the correct procedure.
```
wobog, line 23: The first agent in a failover pair can be installed as a normal agent. In fact, it can also be an existing agent.
The procedure is only applicable if you install a 2nd agent in the failover pair, which is the Failover Agent (at that point it will be the offline agent).

MariekeGO: OK like this?
```suggestion
> - To restore a **backup**, create a **Failover** Agent to pair with an existing Agent, or install an **offline** Agent, you will need to follow a **different procedure** than shown in the video above. Please read the instructions below carefully to make sure you follow the correct procedure.
```

wobog: I would drop the "or install an **offline** Agent," as that would be redundant.

wobog: OK, now I get it. Wording it like this is much clearer yes. Perfect!

MariekeGO (approved):

PR #4671, opened 2025-04-28
RN 42716

EdithLansens:
@HansSolos, could you please review this pull request to confirm the accurate inclusion of release note 42716 in the user guide? Thanks in advance!

### DIFF
--- a/user-guide/Getting_started/Accessing_DataMiner/Accessing_DataMiner_Cube/Opening_DataMiner_Cube/Managing_the_start_window.md
+++ b/user-guide/Getting_started/Accessing_DataMiner/Accessing_DataMiner_Cube/Opening_DataMiner_Cube/Managing_the_start_window.md
@@ -75,7 +75,7 @@ To do so:
 1. Click *Save*.
 
 > [!NOTE]
-> From DataMiner 10.4.0 [CU15]/10.5.0 [CU3]/10.5.6 onwards<!--RN 42716-->, if you select *HTTP or HTTPS*, Cube will send both an HTTP request and an HTTPS request in parallel, each with a timeout of 60 seconds. The first successful response will be used, and the other request will be canceled. If the first response is empty, Cube will fall back on the other request. In earlier versions, Cube will first send an HTTP request with a default timeout of 100 seconds. If that request fails, it will then send an HTTPS request.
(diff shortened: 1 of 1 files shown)


user-guide/Getting_started/Accessing_DataMiner/Accessing_DataMiner_Cube/Opening_DataMiner_Cube/Managing_the_start_window.md
```diff
@@ -74,6 +74,9 @@ To do so:
 
 1. Click *Save*.
 
+> [!NOTE]
+> From DataMiner 10.4.0 [CU15]/10.5.0 [CU3]/10.5.6 onwards<!--RN 42716-->, if you select *HTTP or HTTPS*, Cube will send both an HTTP request and an HTTPS request in parallel, each with a timeout of 60 seconds. The first successful response will be used, and the other request will be canceled. If the first response is empty, Cube will fall back on the other request. In earlier versions, Cube will first send an HTTP request with a default timeout of 100 seconds. If that request fails, it will then send an HTTPS request.
```
HansSolos, line 78: "If the first response is empty" is more like "If the first response is not succesful"
Why? invalid or faulted response are taken into account as well, besides empty response.

HansSolos (changes requested):

user-guide/Getting_started/Accessing_DataMiner/Accessing_DataMiner_Cube/Opening_DataMiner_Cube/Managing_the_start_window.md
EdithLansens:
```suggestion
> From DataMiner 10.4.0 [CU15]/10.5.0 [CU3]/10.5.6 onwards<!--RN 42716-->, if you select *HTTP or HTTPS*, Cube will send both an HTTP request and an HTTPS request in parallel, each with a timeout of 60 seconds. The first successful response will be used, and the other request will be canceled. If the first response is not successful, Cube will fall back on the other request. In earlier versions, Cube will first send an HTTP request with a default timeout of 100 seconds. If that request fails, it will then send an HTTPS request.
```

MariekeGO (approved):

PR #4688, opened 2025-04-30
Add documentation about Wants-/WasOnFilter to allow filtering a dropdown (RN 42064)

slc-peter-vanpoucke:
Related to DCP 259353, RN 42064 & RN 42845.


slc-peter-vanpoucke:
@SkylineCommunications/team-docs-reviewers 
This change will only be available for the new dropdown UI component (RN 42845) that is getting implemented with (DCP 245210). For now, the query parameter (?useNewIASInputComponents=true) wasn't added here yet. Would you prefer me to?

MariekeGO:
> useNewIASInputComponents

Your question isn't entirely clear to me. Do you mean that ?useNewIASInputComponents=true has to be used to be able to use this new feature? If yes, we should definitely mention this to prevent confusion.

slc-peter-vanpoucke:
> Your question isn't entirely clear to me. Do you mean that ?useNewIASInputComponents=true has to be used to be able to use this new feature? If yes, we should definitely mention this to prevent confusion.

That's what I meant indeed. I wasn't sure if that info wouldn't overwhelm the reader. As discussed, added the flag and a ref to a section with extra info. Hope this is fine.

src/DataMiner/SLManagedAutomation/UIBlockDefinition.cs
```diff
@@ -918,6 +918,26 @@ public string ValidationText
...
 		public bool WantsOnChange { get; set; }
 
+		/// <summary>
+		/// Gets or sets a value indicating whether changing the filter of this dialog box item will trigger a <see cref="UIResults.WasOnFilter"/> event.
+		/// </summary>
+		/// <value><c>true</c> if changing the filter of this dialog box item will trigger a <see cref="UIResults.WasOnFilter"/> event; otherwise, <c>false</c>.</value>
+		/// <example>
+		/// <remarks>
+		/// <para>Applicable only in case <see cref="Type"/> is set to <see cref="UIBlockType.DropDown"/>.</para>
+		/// <note type="note">Available from DataMiner 10.5.7/10.6.0 onwards, in Automation scripts launched from web apps and [specifying the `useNewIASInputComponents=true` URL parameter](xref:Configuring_app_actions_and_behavior_via_URL#configuring-app-behavior-via-the-url).</note> <!-- RN 42808 / RN 42845 -->
```
MariekeGO, line 928: This cross-reference does not get rendered correctly; apparently DocFX has trouble recognizing such cross-references within the automatically generated pages. We've recently encountered a similar issue where we could fix the issue by changing the cross-reference format (#4693). Could you check if you can adjust this similarly here? If not, it might be better to just use plain text and leave out the cross-reference.

src/DataMiner/SLManagedAutomation/UIResults.cs
```diff
@@ -198,6 +198,68 @@ public string[] GetExpanded(string key)
...
+		///     filter = results.GetFilterString("FilterDropdown");
+		///   }
+		/// }
+		/// while (!results.WasButtonPressed("Close"));
+		/// </code>
+		/// </example>
+		/// <remarks>
+		/// <para>Applicable only in case <see cref="Type"/> is set to <see cref="UIBlockType.DropDown"/>.</para>
+		/// <para>For <see cref="GetFilterString"/> to work, <see cref="UIBlockDefinition.WantsOnFilter"/> has to be set to true and <see cref="WasOnFilter"/> for the current result should return true.</para>
+		/// <note type="note">Available from DataMiner 10.5.7/10.6.0 onwards, in Automation scripts launched from web apps and [specifying the `useNewIASInputComponents=true` URL parameter](xref:Configuring_app_actions_and_behavior_via_URL#configuring-app-behavior-via-the-url).</note> <!-- RN 42808 / RN 42845 -->
```
MariekeGO, line 257: Same comment as for UIBlockDefinition.cs.

src/DataMiner/SLManagedAutomation/UIResults.cs
```diff
@@ -198,6 +198,68 @@ public string[] GetExpanded(string key)
...
+		/// }
+		/// while (!results.WasButtonPressed("Close"));
+		/// </code>
+		/// </example>
+		/// <remarks>
+		/// <para>Applicable only in case <see cref="Type"/> is set to <see cref="UIBlockType.DropDown"/>.</para>
+		/// <para>For <see cref="GetFilterString"/> to work, <see cref="UIBlockDefinition.WantsOnFilter"/> has to be set to true and <see cref="WasOnFilter"/> for the current result should return true.</para>
+		/// <note type="note">Available from DataMiner 10.5.8/10.6.0 onwards, in Automation scripts launched from web apps and <see href="xref:Configuring_app_actions_and_behavior_via_URL#configuring-app-behavior-via-the-url">specifying the useNewIASInputComponents=true URL parameter</see>.</note> <!-- RN 42808 / RN 42845 -->
+		/// <note type="important">When the options are filtered, the currently selected option (if still relevant) needs to be inserted as an option, regardless of whether the filter matches. Otherwise, the dropdown will consider that value incorrect, and the dropdown will be cleared.</note>
+		/// <note type="tip">Consider only filtering the display value of the options and, if possible, case invariant.</note>
```
MariekeGO, line 259: Could you clarify how "case invariant" is linked grammatically to the rest of the sentence? Something seems off about the sentence, but I'm not sure how best to adjust it. (same for line 498)

slc-peter-vanpoucke: The string GetFilterString() returns, is used to filter out a list of options display in the list of a dropdown. Considering 'fra' as the filter string, my suggestion would be to also consider 'France' as a correct option. That won't be the case when using String.Contains in C# by default.

MariekeGO: Would this be an acceptable alternative?
```suggestion
		/// <note type="tip">Consider only filtering the display value of the options and, if possible, ensure that the filter is not case-sensitive.</note>
```

MariekeGO (approved):

PR #4690, opened 2025-04-30
Update Configuring_a_page_to_update_parameters_on_session_variable_ch…

jarno-lernou:
…anges.md

### DIFF
--- a/user-guide/Basic_Functionality/Visio/session_variables/Configuring_a_page_to_update_parameters_on_session_variable_changes.md
+++ b/user-guide/Basic_Functionality/Visio/session_variables/Configuring_a_page_to_update_parameters_on_session_variable_changes.md
@@ -7,8 +7,7 @@ uid: Configuring_a_page_to_update_parameters_on_session_variable_changes
 It is possible to have parameters updated on an open Visual Overview page when a session variable changes.
 
 > [!TIP]
-> See also:
-> [Linking a shape to a SET command](xref:Linking_a_shape_to_a_SET_command)
+> See also: [Linking a shape to a SET command](xref:Linking_a_shape_to_a_SET_command)
 
 ## Configuring the shape data field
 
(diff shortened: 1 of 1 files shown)


user-guide/Basic_Functionality/Visio/session_variables/Configuring_a_page_to_update_parameters_on_session_variable_changes.md
```diff
@@ -43,3 +43,7 @@ Data =[var:LoadTime]|SetTrigger=ValueChanged-
 Set|218/652|1203|[var:Chain]|SetTrigger=ValueChanged-
 Set|218/652|1205|[var:LoadTime]|SetTrigger=ValueChanged
 ```
+
+> [!TIP]
+> Change by default separators to avoid having problems with dynamic values (such as Element and View names or properties).
```
MariekeGO, line 48: OK to change this like this?
```suggestion
> To avoid problems with dynamic values (such as element and view names or properties), you can change the default separators.
```

jarno-lernou: I believe many mistakes have been made with separators and therefore I wanted to **politely push people** (i.e. can -> should) to change the separators by default to use a separators that are either exceptional/exotic  (e.g. '§', '@' or 'ùµ' (may be multiple characters))  or forbidden in the dynamic values ([Forbidden characters | DataMiner Docs](https://docs.dataminer.services/user-guide/Reference/Naming_Conventions/Forbidden_characters.html?q=forbidden)) (e.g. '?').

MariekeGO: In that case, how about this alternative?
```suggestion
> To avoid problems with dynamic values (such as element and view names or properties), we highly recommend changing the default separators.
```

MariekeGO: OK, how about this? 
```suggestion
> To avoid problems with dynamic values (such as element and view names or properties), we highly recommend changing the default separators. Ideally, you should use separators that are exceptional or combine multiple characters, so that they definitely will not occur in the dynamic values, such as `§`, `@`, or `ùµ`, or [forbidden characters](xref:Forbidden_characters) such as `?`.
```

MariekeGO:
I've moved your tip up and changed it to a note to make sure people will definitely see it. However, I was also wondering, wouldn't it be better if we also removed the current example at the bottom of the page, given that this doesn't follow this new recommendation?

jarno-lernou:
I agree, we should either adapt or remove it. After all instead of [var:], we can/should typically use [cardvar:] as well.
Ideally, we should give the user more context that relates to the example, i.e. textual description: "We have a card variabel "X" with value "Y", ..., the result of the example will be "Y...", but that is maybe overkill.

MariekeGO:
> I agree, we should either adapt or remove it. After all instead of [var:], we can/should typically use [cardvar:] as well. Ideally, we should give the user more context that relates to the example, i.e. textual description: "We have a card variabel "X" with value "Y", ..., the result of the example will be "Y...", but that is maybe overkill.

I've now removed the example so we can already merge this; we can always see about adding a new example later.

MariekeGO (approved):

PR #4694, opened 2025-05-02
CICD - Update info about new versions of .NET tools

MichielOda:
With the latest version of the upload & deploy tools, the organization key is now required (with the exception for volatile upload/deploy). In addition, certain calls that relied on a system key have been removed.

### DIFF
--- a/develop/CICD/GitHub/GitHub_Secrets.md
+++ b/develop/CICD/GitHub/GitHub_Secrets.md
@@ -49,7 +49,7 @@ In the [Admin app](xref:About_the_Admin_app), two different types of keys can be
   - Name: DATAMINER_DEPLOY_KEY
   - Value: See [Managing dataminer.services keys](xref:Managing_dataminer_services_keys).
 
-- A dataminer.services **organization key** is scoped to the specific organization for which it was created and can  be used to perform uploads to and deploy from the Catalog. This includes uploading private artifacts to the Catalog that will only be accessible for the organization that your DMS belongs to. To upload and deploy items, you will need this key.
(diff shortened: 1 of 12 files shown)


develop/CICD/GitHub/GitHub_Secrets.md
```diff
@@ -49,7 +49,7 @@ In the [Admin app](xref:About_the_Admin_app), two different types of keys can be
   - Name: DATAMINER_DEPLOY_KEY
   - Value: See [Managing dataminer.services keys](xref:Managing_dataminer_services_keys).
 
-- A dataminer.services **organization key** is scoped to the specific organization for which it was created and can **only be used to perform uploads to the Catalog**. This includes uploading private artifacts to the Catalog that will only be accessible for the organization that your DMS belongs to. To upload non-connector items, you will need this key.
+- A dataminer.services **organization key** is scoped to the specific organization for which it was created and can be used to upload to and deploy from the Catalog. This includes uploading private artifacts to the Catalog that will only be accessible for the organization that your DMS belongs to. To upload and deploy items, you will need this key.
```
MariekeGO, line 52: Line 47 says that to deploy connectors, you need DATAMINER_DEPLOY_KEY, but now this line says you can use DATAMINER_TOKEN for any items, so I assume this also means connectors - which is a little confusing. Is the text above about the system key still correct? If yes, could you clarify a bit more when you need to use which key?

MichielOda: Changed the text of the system key as that one becomes more and more deprecated. For now it is only used for the volatile upload/deploy. Eventually this will also become deprecated

MariekeGO (approved):

PR #4701, opened 2025-05-05
Update Protocol.Params.Param-historySet.md

LuisFreitas-Skyline:
Added alarm behavior with historySet.
@AnaCatarinaGrilo @MariaCUSkyline

### DIFF
--- a/develop/schemadoc/Protocol/Protocol.Params.Param-historySet.md
+++ b/develop/schemadoc/Protocol/Protocol.Params.Param-historySet.md
@@ -28,7 +28,7 @@ If you mark a parameter as a history set parameter, its last set value will not 
 >
 > - Using history sets in combination with [alarm hysteresis](xref:Configuring_alarm_hysteresis) on the same parameter can cause unexpected behavior in alarm timestamps, as both features modify the parameter and alarm times. For this reason, we do not recommend combining them.
 > - The [smart baseline option](xref:Configuring_dynamic_alarm_thresholds) is incompatible with history sets. If a parameter has history sets enabled, enabling smart baselines will have no effect. From DataMiner 10.4.0 [CU14]/10.5.0 [CU2]/10.5.5 onwards<!--RN 42326-->, you will receive a warning message when attempting to enable smart baselines for a parameter with history sets enabled.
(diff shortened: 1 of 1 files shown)


develop/schemadoc/Protocol/Protocol.Params.Param-historySet.md
```diff
@@ -28,6 +28,7 @@ If you mark a parameter as a history set parameter, its last set value will not
 >
 > - Using history sets in combination with [alarm hysteresis](xref:Configuring_alarm_hysteresis) on the same parameter can cause unexpected behavior in alarm timestamps, as both features modify the parameter and alarm times. For this reason, we do not recommend combining them.
 > - The [smart baseline option](xref:Configuring_dynamic_alarm_thresholds) is incompatible with history sets. If a parameter has history sets enabled, enabling smart baselines will have no effect. From DataMiner 10.4.0 [CU14]/10.5.0 [CU2]/10.5.5 onwards<!--RN 42326-->, you will receive a warning message when attempting to enable smart baselines for a parameter with history sets enabled.
+> - When *historySet* is enabled, DataMiner determines the time of the relevant alarms based on when the parameter value is updated, rather than when the alarms themselves are updated (e.g. properties changed, view impact changed, acknowledged, etc.).
```
MariekeGO, line 31: It's not entirely clear to me what the examples in parentheses are for. Could you elaborate this a little?

LuisFreitas-Skyline: Those are alarm types, as you can see below:
image

MariekeGO: Indeed, but that still doesn't make it clear. Do you mean that the time of the alarm depends only on when the parameter value is updated, and not on when the alarm type changes?

```suggestion
> - When *historySet* is enabled, DataMiner determines the time of the relevant alarms based on when the parameter value is updated, rather than when the alarms themselves are updated (e.g. when alarm properties change, the view impact changes, the alarm is acknowledged, etc.).
```

MariekeGO (approved):

PR #4702, opened 2025-05-05
Web apps troubleshooting flowchart

MariekeGO:


Robin-Debel:
> @ToonCasteele, I've made a draft flowchart based on your input; however, I'm not sure where to fit in the "weird feeding behavior" part of the file you provided. Could you provide some more details as to what this behavior entails? I'm thinking of possibly adding this on the [Troubleshooting - web](https://docs.dataminer.services/user-guide/Troubleshooting/Procedures/Investigating_Web_Issues.html) page instead of in the flowchart. Could you also take a look if everything that is in the draft flowchart now makes sense and if anything should still be changed/added? This is what it currently looks like:
> 
> image

The main reason for this weird feeding behaviour section is because of a known issue we have within our software that results in what we call 'ghost components' these are components that are in the json & are known by the framework itself, but are not visible to any user. Doing CTRL A will result in the sum of component to be deleted not adding up when you have a ghost component. Now, for me personaly it makes sense for it to be between  these 2:

image

MariekeGO:
@ToonCasteele, @Robin-Debel, I've updated the flowchart based on your input. Could you take a look and let me know if any changes are still needed? This is what it now looks like:  
image
image

PR #4707, opened 2025-05-06
Added RN 42406 to components page

MariekeGO:
@alenskyline, could you check if the dashboards from your RN 42406 are correctly added to the components overview with this PR, or if changes are needed?


alenskyline (approved):
Sorry for the late response, looks good

PR #4729, opened 2025-05-09
Update Protocol.Params.Param-setter.md

MichielClepkensSkyline:
Added a use case where a write parameter is linked to a trigger and a QAction.


develop/schemadoc/Protocol/Protocol.Params.Param-setter.md
```diff
@@ -6,6 +6,12 @@ uid: Protocol.Params.Param-setter
 
 Specifies whether the value of the write parameter will be copied to the corresponding read parameter (without the need to add a trigger or an action).
 
+For example, if a read-write parameter combination uses the setter=true option, a trigger goes off when the value of the write parameter changes, and the write parameter also triggers a QAction, the order of execution is the following:
```
PedroDebevere, line 9:
```suggestion
Refer to [Parameter change events](xref:LogicParameters#parameter-change-events) for more information on when the set is executed in relation to other operations. For example, if a read-write parameter combination uses the setter=true option, a trigger goes off when the value of the write parameter changes, and the write parameter also triggers a QAction, the order of execution is the following:
```

develop/schemadoc/Protocol/Protocol.Params.Param-setter.md
```diff
@@ -6,6 +6,12 @@ uid: Protocol.Params.Param-setter
...
+
+1. The regular triggers on the write parameter are executed. Actions executed by these triggers will not yet see the read parameter value adapted to the new write parameter value. The read parameter will still have the old value at this point.
+1. The copy action is executed to the read parameter.
```
PedroDebevere, line 12:
```suggestion
1. The setter is executed, which results in the value being copied to the read parameter.
```

MariekeGO (approved):

PR #4734, opened 2025-05-12
KI: user locked out after failed login attempt

MariekeGO:


LuisFreitas-Skyline (approved):

user-guide/Troubleshooting/Known_issues/KI_user_locked_out_after_single_failed_attempt.md
```diff
@@ -0,0 +1,33 @@
...
+
+# User locked out after single login attempt
+
+## Affected versions
+
+From DataMiner 9.5.0 [CU12]/9.6.3 onwards.<!-- RN 20426 -->
+
+## Cause
+
+A failed login attempt will result in an additional attempt by SLNet in order to check if the password was expired. In addition, Cube will automatically and silently try to log in at startup with saved credentials, which will also lead to four additional login attempts in case the password had been changed. Cube will also try to automatically reconnect every second in the background under certain conditions. As a result of this, if the security policy only allows a small number of login attempts (e.g. 5), one failed attempt can already cause a user to be locked out.
```
bertbsky, line 13: > Cube will also try to automatically reconnect every second in the background under certain conditions.

I would not mention that here as it is not desired behavior but a bug that needs fixing (DCP269613).

bertbsky, line 13: > A failed login attempt **with a Windows user** ...

I would clarify that this only applies to Windows users (local or domain), not when using any form of external authentication (SAML, RADIUS, CROWD).

bertbsky (changes requested):

user-guide/Troubleshooting/Known_issues/KI_user_locked_out_after_single_failed_attempt.md
MariekeGO: As this is a known issue page, I'd be inclined to keep this info but instead make it more clear that this is unwanted behavior:
```suggestion
A failed login attempt will result in an additional attempt by SLNet in order to check if the password was expired, and Cube will also automatically and silently try to log in at startup with saved credentials, which will also lead to four additional login attempts in case the password had been changed. In addition, because of an issue in the software, Cube will also try to automatically reconnect every second in the background under certain conditions. As a result of this, if the security policy only allows a small number of login attempts (e.g. 5), one failed attempt can already cause a user to be locked out.
```

MariekeGO: I've added a sentence below this paragraph to clarify this.

bertbsky (approved):

PR #4742, opened 2025-05-14
DCP261884: Add memory file support in scheduler templates

AndreRodriguesSkyline:
RN 42904

### DIFF
--- a/user-guide/Advanced_Modules/scheduler/Creating_a_Scheduler_template.md
+++ b/user-guide/Advanced_Modules/scheduler/Creating_a_Scheduler_template.md
@@ -29,17 +29,23 @@ To create scheduler templates, which will show up in the EVENT list in Scheduler
 
      Example: To define an Event Profile called “MyPresets” containing a preset value for two parameters called “From” and “To” and one dummy called “MyDummy”, the following comment is added: *Preset="MyPresets" From="a" To="b" MyDummy="34/2"*. In this example, the dummy value is “34/2”, meaning Element ID 2 of DMA ID 34.
 
+     > [!NOTE]
+     > From DataMiner 10.6.0/10.5.8 onwards, you can also use `[Memory file name]="value"` when defining a profile to specify the value of a memory file.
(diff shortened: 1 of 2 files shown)


user-guide/Advanced_Modules/scheduler/Creating_a_Scheduler_template.md
```diff
@@ -29,18 +29,24 @@ To create scheduler templates, which will show up in the EVENT list in Scheduler
 
      Example: To define an Event Profile called “MyPresets” containing a preset value for two parameters called “From” and “To” and one dummy called “MyDummy”, the following comment is added: *Preset="MyPresets" From="a" To="b" MyDummy="34/2"*. In this example, the dummy value is “34/2”, meaning Element ID 2 of DMA ID 34.
 
+     > [!NOTE]
+     > From DataMiner 10.6.0/10.5.8 onwards, you can also use `[Memory file name]="value"` when defining a profile to specify the value of a memory file.
+     > For example, by adding `MyMemory="config"` to the comment, the memory file called "MyMemory" will be set to the value “config” for the memory file name.
```
MariekeGO, line 34: This isn't entirely clear to me, as both "MyMemory" and "config" seem to refer to a memory file name according to this explanation. Could you clarify this a bit so the difference between the two is more clear?

AndreRodriguesSkyline: You're right. Apologies for the confusion.
I used "name" because the value of a memory file refers to the name of the file that represents that memory file.
It might be clearer if I say it like this:
 From DataMiner 10.6.0/10.5.8 onwards, you can also use `[Memory file]="value"` when defining a profile to specify the value of a memory file.
 For example, by adding `MyMemory="config"`  to the comment, the memory file named "MyMemory" will be set to the value "config".
 
 (By the way, please don't merge this yet. The task is still in development)

MariekeGO (approved):

PR #4748, opened 2025-05-15
Update Installing_DM_using_the_DM_Installer.md

robin-devos-skyline:
Adding the importance of using a static IP

### DIFF
--- a/user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
+++ b/user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
@@ -47,6 +47,10 @@ You can follow the steps below or watch this short video, which shows you how to
 > [!NOTE]
 > The installation requires that [Npcap](https://nmap.org/npcap/) or WinPcap (deprecated) is installed for systems intended for DataMiner Failover based on virtual IP. If you intend to configure a [Failover setup based on hostname](xref:Failover_configuration_in_Cube), this software will not be needed.
 
+> [!IMPORTANT]
+>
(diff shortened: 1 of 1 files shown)


user-guide/Getting_started/Installing_a_DMA/Installing_DM_using_the_DM_Installer.md
```diff
@@ -47,6 +47,10 @@ You can follow the steps below or watch this short video, which shows you how to
 > [!NOTE]
 > The installation requires that [Npcap](https://nmap.org/npcap/) or WinPcap (deprecated) is installed for systems intended for DataMiner Failover based on virtual IP. If you intend to configure a [Failover setup based on hostname](xref:Failover_configuration_in_Cube), this software will not be needed.
 
+> [!IMPORTANT]
+>
+> DataMiner requires a **static IP** to be configured. Make sure to do this before continuing with the below steps. If you have to change the IP afterwards, you can do so by following the steps described in [Changing the IP of a DMA](xref:Changing_the_IP_of_a_DMA).
```
MariekeGO, line 52: As this procedure will also be seen by first-time users without too much background knowledge who just want to give DataMiner a quick try, and we don't want to scare them off or confuse them, I think it would be better if we added this as a step in the procedure above instead of an intimidating "important" box. We should also link to more info about how they can make sure a static IP is configured, because many of those first-time users won't even know how to begin with this. I'd also leave out the link to the "changing the IP" procedure, as this also adds unnecessary complexity that is likely to scare people off at this point.

robin-devos-skyline: Excluding the link is ok and adding it in the procedure above is fine for me, as long as it's in there. 
Otherwise, if they restart the PC/server, Dataminer won't work anymore.

MariekeGO (approved):

PR #4749, opened 2025-05-15
Update MaintenanceSettings_xml.md

HansSolos:
Puttin docs code into production result in DataMiner not starting up because of DataGateway crashes. Source: https://dumpserver.skyline.local/Result/2025%5C05%5C15%5CSLDataGateway.exe_11764_2025_05_15_16_08_50.high-b2108f9541394328baa94f289f9d6a02

### DIFF
--- a/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
+++ b/user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
@@ -57,8 +57,6 @@ This is an example of a *MaintenanceSettings.xml* file:
   <Trending>
     <EDCurves></EDCurves>
     <SDCurves></SDCurves>
-    <!-- CAUTION: SLDataGateway crashes on startup when MaintenanceSettings.xml contains <TimeSpan1DayRecords window="0"> -->
-    <TimeSpan1DayRecords window="0" />
     <TimeSpan1HourRecords window="60" />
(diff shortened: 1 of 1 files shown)


user-guide/Reference/Skyline_DataMiner_Folder/More_information_on_certain_files_and_folders/MaintenanceSettings_xml.md
```diff
@@ -57,6 +57,7 @@ This is an example of a *MaintenanceSettings.xml* file:
   <Trending>
     <EDCurves></EDCurves>
     <SDCurves></SDCurves>
+    <!-- CAUTION: SLDataGateway crashes on startup when MaintenanceSettings.xml contains <TimeSpan1DayRecords window="0"> -->
```
MariekeGO, line 60: Wouldn't it be better if we updated the example to a different value instead of adding this caution?

HansSolos: Fine by me, howeever:
- this is not enbaled by default, so user should not assume this is added by default
- it should then document that "0" is not a valid value

MariekeGO: I've updated your pull request. Could you check if everything still looks OK to you?

HansSolos (commented):
LGTM

MariekeGO (approved):

PR #4752, opened 2025-05-15
Update Using_a_pre_installed_DataMiner_Virtual_Hard_Disk.md

robin-devos-skyline:
remove the important note on "use static IP" and insert it into the procedure when starting up the VM

### DIFF
--- a/user-guide/Getting_started/Installing_a_DMA/Using_a_pre_installed_DataMiner_Virtual_Hard_Disk.md
+++ b/user-guide/Getting_started/Installing_a_DMA/Using_a_pre_installed_DataMiner_Virtual_Hard_Disk.md
@@ -69,7 +69,9 @@ When you have downloaded the virtual hard disk, you can start to create a VM in 
 
 1. When the login screen is shown, log in to the VM.
 
-1. Verify via the network configuration that the network interface uses a static IP instead of DHCP, because DataMiner requires a static IP.
+1. Verify in the network configuration that the network interface uses a static IP instead of DHCP, because DataMiner requires a static IP.
+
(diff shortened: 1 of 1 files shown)


user-guide/Getting_started/Installing_a_DMA/Using_a_pre_installed_DataMiner_Virtual_Hard_Disk.md
```diff
@@ -69,6 +69,8 @@ When you have downloaded the virtual hard disk, you can start to create a VM in
 
 1. When the login screen is shown, log in to the VM.
 
+1. Verify via the network configuration that the network interface uses a static IP instead of DHCP, because DataMiner requires a static IP.
```
MariekeGO, line 72: Where can users find this network configuration? We'll need to be a bit detailed here to make sure first-time users don't give up immediately.

MariekeGO: Possibly we could refer to this page: https://support.microsoft.com/en-us/windows/essential-network-settings-and-tasks-in-windows-f21a9bbc-c582-55cd-35e0-73431160a1b9? (The section "Change TCP/IP Settings" seems to refer to this, if I'm not mistaken.)

robin-devos-skyline: Indeed, section "Change TCP/IP Settings" does mention how to change from DHCP to a static IP

MariekeGO (approved):

PR #4753, opened 2025-05-15
added raw md code for easy copying when creating a new marketing/tech page

SkylineGBO:
I am not sure if duplicating the page is best, with having the raw MD data in the bottom, or just making the whole page the MD data. It has just been a little annoying recently having to go to the repo or edit the page to copy the MD template. Open to ideas on best way to solve this slight inconvenience. (Another solution would be to add the template to DIS, but not sure how long that might take).


MariekeGO:
I like the idea of having the template in markdown ready for easy copy-paste. However, the duplication is indeed a bit odd, so I've now gone for a markdown version only instead.

MariekeGO (approved):

PR #4767, opened 2025-05-20
Add placeHolders to email action in automation

PieterCarlu1:


user-guide/Advanced_Modules/Automation_module/Automation_script_actions/Email.md
MariekeGO, line 1: @PieterCarlu1, I've made some changes to make the page a bit more user-friendly and easily fit in your addition. However, I was wondering, could it be that the info about "adding required elements, parameters, and so on" (line 20) for a report/dashboard is actually outdated and refers to the legacy reports only?

PieterCarlu1: Hi Marieke, it is indeed correct that line 20 is for the legacy reporter. Only the configure button is for the new Dashboards. However both of the reporters are still supported

MariekeGO: I've now made some more changes to make sure the difference in handling dashboards and legacy reports is more clear. Could you take a look and confirm if everything still looks OK for you?

MariekeGO (approved):

PR #4802, opened 2025-05-30
Update Connecting_to_an_element_using_telnet.md

LaurensMT:
Added more info on how to configure an element to allow a telnet connection

### DIFF
--- a/user-guide/Troubleshooting/Logging/Monitoring_real-time_communication/Connecting_to_an_element_using_telnet.md
+++ b/user-guide/Troubleshooting/Logging/Monitoring_real-time_communication/Connecting_to_an_element_using_telnet.md
@@ -4,7 +4,7 @@ uid: Connecting_to_an_element_using_telnet
 
 # Connecting to an element using telnet
 
-1. Choose *Start* > *Run*
+1. In Windows, open the *Run* window (e.g. by pressing Windows key + R).
 
 1. Enter *telnet*, followed by the virtual IP address of the element.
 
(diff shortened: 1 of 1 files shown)


user-guide/Troubleshooting/Logging/Monitoring_real-time_communication/Connecting_to_an_element_using_telnet.md
```diff
@@ -4,6 +4,14 @@ uid: Connecting_to_an_element_using_telnet
 
 # Connecting to an element using telnet
 
-1. Choose *Start \> Run...*
+1. Choose *Start* > *Run*
```
MariekeGO, line 7: Now that I'm taking a closer look at this page, this instruction really seems very vague. Where is it that the user should select this? In the element context menu in the Surveyor?

LaurensMT: That is the explanation on how to do this from Windows. "Start" is clicking the start button of Windows, then typing "Run" and in the next window that is opened typing "telnet " followed by the virtual IP of the element. As this is the client side this is like surfing to a website, which can be done from various operating systems like Linux, Mac, or Windows.

MariekeGO: I've now changed this so it's clear that the instruction is for a Windows system. OK like this?

LaurensMT: Yes, with one small remark that there is a space between "telnet" and the IP address.

MariekeGO (approved):

PR #4810, opened 2025-06-02
Update Regarding_antivirus_software.md to include a table of DataMiner Process paths

MicBroSky:
I added a table listing the paths of all the DataMiner processes currently being used based on DataMiner 10.5.5.0-15690. We had a customer that wasn't able to exclude processes through a wildcard and needed exact paths. If there's another way to have this documentation up on here or if details on which versions of DataMiner and on have these processes, let me know so I can make the changes.


MariekeGO:
> I fear the maintainability of this list. Whilst this list does cover the active services and DxMs and is up to date to the latest, there's a whole host of smaller executables and scripts scattered throughout the C:\Skyline DataMiner directory (i.e. dcomConfig.exe which most recently was picked up by various AV software.). Along with any executables that scripts/connectors/... may depend on. These are not covered in this list, but adding all of them to this list would make it very extensive.

I've adjusted the pull request so it now clearly mentions that the list is not exhaustive. Does it look acceptable to you now? Note that I also changed the order of the items in the list, as the last item under the table was otherwise very likely to get overlooked.

PR #4816, opened 2025-06-03
Diminishing the size of alarm trees

MariekeGO:
@SLC-Laurens-Vercruysse, this adds your information about diminishing the size of alarm trees. To add this, I've changed the former guidelines for assigning alarm severity levels to a "best practices" page, containing stricter guidelines than previously, including the ones for alarm trees. The guidelines that aren't strictly for alarm templates are included there so we can have them all in one place, but they are also mentioned where relevant on other pages. I've also adjusted the "alarm linking" page, as it seemed more logical to name this "alarm trees" — this will hopefully help users find more quickly what we mean when we refer to alarm trees.


SLC-Laurens-Vercruysse (changes requested):
Looks good in general, but I think I spotted a small mistake where a negation was forgotten. 
Other than that, elegantly put together!

PR #4825, opened 2025-06-04
Update Smart_Serial_Connection.md

LaurensMT:
Adapted the note with accepted ip address. This has always been enabled by default.

### DIFF
--- a/user-guide/Basic_Functionality/Elements/Adding_Elements/Smart_Serial_Connection.md
+++ b/user-guide/Basic_Functionality/Elements/Adding_Elements/Smart_Serial_Connection.md
@@ -22,7 +22,7 @@ For smart-serial connections, you can specify the following connection settings 
   To add another accepted IP address, below the box, click *Add*.
 
   > [!NOTE]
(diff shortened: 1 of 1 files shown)


user-guide/Basic_Functionality/Elements/Adding_Elements/Smart_Serial_Connection.md
```diff
@@ -22,7 +22,7 @@ For smart-serial connections, you can specify the following connection settings
   To add another accepted IP address, below the box, click *Add*.
 
   > [!NOTE]
-  > This functionality is only available if *AllowedIPAddresses.Disabled* is set to “false” in the user settings of the smart-serial connection in the protocol.
+  > This functionality can be disabled if *AllowedIPAddresses.Disabled* is set to “true” in the user settings of the smart-serial connection in the protocol.
```
MariekeGO, line 25: The current phrasing suggests that it is possible to disable the functionality if that setting is set to true, but it isn't necessarily disabled based on the setting. If this isn't what you meant, I'd suggest rephrasing this like this:
```suggestion
  > This functionality can be disabled by setting *AllowedIPAddresses.Disabled* to “true” in the user settings of the smart-serial connection in the protocol.
```

PR #4830, opened 2025-06-05
Update Cassandra_expired_blocking_SSTables.md

SLC-Laurens-Vercruysse:
After executing the procedure, noticed it would overwrite the "compaction_window_size" and "compaction_window_unit" parameters. Since these are quite important, decided to update docs.


MariekeGO:
@SLC-Laurens-Vercruysse, I've reviewed your PR and made some changes, also putting some text in bold to make the page easier to scan for a user. Could you check if everything still looks OK for you?

PR #4845, opened 2025-06-12
Cloud - Release 2025 06 12

brayunm:


release-notes/Cloud/dataminer_services_change_log_2025.md
```diff
@@ -9,6 +9,71 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+The lifetime of both the access token and refresh tokens have been reduced to make these more secure.
+
+When a user is unauthenticated, the URL will stay it  as is, but the user will see an unauthenticated view. This way, users can share and refresh links even when they are signed out.
+
+### 12 June 2025 - Admin - Add user popup, dropdown
+
+This change applies to Admin users.
+Changing the add organization user and add DMS user into a popup.
+Changing the DMS user to a dropdown. This dropdown consists of organization users that are not in the current DMS yet.
```
MariekeGO, line 24: This description is not clear enough; could you change this to a more extensive description with full sentences, so users will easily be able to understand what you mean with this?

release-notes/Cloud/dataminer_services_change_log_2025.md
```diff
@@ -9,6 +9,71 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+### 12 June 2025 - Enhancement - Account Linking - Obsolete terminology updated
+
+When a user links their DataMiner account to their dataminer.services account, the page will now mention "dataminer.services account" instead of the obsolete term "DCP account".
+
+### 12 June 2025 - Enhancement - Catalog - Documentation improvements + cache notifier removal
+
+Removed the new version notifications.
```
MariekeGO, line 32: Could you clarify what you mean with this? What exactly has been removed and why?

release-notes/Cloud/dataminer_services_change_log_2025.md
```diff
@@ -9,6 +9,71 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+Up till now, it was not possible to scroll through the shared items list, making it impossible to see the items towards the end of the list in case it contained a lot of items. Now scrolling will be possible.
+
+### 12 June 2025 - Enhancement - Catalog - Improved alignment of featured sections on Catalog home page
+
+The alignment of the featured sections on the Catalog home page has been improved to account for smaller screens. The page will now switch between one and three columns when necessary.
+
+### 12 June 2025 - Enhancement - Admin - Improve unreachable nodes on DMS overview page
+
+Before this change the amount of unreachable nodes that could be shown was unlimited.
```
MariekeGO, line 65: What do you mean with "the amount was unlimited"? Could you clarify?

release-notes/Cloud/dataminer_services_change_log_2025.md
```diff
@@ -9,6 +9,71 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+With this change the user will now see the amount of unreachable nodes and when hovered over the exact names of the unreachable nodes.
+
+The unreachable nodes row has also been moved to the Connection section.
+
+### 12 June 2025 - Fix - Catalog - Tags not updating after switching details page
+
+In some cases, it could occur that when you went from one Catalog page to another, the tags of the second Catalog item were not correctly displayed.
+
+Fixed a visibility issue where the aside of a private item would still be visible when fe. changing organization.
+The aside will now properly be hidden when an error occurred regarding the item fetch
```
MariekeGO, line 75: What do you mean with "the aside"? Could you rephrase this in terms users will easily understand?

MariekeGO (changes requested):

release-notes/Cloud/dataminer_services_change_log_2025.md
brayunm: In the catalog app, every time we released a new version the user would receive a notification when they were running an outdated version. After a reload of the page another notification would be shown to indicate the user was upgraded to the latest version.

We originally introduced this when we would have issues with caching and users would not be upgraded to the latest version, this helped with clarity because it would also notify users if the upgrade failed.

Since the original implementation we haven't received any reports that this was still causing issues.
Hence, we removed it since it was also more annoying than helpful

release-notes/Cloud/dataminer_services_change_log_2025.md
brayunm: Before it would list every single node that would be unreachable e.g.
> node 1
> node 2
> node 3
> ...

Now it will just mention how many unreachable nodes there are
> 3 nodes unreachable

release-notes/Cloud/dataminer_services_change_log_2025.md
brayunm: I think the "side panel" is a better word for it.
It the smaller section on the right side of the page where more information is shown such as the publisher, element type, owners, extra links, ...

brayunm:
LGTM

EdithLansens (approved):

PR #4855, opened 2025-06-16
Removed staas@dataminer.services

MariekeGO:


ToonCasteele (approved):
Just want to remark that I would still prefer us to clean up these references to any support mail address throughout docs. Reason: It should be clear in a central location (footer?) that all support requests can go to this mail address. We should not need to clutter the docs by mentioning it everywhere separately.

PR #4867, opened 2025-06-17
Update Creating_an_information_template.md To Include Description Details When Renaming Table Column

MicBroSky:
We ran into an issue recently where a table that had the rest of its columns include the table name in parenthesis for the Description (Ex: <Description>Primary Key (Information Table)</Description>), you'll need to make sure to keep that table name for your rename in the information template unless the rest of the columns will start to display the Parenthesis section.


MariekeGO:
@MicBroSky, I've reviewed your pull request and made some changes. Could you check if everything looks OK for you? Also, I quickly tested this on my DMA and didn't encounter this issue when I changed the description for a table column. Is this an issue that only occurs in specific cases?

MariekeGO (approved):

PR #4891, opened 2025-06-20
Update Migration/Replication for STaaS

joaobarraca:


user-guide/Advanced_Functionality/Databases/STaaS/STaaS_cost_estimation.md
```diff
@@ -18,31 +18,24 @@ To request a cost estimation, follow the procedure below:
...
-      This way you will have the most accurate estimation of your system when running the script.
+   1. Locate the *CloudStorageMigration* script and [execute the script](xref:Manually_executing_a_script).
 
-   1. Click *Start migration*.
+   1. Click *Stop Migration/Replication* to stop the test run.
 
-1. Let the script run for 24 hours without restarting the DataMiner System (DMS).
-
-1. After the 24-hour period, restart the DMS to stop the estimation process.
+      1. Optionally, you can also restart your DMS.
```
MariekeGO, line 35: What would be the purpose of restarting the DMS here? Most users will want to avoid this, so unless there's a specific purpose, I'd leave this out. But if there is a specific purpose, we should definitely mention it.

joaobarraca: after the review on friday, we should remove the Optional line regarding the restart @MariekeGO

MariekeGO (approved):

PR #4937, opened 2025-06-30
Update documentation of http parameter.

BramDevlaminck-Skyline:
The updated documentation reflects that the parameter of the provided pid should be of rawtype other and type string.
Otherwise, the value is not read out of the parameter to be put in the http parameter.

### DIFF
--- a/develop/devguide/Connector/ConnectionsHttpImplementing.md
+++ b/develop/devguide/Connector/ConnectionsHttpImplementing.md
@@ -36,21 +36,22 @@ The following example shows an implementation of an HTTP request:
 </Request>
 ```
 
-Using the verb attribute, the request defines which HTTP method (or verb) is to be used, e.g. "GET", "POST", "PUT", etc.
+Using the *verb* attribute, the request defines which HTTP method (or verb) is to be used, e.g. "GET", "POST", "PUT", etc.
 
-The url attribute defines the path, optionally including a query string and/or fragment string.
+The *url* attribute defines the path, optionally including a query string and/or fragment string.
 
 > [!NOTE]
 >
-> - Instead of using the url attribute, you can specify the ID of a parameter that holds the request path. This is done using the pid attribute:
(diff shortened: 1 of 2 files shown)


develop/schemadoc/Protocol/Protocol.HTTP.Session.Connection.Request.Parameters.Parameter.md
```diff
@@ -27,6 +27,9 @@ Via HTTP, this has to be done by means of key/value pairs.
 
 Contains a string that specifies the value of the parameter.
 
+> [!NOTE]
+> When you provide the value to be sent through the PID of a parameter, that parameter needs to be of RawType `other` and Type `string`.
```
PedroDebevere, line 31: Only mention Type "string" requirement.
This note could also be added to the second note in this section: https://docs.dataminer.services/develop/devguide/Connector/ConnectionsHttpImplementing.html#request
(I checked the code and this does not seem to apply for Data)

develop/devguide/Connector/ConnectionsHttpImplementing.md
```diff
@@ -36,21 +36,22 @@ The following example shows an implementation of an HTTP request:
...
 >
 >   ```xml
 >   <Request verb="GET" url="/barco-webservice/rest/NetworkWall/alarm">
 >   <Request verb="GET" pid="5045"> <!-- content parameter 5045: "barco-webservice/rest/NetworkWall/alarm" -->
 >   ```
 >
->   - In case the pid attribute is used, parameter 5045 does not need a leading slash ("/").
->   - In case both the url and pid attributes are specified, the pid attribute will be ignored. Typically, only one of these attributes is specified.
+>   - In case the *pid* attribute is used, parameter 5045 does not need a leading slash ("/").
+>   - The parameter referred to in the *pid* attribute must be of type `string`.
```
MariekeGO, line 53: @PedroDebevere, does this match what you had in mind?

MariekeGO (approved):

PR #4994, opened 2025-07-10
DVE - Alarm template inherited from the parent element

MariekeGO:


LuisFreitas-Skyline:
Hi Marieke,

Catarina is OoO.
I think the explanation is too brief. It doesn't clearly state that this behavior only affects exported parameters.

Here's a suggestion:

In scenarios where only the (DVE) parent element has an alarm template assigned, and no alarm template is directly assigned to the DVE child, the child element inherits the alarm template from its parent.
Therefore, when an alarm is triggered on a parameter exported to a DVE, that alarm affects only the state of the DVE child element, not the parent element. This behavior is intentional and designed to avoid duplicate alarms appearing in the Alarm Console for both the parent and child.
If an alarm is triggered on a parameter that is not exported, the alarm will affect the parent element only, not the child element.

LuisFreitas-Skyline (approved):
Now, I think it's clear.

PR #5024, opened 2025-07-16
Add documentation DoD

RobbieDemaegdt:


develop/TOOLS/DefinitionOfDone/About_DoD.md
```diff
@@ -0,0 +1,115 @@
...
+
+- **Driver Development** - New connector development tasks
+- **New Driver Feature** - Adding new features to existing connector
+- **Driver Issue** - Bug fixes and issue resolution in existing connector
+
+## How to Use the Tool
+
+### 1. **Getting Started**
+
+1. Navigate to the DoD Tool home page
```
MariekeGO, line 83: Where is this home page? This is pretty essential info. And is this accessible for Skyline employees only, or also for external users?

develop/toc.yml
MariekeGO, line 1: You've now added this under the third-party tools, but based on the screenshot I get the impression that this is a custom Skyline tool. If this is indeed the case, can you move this one level up in the toc.yml?

develop/TOOLS/DefinitionOfDone/About_DoD.md
MariekeGO: > Only skyline users will have access to this, the home page is just dod.skyline.be. Should this be mentioned in the documentation?

Yes, for both things. The documentation has to include where you can find the tool, and if external users can't access this, we need to mention this.

RobbieDemaegdt: Added both things in the latest commit

EdithLansens:
Hi @RobbieDemaegdt, I've reviewed the DoD tool documentation and restructured the page to make it more intuitive. I mainly focused on removing duplicate information that was already covered elsewhere. This is why, for example, the introduction is now shorter. Let me know if you're OK with the changes, and I'll go ahead and merge the PR.

EdithLansens (approved):

PR #5037, opened 2025-07-22
Updated style guide

skylinerens:
- Changed the guides to rules
- Added rules


dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
-  ![Transparent theme](~/dataminer/images/LCA_Style_Guide_Theme_Transparent.png)
+- Use the same theme consistently across all pages and panels.
 
-- When you define a background color for pages/panels (as part of the theme), never use full white (#FFFFFF) or full black (#000000). Here are some good default colors:
+- [Customize component themes](xref:Customize_Component_Layout) only to make minor adjustments, such as removing padding inside a maps component.
 
-   |☀️  Light background |🌙 Dark background|
-   |---------------------|----------------|
-   |#F4F4F5              |#1E2530         |
+- Always use the *Default* component style, except for the following components, where you should use the *Transparent* style:
```
MariekeGO, line 32: Could you illustrate why we recommend this? If people understand the reason for this rule, they're more likely to follow it.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+- Adjust component sizes to the content they display (e.g. avoid oversized ring and state components or undersized tables and timelines filled with data).
 
-## Panels
+- Before you publish a low-code app, test your layout on different screen resolutions to ensure that it displays correctly on all target devices.
 
-- Panels in a low-code app should primarily be used to show additional information about an item selected on a page. This can for example be a DOM form or additional details/metadata that were not shown on the original page.
+## Navigation
 
-- Panels can be shown on the left or the right of the screen or as a pop-up window. Try to always show a panel in the same location in an app, with the same width, etc. As a default, we recommend showing DOM forms in a pop-up window and showing additional information about a selected table row in a side panel on the right.
+- Use the sidebar as the main navigation to the different pages defined in the app. When there are too many pages in your sidebar, or if you want to show the same data in another way, use the header bar as tabs to open hidden pages.
```
MariekeGO, line 105:
```suggestion
- Use the sidebar as the main navigation to the different pages defined in the app. When there are too many pages in your sidebar, or if you want to show the same data in another way, use buttons in the header bar as tabs to open hidden pages.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-  ![Form pop-up window](~/dataminer/images/LCA_Style_Guide_Form_Popup.gif)
+  > [!NOTE]
+  > When up to two components fill an entire page with no margins, you can place action buttons on the left side of the header bar. Just do not use them as tabs.
```
MariekeGO, line 108: This note isn't clear to me. Could you clarify? Possibly this is a good spot for a screenshot?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-- Every panel should have a close button in the top-right corner. This can be in the menu bar or it can be a separate button.
+- If you are using the header bar as tabs, make sure you **show a title on each page** or **use a circle icon that is filled** on the selected page.
```
MariekeGO, line 110:
```suggestion
- If you are using buttons in the header bar as tabs, make sure you **show a title on each page** or **use a circle icon that is filled** on the selected page.
```

MariekeGO, line 110: Could you add a screenshot to illustrate what you mean with the circle icon?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+- Show **additional information in a side panel on the right** for each item on a page.
 
-- Do not use a background color for titles.
+- If your panel updates based on a selection, **add a title** with the selected item's name.
 
-- Do not make the titles too big: the focus should be on the content of the page.
+- Show **forms in a pop-up** panel.
 
-- We recommend adding a title to a page as a web component instead of just text. For example:
+- Always choose the **same width** in your actions if you are using the same panel more than once.
```
MariekeGO, line 124: Could you clarify or illustrate this rule?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
-    <div style="height:100%;display:flex;flex-direction:column;justify-content:center">
-      <H1 style="color: rgb(20,33,42); font-family:'Segoe UI Light','Segoe UI Web Light','Segoe UI Web Regular','Segoe UI',Arial,sans-serif;margin:0;font-weight:100;font-size:24px">
-        <b>People</b>
-      </H1>
-      <p style="color: rgb(20,33,42,0.7); font-family:'Segoe UI Light','Segoe UI Web Light','Segoe UI Web Regular','Segoe UI','Segoe UI Symbol',HelveticaNeue-Light,'Helvetica Neue',Arial,sans-serif;margin:0;font-weight:600;font-size:16px">
-      An overview of all managed people
-      </p>
-    </div>
-    ```
+- If you have too much information in one panel, you can split it into multiple panels and **use the header bar as tabs**.
```
MariekeGO, line 128:
```suggestion
- If you have too much information in one panel, you can split it into multiple panels and **use buttons in the header bar as tabs**.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-  - Result:
+- **Use the “As overlay” option** when opening a panel.
 
-    ![Good title](~/dataminer/images/LCA_Style_Guide_Title_Good.png)
+- Do not use panels as pages.
```
MariekeGO, line 132: Could you clarify this? How could panels be used as pages? Maybe we should rephrase this to a positive rule: what do you use panels for vs. what do you use pages for?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+- Only include **one access point** for a certain action per page/panel.
 
-- Use short labels.
+- **Confirm** important data change actions with a notification.
 
-- Add an icon to a button whenever possible. This will make it easier for users to see at a glance what the button is for.
+## Components
 
-- Make sure buttons have a transparent background.
+- Make sure you **choose the right visualization** for your data.
```
MariekeGO, line 142: This is very vague, so we should probably clarify this with examples.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-   | 👍 Good | 👎 Bad |
-   |------|-----|
-   |![Good button](~/dataminer/images/LCA_Style_Guide_Button_Good.png)|![Bad button](~/dataminer/images/LCA_Style_Guide_Button_Bad.png)|
+- Make sure you add **[a custom empty component message](xref:Tutorial_Dashboards_Displaying_a_custom_empty_component_message)** that guides your user.
```
MariekeGO, line 144:
```suggestion
- If the data displayed by a component depend on a query, make sure you add **[a custom empty component message](xref:Tutorial_Dashboards_Displaying_a_custom_empty_component_message)** that helps the user in case the query returns no results.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-### Images/icons
+### Title (Web component)
```
MariekeGO, line 146: What do you mean with "web component"? Instead of adding this as part of this header, it might be better to explain this below.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-- Do not add customer logos to apps. While these may look useful for marketing purposes, they usually take up space without adding much value.
+- Titles are optional on pages, depending on whether users will need **more context information** about what they are looking at.
```
MariekeGO, line 148: Could you add one or more examples here?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+- Label your buttons with **enough context** so users understand exactly what each action will do.
 
-  For example:
+- **Add an icon** whenever possible.
 
-  ![Good table](~/dataminer/images/LCA_Style_Guide_Table_Good.png)
+### Images
 
-- When you use HTML inside a table template, there **must** be an **outer HTML element** with a title attribute (with the same value as the table cell); otherwise, if you do not specify a title attribute, the HTML will show up as a tooltip:
+- Do not add customer logos.
```
MariekeGO, line 179: Could you clarify this, keeping in mind that external users may also be using this guide?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
 
-  | 👍 Good | 👎 Bad |
-  |--|--|
-  | `<span title="{Name}" style="font-weight:500">{Name}</span >` | `<span style="font-weight:500">{Name}</span >` |
-  | ![Good table template](~/dataminer/images/LCA_Style_Guide_Table_Template_Good.png) | ![Bad table template](~/dataminer/images/LCA_Style_Guide_Table_Template_Bad.png) |
+- Do not add unnecessary images.
 
-### Grids
+- **Only use dynamic images** with a web component on a page or HTML tool in templates.
```
MariekeGO, line 183: Could you add an example to clarify this?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+- **Add a component title** to your table mentioning which kind of data it is.
 
-    ![Grid summary](~/dataminer/images/LCA_Style_Guide_Grid_Summary.png)
+- Do not show IDs.
 
-- As with tables, the *Empty result message* **must** be customized so that it clearly states why no data is shown in the table, because the default message is not user-friendly. See [Displaying a custom empty component message](xref:Tutorial_Dashboards_Displaying_a_custom_empty_component_message).
+- **Rename your column titles** to straightforward, clear but concise names.
 
-- Like for a table, when you use HTML inside a grid template, there **must** be an **outer HTML element** with a title attribute (with the same value as the grid cell); otherwise, if you do not specify a title attribute, the HTML will show up as a tooltip.
+- If you need a selected state, make sure it is **clear on the full row**.
```
MariekeGO, line 207: Could you add an example to clarify this?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
```diff
@@ -4,288 +4,210 @@ uid: LCA_Style_guide
...
+> Creating a template from scratch? You can find inspiration on [Dribbble](https://dribbble.com/search/shots/popular/product-design?q=cards), sketch and structure parts of your template in [Figma](https://figma.com), and test your HTML/CSS layers in [Codepen](http://codepen.io/).
 
-In order to have consistent navigation, we recommend applying the following order for navigation in a low-code app:
+## Layout
 
-1. Sidebar: Main navigation to the different pages defined in the app.
+- Make sure components on a page/panel are aligned vertically and horizontally.
 
-   ![Navigation via sidebar](~/dataminer/images/LCA_Style_Guide_Navigation_Pages.png)
+- Always use a 50-column grid for your page layout.
```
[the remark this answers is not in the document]
skylinerens: I mean the "number of columns" page setting. By default it isn't 50, but we will update that later.
image

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
skylinerens: We saw that many dashboards and apps display logos of customers or DataMiner. Internally, it was decided to stop doing this. To make this applicable to both internal and external users, we could maybe remove the word "customer" from that sentence?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
skylinerens: If your data is event-based, show it in a timeline. If your data is location-based, show it on a map.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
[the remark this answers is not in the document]
MariekeGO: In that case, I'd change this entry as follows and add a screenshot similar to the one you posted below it (but with the correct value):
```suggestion
- In the page settings, always set the *Number of columns* setting to 50.
```
What is the reason why we recommend this? Adding that info will make users more inclined to stick to this rule.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: So do you mean that a panel's width should never be set to 100%, because it should be clear that it is a panel instead of a page?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: I think this is something we'll need to clarify in this iteration. Why the action buttons? This isn't clear right now.

skylinerens: When header bar actions are used as tabs, we don’t want users to mix them with other actions. If they’re not used as tabs, and the page is fully filled with two components, the header bar can be used for actions.

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
skylinerens: By dynamic images I mean images that change based on your data. For example, each movie has its own image. We don’t want static images that always stay the same and add nothing beyond basic context.
image

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: Could we replace this like this then?
```suggestion
  > Make sure you do not use both regular action buttons in the header and buttons that are meant to be used as tabs, as this will lead to confusion. Regular action buttons should only be added when no tabs are needed.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: So could I change this like this?

```suggestion
- Avoid adding static images that have no added value, but instead add **dynamic images** with a web component on a page or HTML tool in templates, so that the images change based on the displayed data.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: So how should users configure this instead?

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
MariekeGO: Next attempt:
```suggestion
- Only add images if they are **dynamic**, so that they change based on the displayed data. You can do with a web component on a page or HTML tool in templates.
```

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Style_guide.md
skylinerens: Template**S**, because you configure a template for each column. Suggestion: If you override the default row selection indication, make sure that the overrides you configure in the templates is visible on the **entire row**.

PR #5040, opened 2025-07-22
IGQIUpdater - Extend the Add and Update method

SebastiaanSL:


develop/devguide/GQI Extensions/API_Reference/IGQIUpdater.md
```diff
@@ -19,7 +19,7 @@ It allows adding, removing, and updating rows, or updating individual cells.
 
 ### void AddRow(GQIRow)
 
-Adds a new row to the query result.
+Adds a new row to the query result. If a row with the same row key already exists, all cells will be updated instead. This method effectively acts as an *add or update* mechanism.
```
skyline-rge, line 22: Note that it not only updates cells but also metadata, and possibly other row attributes in the future.

```suggestion
Adds a new row to the query result. If a row with the same row key already exists, that row will be updated instead. This method effectively acts as an *add-or-update* operation.
```

develop/devguide/GQI Extensions/API_Reference/IGQIUpdater.md
```diff
@@ -38,7 +38,7 @@ Removes an existing row from the query result.
 
 ### void UpdateRow(GQIRow)
 
-Updates the cells of an existing row in the query result.
+Updates all cells of an existing row in the query result. If the specified row does not exist yet, the update will be queued and applied once the row is added.
```
skyline-rge, line 41: Same here. The previous docs were not quite up-to-date.

```suggestion
Updates an existing row in the query result. If a row with a matching row key does not exist yet, the update event will be queued and applied once the row is added.
```

PR #5045, opened 2025-07-24
Example: Optimizing sort operations for ad hoc data sources

skyline-rge:
- Moved the existing example from the API reference to the Examples section
- Extended the example with a more realistic use case, more detailed explanations and links to other relevant pages


skyline-rge:
@SebastiaanSL  @KlaasSL can you also have a quick look?

MariekeGO:
@skyline-rge, I've reviewed your pull request and made some changes. Could you check if everything is still OK for you? I also see that your branch has some merge conflicts with the main branch. Could you check if you could resolve these?

PR #5046, opened 2025-07-24
Update cloudgateway_change_log.md

JordyGit:


LouisDJ-SL (approved):

release-notes/DxMs/cloudgateway_change_log.md
```diff
@@ -4,6 +4,10 @@ uid: cloudgateway_change_log
 
 # Cloud Gateway change log
 
+#### 24 July 2025 - Fix - CloudGateway 2.17.11 - Incorrect connectivity warnings due to incomplete validation
+
+An issue was introduced in CloudGateway 2.17.5 where the connection to dataminer.services was only validated by IP address, without falling back to DNS as expected. This fallback is needed on systems where the firewall is configured to allow traffic based on DNS rather than IP. This issue has now been resolved.
```
EdithLansens, line 9: The title focuses on connectivity warnings, but the body of the text doesn't mention any warnings being shown. What do you think of this revised version, @JordyGit?
```suggestion
An issue was introduced in CloudGateway 2.17.5 where the connection to dataminer.services was only validated by IP address, without falling back to DNS as expected. On systems where the firewall is configured to allow traffic based on DNS rather than IP, this could result in incorrect connectivity warnings. This issue has now been resolved.
```

PR #5050, opened 2025-07-28
GQI - Improve docs on changing the minimum log level

skyline-rge:
- GQI DxM: Improved explanation on the configuration files
- GQI Logging:
  - Move docs for DxM and SLHelper to separate tabs
  - For GQI DxM: link to configuration files
  - For GQI DxM: remove log settings for the GQI extensions
- GQI extension logging:
  - Add a new section "Changing the minimum log level"
    - Add explanation for DxM and SLHelper in separate tabs
    - Link back to global GQI settings for changing the minimum log level
    - Link to the MinimumLogLevel property on the IGQILogger
    - Explain the behavior of how the settings and property interact

### DIFF
--- a/dataminer/Functions/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.md
+++ b/dataminer/Functions/Dashboards_and_Low_Code_Apps/GQI/GQI_DxM.md
@@ -69,9 +69,11 @@ The GQI DxM comprises multiple processes that work together to handle GQI reques
 
 ## Configuration
 
-Below you can find an overview of the different settings you can configure in the file `C:\Program Files\Skyline Communications\DataMiner GQI\appsettings.custom.json`.
+The GQI DxM can be configured via json configuration files. The default configuration can be found in `C:\Program Files\Skyline Communications\DataMiner GQI\appsettings.json`. Do **not** use this file to change settings because it will be overwritten whenever a new version of the GQI DxM is installed.
 
-Do not edit the file *appsettings.json*, because that file gets overwritten when a new version of the GQI DxM is installed.
+To change configuration settings, create a new configuration file named `appsettings.custom.json` in `C:\Program Files\Skyline Communications\DataMiner GQI\appsettings.json` if it does not exist yet.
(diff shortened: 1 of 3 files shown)


KlaasSL (approved):
LGTM

develop/devguide/GQI_Extensions/GQI_Extensions_Logging.md
```diff
@@ -104,6 +104,47 @@ Example:
...
 
+## Changing the minimum log level
+
+The minimum log level determines which log entries are included in the log files based on their [log level](xref:GQI_GQILogLevel). This can be used to increase or decrease the number of log entries without having to edit the code, for example while debugging.
+
+### [GQI DxM](#tab/gqi-dxm)
+
+The minimum log level is determined by a combination of:
+
+- The global [GQI extension settings](xref:GQI_Logging#changing-the-minimum-log-level).
```
MariekeGO, line 115: As the global GQI settings section you link to here immediately links back to this location for anything related to extensions, I'm wondering if it might not be better to either not include this link here (and perhaps move up the info from line 123-135 below so users see it immediately), or else to put the global GQI log settings for extensions on that general page alongside the other global settings. The effect we get now is that users are sent in a loop: they click the link to find out more about the global extension settings, and then immediately get referred back to the place they came from. (Same for line 141.)

MariekeGO: I mainly want to avoid getting a confusing situation where a user wants to know more about the global GQI extension settings, clicks the link here, and then finds themself immediately referred back to the place they came from, because no information related to the extensions logging is included on the place we linked to.

skyline-rge: I see your point. I would keep the links but maybe move them a bit and add extra clarifications on both pages:

- explain what exactly what they can expect before clicking the link at the source
- explain all cases for why you might have been linked there at the destination

Let me know if you want me to try that or if you want to implement it.

PR #5059, opened 2025-07-31
VerifyGRPCConnection prerequisite/dmupgrade when gRPC becomes default in 10.5.10

bertbsky:


MariekeGO:
@bertbsky, which RN adds this feature? Is it 43190, or possibly another RN as well? What is the main release version where it is introduced?

MariekeGO (approved):

PR #5087, opened 2025-08-07
Documentation related to PDF processing using AI demo

jkockxskyline:
Related to this catalog item: https://catalog.dataminer.services/details/cb4988f9-f3e5-45b7-aade-144f21e9755a


MariekeGO:
@EdithLansens, could you add an icon for this solution?

PR #5091, opened 2025-08-08
RN 43222

EdithLansens:
@Robin-Debel, could you please review this pull request to confirm the accurate inclusion of release note 43222 in the documentation?

Note that I have updated the terminology for consistency: the "*Link to data* icon" is now referred to as the "link icon", and the "*Link to (Data)* pop-up window" is now the "*Link to* pop-up window". While it could technically still be called the *Link to Data* pop-up window in some cases (when only data can be linked), I opted for these new terms because the icon and pop-up were previously also called *Link to feed*, and I believe this updated wording is more future-proof and avoids unnecessary references to older versions. In addition, this new terminology will be clearly understood by users regardless of whether their UI displays "Link to feed", "Link to data", or "Link to."

Thanks in advance!


dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Low-Code_Apps_configuration/LowCodeApps_event_config.md
```diff
@@ -51,6 +51,23 @@ To configure an event to launch a script:
...
+To use this output as data:
+
+1. If available, click the link icon in the action configuration.
+
+1. Navigate to the *Action* tab.
+
+1. Set *Type* to `Script outputs`.
+
+1. Select the desired property name from the *Property* dropdown list.
+
```
Robin-Debel, line 65: the user needs to type the property name, might be best to mention this to avoid confusion.

Robin-Debel (changes requested):

dataminer/Functions/Dashboards_and_Low_Code_Apps/Low_Code_Apps/Low-Code_Apps_configuration/LowCodeApps_event_config.md
EdithLansens: Thanks for the feedback! Adjusted in [commit dd1883e](https://github.com/SkylineCommunications/dataminer-docs/pull/5091/commits/dd1883e57d3724255ed598e09216f72e90e3acc7).

MariekeGO (approved):

PR #5105, opened 2025-08-12
Add BPa 'Check Time Server'

HansSolos:
Copy pasted documentation file from the BPA project and added uid to it.


MariekeGO:
@HansSolos, I've reviewed your PR and made some changes. Could you check if everything still looks OK for you?

PR #5108, opened 2025-08-12
RN 43320

EdithLansens:
@tobedeprez, could you please review this pull request to confirm the accurate inclusion of release note 43320 in the documentation? Thanks in advance!


dataminer/Operator_guide/Alarms/Working_with_alarms/Advanced_analytics_features/Relational_Anomaly_Detection/Relational_anomaly_detection.md
tobedeprez, line 1: I think this is a bit confusing now that there isn't an xml file anymore. If you're using the RAD Manager, or the RAD API on agent 1, you can use it to configure groups on agent 2: if all parameters in the group are hosted on agent 2, the message will be forwarded automatically to agent 2. Maybe it is better to write something like 'Currently, RAD is not able to monitor parameters hosted on multiple agents in a single group. Parameters inside a single group must all be hosted on the same agent. Moreover, prior to DataMiner 10.5.9/10.6.0.0, the *RelationalAnomalyDetection.xml* file on a given DMA must only mention parameters hosted on that agent.'

tobedeprez (changes requested):

PR #5119, opened 2025-08-13
Update SAML_using_Entra_ID.md

FlavioJA:
Added DaaS system cases.
Added indication of automation user addition/removal.


dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
```diff
@@ -211,6 +217,7 @@ Once authentication has been configured, you need to make sure users are provisi
 
 > [!IMPORTANT]
 > If you import over 1000 users or groups, it may take some time before the import is complete. Avoid importing over 10,000 users and groups, as this can result in a timeout.
+> Additionally, users will be automatically added to or removed from the groups in line with the changes made in Azure.
```
MariekeGO, line 220: What is the purpose of this sentence? Is it meant as an addition to the previous one, to tell users that they don't need to import everything because users will get automatically added anyway. Or is it actually entirely separate from the previous sentence, i.e. another thing that is good to know but that isn't linked to the import limitation in any way?

PR #5134, opened 2025-08-19
Changes to account for DevCenter's disappearance

MariekeGO:
As DevCenter is no longer supported and no longer available for download, it had to be removed from our procedures.


MariekeGO:
@SeppeDejonckheere, could you check if the changes to the security section are OK in this PR?

PR #5148, opened 2025-08-25
Switching to Failover with hostname

MariekeGO:
@LuisFreitas-Skyline, I've created this pull request based on your feedback and based on https://community.dataminer.services/question/switching-from-failover-with-vip-to-failover-with-shared-hostname/. Would this be sufficient, or are further details needed?


MariekeGO:
> yes, this is sufficient.

Would you like me to wait until Bert has been able to review this as well, or is it OK to merge already?

LuisFreitas-Skyline:
I think it's safe to proceed with the merge.

PR #5149, opened 2025-08-26
RN 43553

MariekeGO:


release-notes/MediaOps/MediaOps_1.4.0.md
```diff
@@ -17,6 +17,10 @@ While previously new jobs were always created in the draft state, now it is also
 
 ### Enhancements
 
+#### Improvements to capability and capacity filters and node configuration [ID 43553]
+
+In the *Edit job* panel, when nodes are added or resources are picked or swapped for a job, the capability and capacity filters used will now be saved in the node configuration the moment the button to add a node or pick/swap a resource is clicked. This will make sure that the filters used to find a resource are available for when a resource is swapped for a node. As the changes of the filter are only applied when the swap button is clicked, this also means that it is now possible to swap resources for nodes while changing the capabilities at the same time.
```
SkylineCookieMonster, line 22: Maybe it can be simplified:
'When resources are added, picked or swapped, the capability and capacity filters used will take effect when executing the add, pick or swap action. In the past both configuration and node manipulations had to happen separately and could lead to challenges when swapping resources while changing capabilities. Now it is possible to swap resources for other resources while changing the capabilities at the same time.'

SkylineCookieMonster (changes requested):

PR #5179, opened 2025-08-29
Update terminology from 'tool' to 'platform' in DoD doc

SimonV-Skyline:
Feedback from Abel: people tend to be allergic to the word "tool". Therefore, I tried to removing it where not necessary and to replace it by platform where simply removing it didn't make sense.

It allowed me to spot and fix an inconsistency between "The DoD Tool header bar" title vs "Checklist table" and "Smart Actions" ones. Indeed, for consistency, either they should all start with "The DoD Tool ...." or none of them should. I went ahead with second option for simplicity

### DIFF
--- a/develop/TOOLS/DefinitionOfDone/About_DoD.md
+++ b/develop/TOOLS/DefinitionOfDone/About_DoD.md
@@ -2,7 +2,7 @@
 uid: About_DoD
 ---
 
-# Definition of Done (DoD)
+# Definition of Done (DoD) platform
 
 The Definition of Done (DoD) platform is a web application that helps development teams ensure all necessary steps are completed before a task is marked as done. It automates and streamlines task workflows, improves quality assurance, and simplifies documentation and collaboration.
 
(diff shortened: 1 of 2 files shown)


develop/TOOLS/DefinitionOfDone/About_DoD.md
```diff
@@ -2,11 +2,11 @@
 uid: About_DoD
 ---
 
-# Definition of Done (DoD) tool
+# Definition of Done (DoD)
```
MariekeGO, line 5: I think it might be better to add "platform" here as well (and for the sake of consistency possibly also in the toc.yml). This for people who come across this page from the search, so they immediately have a better idea of what the page is about despite the lack of context, as otherwise they might initially think it's about the concept "definition of done".

SimonV-Skyline: done.
On the page itself I added "platform" starting lower case as it made more sense to me and tool also used to be lower case (feel free to change as you see fit)
However, on the toc, I added "Platform" starting with upper case for consistency with "Skyline Driver Passport Platform" just next to it. I would feel like using lower case but then if felt we would also have to adapt "Skyline Driver Passport platform" for consistency and wasn't sure if I should (again, feel free to adapt as you see fit)

PR #5194, opened 2025-09-02
Ad-Hoc Data Source Tutorial using the Class Library

ArneMaes0:
Added a new tutorial using the Class Library to fetch data from the DMS, instead of raw slnet calls.
It's a reworked version of the other one: [Building a GQI data source that retrieves data from a DMS](https://docs.dataminer.services/develop/devguide/GQI_Extensions/Ad_hoc_data/Ad_hoc_Tutorials_GQIDMS.html). But instead fetches some information on the agents in a cluster.

@skyline-rge i've added in the prerequisite that GQI DxM needs to be installed and a note saying version 10.5.4 as I found in the GQI docs that this is the minimum official supported version with GQI DxM.


develop/devguide/GQI_Extensions/Ad_hoc_data/Ad_hoc_Tutorials_Class_Library.md
```diff
@@ -0,0 +1,143 @@
...
+---
+
+# Building a GQI data source that retrieves data from a DMS using the Class Library
+
+In this tutorial, you will learn how you can create a GQI data source that retrieves the version info of the agents your DMS.
+
+Expected duration: 15 minutes.
+
+> [!NOTE]
+> This tutorial uses DataMiner version 10.5.4.
```
skyline-rge, line 12: Instead of a note, this should probably be included in the prerequisites alongside the GQI DxM requirement.

I wonder if this is actually the lowest version that could be used.
Previously we updated [another example](https://github.com/SkylineCommunications/dataminer-docs/pull/4475), where we were able to use GQI in SLHelper with minimum version 10.4.6.
The only requirement was that GQI could provide the raw SLNet connection which is available from [RN39489](https://collaboration.dataminer.services/releasenotes/39489).
Is there another requirement that I'm missing?

ArneMaes0: We could do the approach Simon suggests, but I made another example since in the original tutorial it was displaying client connections but that call doesn't have a nice wrapper in the class library so I choose to display the agent versions as an example. If we do the tab approach than the example would need to align. But I like the tab way.

skyline-rge: If we decide that the GQI DxM is required, then I would stick with the 10.5.4 as you did now. That's probably the easiest. I was just thinking about consistency with that other code example. Since technically it is still possible to use the class library if you manually configure the dll dependencies correctly.

ArneMaes0:
I reworked the example to have the tabs, but I was a bit hesitant to remove the old page, because in the [kata video](https://community.dataminer.services/courses/kata-35/) they showcase that exact document and say that you can view it on the docs.

SimonV-Skyline:
> I reworked the example to have the tabs, but I was a bit hesitant to remove the old page, because in the [kata video](https://community.dataminer.services/courses/kata-35/) they showcase that exact document and say that you can view it on the docs.

I discussed this with Leander and we believe it's best to simply remove the old page so people are not tempted to use the old (slnet way). We'll then look into updating the kata description and provide in there the link to the new page stating the example is a bit different and improved using newer tech by now.

SimonV-Skyline (approved):

develop/devguide/GQI_Extensions/Ad_hoc_data/Ad_hoc_Tutorials_Interact_With_DMS.md
```diff
@@ -0,0 +1,221 @@
...
+---
+
+# Building a GQI data source that retrieves data from a DMS
+
+In this tutorial, you will learn how you can create a GQI data source that retrieves the version info of the Agents in your DMS.
+
+Expected duration: 15 minutes.
+
+> [!TIP]
+> See also: [Kata #35: Interact with your DMS using an ad hoc data source](https://community.dataminer.services/courses/kata-35/) on DataMiner Dojo ![Video](~/dataminer/images/video_Duo.png)
```
MariekeGO, line 12: If the Kata video is now outdated, perhaps we shouldn't include this link anymore, since it won't really be helpful for the user now. What do you think?

SimonV-Skyline: I would tend to agree with removing it indeed. We shoudln't remove the kata but we could promote it less by removing such references to it. I believe the tuto is self-explanatory enough to not need to forward readers to the Kata.

develop/devguide/GQI_Extensions/Ad_hoc_data/Ad_hoc_Tutorials_Class_Library.md
SimonV-Skyline: I would not be too worried there. The old way (SLNET calls) was known to be a not idea solution. While the new way (class library) is meant to stay on the long run. I don't see this changing any time soon.

MariekeGO (approved):

PR #5248, opened 2025-09-16
DataMiner Asset Management application removed

MariekeGO:


jarno-lernou:
After discussing with Leander, it was concluded to hide this documentation in the public docs. Reasons: This is a private catalog item, designed as a sample / use case. Therefore we prefer that we validate this catalog item, and adjust the catalog description accordingly with the information provided here. It is by no means considered a standard solution, as this will be done through the efforts in [SLC-SE-DataMiner Solutions-Asset and Inventory Management | Collaboration](https://collaboration.dataminer.services/project/18872/list)

PR #5252, opened 2025-09-16
Update of RAD API

VeerleLedoux:
Include extended Relational Anomaly Detection functionality available to users through the RAD API starting from DataMiner version 10.5.9.

### DIFF
--- a/src/DataMiner/SLAnalyticsTypes/Rad/RadSubgroupInfo.cs
+++ b/src/DataMiner/SLAnalyticsTypes/Rad/RadSubgroupInfo.cs
@@ -55,7 +55,7 @@ namespace Skyline.DataMiner.Analytics.Rad
         /// <summary>
         /// The parameters in the subgroup.
         /// In case multiple RADSubGroupInfo objects are defined for the same shared model group, the parameters in the subgroups need to be consistent with each other.
-        /// This means that either the parameters in the subgroups have the same label, or have consistent order of the parameters in them.
+        /// This means that either the parameters in the subgroups have the same label, or the order of the parameters in the subgroups is consistent.
         /// </summary>
         public List<RADParameter> Parameters { get; set; }
 


src/DataMiner/SLAnalyticsTypes/Rad/GetRADDataMessage.cs
```diff
@@ -23,6 +23,16 @@ public class GetRADDataMessage : GetSessionedAnalyticsMessage
         /// Gets or sets the end time of the historical time range over which anomaly scores need to be retrieved.
         /// </summary>
         public DateTime EndTime { get; set; }
+        /// <summary>
+        /// Gets or sets the name of the subgroup for which one wants to retrieve anomaly scores in case of a RAD shared model group. Available since DataMiner 10.5.9..
```
tobedeprez, line 27: Remove double .

tobedeprez: Also mention that this does not need to be specified if SubgroupID is specified and that one only has to specify this if the group is a shared model group (same SubgroupID)

src/DataMiner/SLAnalyticsTypes/Rad/RemoveRADSubgroupMessage.cs
```diff
@@ -0,0 +1,51 @@
...
+        public string SharedModelGroupName { get; set; }
+
+        /// <summary>
+        /// The name of the subgroup to be removed.
+        /// </summary>
+        public string SubgroupName { get; set; }
+
+        /// <summary>
+        /// The unique identifier of the subgroup to be removed.
+        /// This field does not need to be specified when SubgroupName has been specified.
```
tobedeprez, line 27: Same as before: mention the converse in SubgroupName as well

tobedeprez (approved):

PR #5294, opened 2025-09-24
Update Overview_Training_dedicated.md

EmmaSaldi:
Please also update the name of this page to DataMiner Foundational training. I'll see where I can add the info about Online Training Consultancy

### DIFF
--- a/dataminer/Professional_services/Training_courses_and_certification/Overview_Training_dedicated.md
+++ b/dataminer/Professional_services/Training_courses_and_certification/Overview_Training_dedicated.md
@@ -1,22 +1,45 @@
 ---
 uid: Overview_Training_dedicated
 ---
+# DataMiner Foundational Training
 
-# Dedicated DataMiner training
+The DataMiner Foundational Training provides a solid introduction to working with DataMiner. It combines theory, demonstrations, and hands-on exercises to help you and your team unlock the full potential of your DataMiner System.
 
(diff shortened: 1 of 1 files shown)


MariekeGO:
Because there is an aka link to the online training consultancy section that you remove here, I can't merge this pull request without breaking that link. As such, I'd prefer if we immediately added this info again already. Perhaps it could be included under [Free self-paced DataMiner training](https://docs.dataminer.services/dataminer/Professional_services/Training_courses_and_certification/Overview_Training_selfpaced.html)?
Also, I'm thinking of moving the information about the video library to that same section, because right now it doesn't seem like a very good fit in the overview of our various training services, especially since we don't really have a dedicated library on Dojo anymore. What do you think?

EmmaSaldi:
Agree. I think free self-paced training is a good place for the OTC.

You can indeed move the information about the video library because I also added this to my list because it didn't make sense to me to add this in the overview of the various training services.
That's part of the free self-paced training as well.

Thank you

________________________________

MariekeGO:
> Agree. I think free self-paced training is a good place for the OTC. You can indeed move the information about the video library because I also added this to my list because it didn't make sense to me to add this in the overview of the various training services. That's part of the free self-paced training as well. Thank you

Done! I've also adjusted the information on the Learning section, as the books page and the video library are no longer available there, and I've removed the video library image, as this was pretty outdated by now. Could you take another look and let me know if everything looks OK for you?

EmmaSaldi (commented):
Changes look good

MariekeGO (approved):

PR #5306, opened 2025-09-25
RN 43793

MariekeGO:
@RobbeDGH, could you check if this pull request is OK to add RN 43793 to the documentation?

### DIFF
--- a/dataminer/Administrator_guide/Swarming/Swarming.md
+++ b/dataminer/Administrator_guide/Swarming/Swarming.md
@@ -34,7 +34,10 @@ Swarming can be enabled from DataMiner 10.5.1/10.6.0 onwards.<!-- RN 41490 --> N
 
 ## Upcoming features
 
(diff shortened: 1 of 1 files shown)


RobbeDGH:
Maybe somthing to clarify that 10.5.1 is the bare minimum for basic elements but that other features might require higher version. To see upcoming feature list?

RobbeDGH:
> > Looks good.
> > Wondering if we should add something to EnableSwarming.html page that mentions the higher version requirement for DVE/VF?
> 
> I wondered about that too, but I'm worried about adding the same information in too many places, since that's not very update-friendly. So as that page only seems to focus on what you do to activate swarming instead of what you have to do to swarm specific items (e.g. bookings), I'd be inclined to leave it as it is.

Agreed, even though it should probably never need updating.

We could also make a table somewhere with a list of what functionality requires which version and then refer to the table for specfics while keeping the minimum requirement 10.5.1/10.6 everywhere

RobbeDGH (approved):
LGTM

PR #5318, opened 2025-09-26
Web DcM is not included in 10.4

wimme:
The Web DcM is not included in 10.4.0 CU20 and is not supported on any 10.4 version.

(Internal release notes on the web projects include 10.4.0 CU20 because all web changes are automatically included in the 10.4 MR and 10.5 MR, but the Web DcM installer itself does not get included in 10.4 versions).


dataminer/Administrator_guide/DataMiner_Agents/Upgrading_a_DMA/Upgrading_Downgrading_Webapps.md
```diff
@@ -15,6 +15,7 @@ Note that it is not possible to upgrade the web apps to a DataMiner 10.5.x versi
 > - Executing a DataMiner upgrade will override any DataMiner web upgrades that have been performed in the past.
 > - New web functionality that depends on new features in the core processes will not be available until you have also upgraded the core software. For example, GQI is part of the core DataMiner software, so new GQI features will only become available with a full DataMiner upgrade.
```
wimme, line 16: This example might be outdated, since new GQI features only get added in the GQI DxM, and the GQI DxM is included in web-only upgrades.

MariekeGO: Perhaps we can just remove the example then?
```suggestion
> - New web functionality that depends on new features in the core processes will not be available until you have also upgraded the core software.
```

MariekeGO (approved):

PR #5327, opened 2025-09-29
Improved user information page

MariekeGO:


dataminer/Administrator_guide/Security/About_DMS_Security/Viewing_user_information.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# Viewing user information
+
+In DataMiner Cube, you can view various detailed information about users, provided that you have the necessary user permissions to do so (i.e. [Administrator](xref:DataMiner_user_permissions#modules--system-configuration--security--administrator) or [Limited administrator](xref:DataMiner_user_permissions#modules--system-configuration--security--specific--limited-administrator), with or without [View users from other groups](xref:DataMiner_user_permissions#modules--system-configuration--security--view-users-from-other-groups), depending on whether access to user information from other groups should be available).
+
+There are two ways you can view this information:
+
+- Go to System Center > *Users / Groups* > *users* and select the user.
+- Look up the user in the search box at the top of the Cube UI, and double-click the search result to open the user card.
```
SeppeDejonckheere, line 12: Is it the goal for this page to also replace the 'Opening_a_user_card' page or do they both exists with more or less the same information?

dataminer/Administrator_guide/Security/Basic_security_configuration/Managing_users/Viewing_user_information.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# Viewing user information
+
+In DataMiner Cube, you can view various detailed information about users, provided that you have the necessary user permissions to do so (i.e. [Administrator](xref:DataMiner_user_permissions#modules--system-configuration--security--administrator) or [Limited administrator](xref:DataMiner_user_permissions#modules--system-configuration--security--specific--limited-administrator), with or without [View users from other groups](xref:DataMiner_user_permissions#modules--system-configuration--security--view-users-from-other-groups), depending on whether access to user information from other groups should be available).
+
+There are two ways you can view this information:
+
+- Go to System Center > *Users / Groups* > *users* and select the user.
+- Look up the user in the search box at the top of the Cube UI, and double-click the search result to open the user card.
```
MariekeGO: Good question. It doesn't seem very user-friendly to make the user click through to that page here if we can just summarize the info in one sentence instead, but on the other hand it does mean we have duplicate info if we keep the user card page. I've now updated the pull request so all the info about user cards is included on this same page instead. While I was at it, I've also moved and improved the info on System Center and on user groups to also make this more user-friendly. Could you take another look?

SeppeDejonckheere (approved):

PR #5336, opened 2025-10-01
DOM editor - configure network share to store DOM Attachments

jaklke:


dataminer/Functions/DOM/DOM_DevOps/DOM_Editor/DOM_editor_attachments_networkshare.md
```diff
@@ -0,0 +1,42 @@
...
+1. In the *Network Share Path* box, fill in the UNC path to the network share. It needs to start with \\\\ and cannot contain any characters that are illegal for a path (e.g. "<") or strings that allow directory traversal (e.g. "../").
+
+    ![DOM Editor: network attachement settings window](~/dataminer/images/DOM_Editor_networkshare.png)
+
+1. Click *Back* again and then click *Apply* and *OK*.
+
+## Notes
+
+- The configuration creates a [DomInstanceNetworkAttachmentSettings](xref:DOM_DomInstanceNetworkAttachmentSettings) object.
+- Changes are applied immediately and do not require a restart.
```
Skyline-ThomasGH, line 42: You do need to reinitialize the DOM manager. Seems like this is not cleary documented in the 'DomInstanceNetworkAttachmentSettings' docs page as well. Only on the main 'DOM module settings' page. Probably best to add something here to remind them.

dataminer/Functions/DOM/DOM_DevOps/DOM_Editor/DOM_editor_attachments_networkshare.md
```diff
@@ -0,0 +1,42 @@
+---
+uid: DOM_editor_attachments_networkshare
```
Skyline-ThomasGH, line 2: Maybe a reference could be added from the DomInstanceNetworkAttachmentSettings page. This already mentions the DOM editor tool, so we could point to this page instead.

dataminer/Functions/DOM/DOM_DevOps/DOM_Editor/DOM_editor_attachments_networkshare.md
jaklke: Added a step to reinitialize the module

dataminer/Functions/DOM/DOM_DevOps/DOM_Editor/DOM_editor_attachments_networkshare.md
```diff
@@ -0,0 +1,48 @@
...
+
+   ![DOM Editor: edit module window](~/dataminer/images/DOM_Editor_edit_module.png)
+
+1. Click *Manager Settings*.
+
+1. Click *Network Share Attachment Settings* to open a new page to configure the settings.
+
+    ![DOM Editor: manager settings window](~/dataminer/images/DOM_Editor_managersettings.png)
+
+1. Select the correct set of credentials from the *Credential* dropdown.
```
MariekeGO, line 33: When "credentials" is used to refer to a set of identifiers for authentication, it should always be used in the plural form (as it refers to e.g. a username *and* a password). I've adjusted this wherever necessary in this text, except here, as this is a direct reference to the UI. Is this something that could be corrected in the UI as well? Then we could either correct this straight away here (since the difference with the current UI is minimal), or merge it as it is and correct it when the UI has been updated, as you prefer.

MariekeGO (approved):

PR #5341, opened 2025-10-02
RN 43847

MariekeGO:
@SkylineCookieMonster, is it OK to keep adding the 1.4.1 RNs to MediaOps 1.5.0, or will a 1.4.1 version be created after all?

### DIFF
--- a/release-notes/MediaOps/MediaOps_1.5.0.md
+++ b/release-notes/MediaOps/MediaOps_1.5.0.md
@@ -19,6 +19,10 @@ uid: MediaOps_1.5.0
 
 On the *Resources* page of the Resource Studio app, when a large number of resources are deprecated or restored, a loading indicator is now displayed, providing feedback on the action in progress. In addition, when you start an action to deprecate or restore multiple resources, you will now first be asked to confirm the action.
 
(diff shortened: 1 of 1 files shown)


SkylineCookieMonster:
@MariekeGO hard to tell, will depend if customers report issues that require fixing before we can finalize the next release with features, then yes. Otherwise, the next release will be 1.5.0 or 2.0.0 depending if we introduce breaking changes or not ...

MariekeGO:
> @MariekeGO hard to tell, will depend if customers report issues that require fixing before we can finalize the next release with features, then yes. Otherwise, the next release will be 1.5.0 or 2.0.0 depending if we introduce breaking changes or not ...

Are you OK with the current solution of keeping everything on a 1.5.0 page, or would you prefer that I create a 1.4.1 page and possibly move the RNs to 1.5.0 later?

MariekeGO:
> @MariekeGO think it makes sense to only have a preview on 1.5.0. If needed we can creat the page for 1.4.1 when needed

Ok, in that case, OK to merge this one?

SkylineCookieMonster (approved):

PR #5369, opened 2025-10-07
Put applications on top

SkylineCookieMonster:
Noticed that for navigating it is easier if these are on top (otherwise you have to scroll down every time)


solutions/standard_solutions/MediaOps/MediaOps.md
MariekeGO, line 1: I'm afraid that if we put this at the top, the section that now starts at line 31 loses much of its effect. Perhaps we could go for an in-between solution and keep that one section at the top, and put this applications section after that, before "Who can benefit" etc.?

MariekeGO (approved):

PR #5401, opened 2025-10-15
Made automation methods that use parameter Description as obsolete (RN 43948)

JorgeFilipeDias:
Made automation methods that use parameter Description as obsolete


MariekeGO:
I'm not sure if this change really makes sense for methods like FindParameterID and FindWriteParameterID, where a user is trying to find the ID based on the description. I don't think using an overload that uses the ID is at all an option there, because the point of the method is to find the ID. What do you think?

PR #5405, opened 2025-10-15
Expand explanation for OctetStringASCII, clarify character encoding dependencies, and provide examples

SkylineThomasCR:


develop/schemadoc/Protocol/Protocol.Params.Param.SNMP.Type.md
```diff
@@ -48,7 +48,15 @@ One of the following predefined values:
...
+  >
+  > For example, if the device sends the hex value `0x4D61E7E3`, this represents `Maçã` under the Western European (Windows-1252) code page, where `0xE7` and `0xE3` map to `ç` and `ã`. However, if the Windows system locale is set to a Cyrillic (Windows-1251) code page, those same byte values would display as `Maзг`.
+  >
+  > By contrast, if `Maçã` were actually encoded in UTF-8 `0x4D61C3A7C3A3` by the device, reading it as `OctetStringASCII` (using the system ANSI code page, e.g. Windows-1252) would produce `MaÃ§Ãã`. In this scenario, use `OctetStringUTF8` instead.
+  >
+  > SNMP itself treats these values purely as octets and provides no context about character encoding. It is the connector's responsibility to interpret those bytes correctly.
+  >
+  > See also: [ANSI code page](https://en.wikipedia.org/wiki/Windows_code_page)
```
MariekeGO: Links to Wikipedia are generally avoided in documentation, as this isn't considered a very reliable source. Is there any other page we could link to? If not, it might be better to just remove this link.

SkylineThomasCR: This page gives an overview of possible code pages:
[code page identifiers](https://learn.microsoft.com/en-us/windows/win32/intl/code-page-identifiers)

This is the start of a bunch of pages that cover some of the same topics as the wikipedia page:
[unicode and character sets](https://learn.microsoft.com/en-us/windows/win32/intl/unicode-and-character-sets)

Do you think it would be useful to link these instead?

PR #5422, opened 2025-10-20
Add small note to .NET installation requirements

RobbeDGH:
Based on the confusion from following dojo question:

https://community.dataminer.services/question/net-9-version/

### DIFF
--- a/dataminer/Reference/DataMiner_Compute_Requirements.md
+++ b/dataminer/Reference/DataMiner_Compute_Requirements.md
@@ -84,8 +84,7 @@ In the table below, you can find which .NET (Framework) versions are required fo
 >
 > - If you install DataMiner using the [v5.0 Installer](xref:Installing_DM_using_the_DM_installer), the required .NET (Framework) versions will be **included in the installation**, so there is no need to install these separately.
 > - We recommend always upgrading to the latest .NET Framework version.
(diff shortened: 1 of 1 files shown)


dataminer/Reference/DataMiner_Compute_Requirements.md
```diff
@@ -85,6 +85,7 @@ In the table below, you can find which .NET (Framework) versions are required fo
 > - If you install DataMiner using the [v5.0 Installer](xref:Installing_DM_using_the_DM_installer), the required .NET (Framework) versions will be **included in the installation**, so there is no need to install these separately.
 > - We recommend always upgrading to the latest .NET Framework version.
 > - Major .NET versions are not cross-compatible, which means that you always need to install the appropriate version mentioned above. If you are for example using a DataMiner version that requires .NET 8.0, this will not work if only .NET 9.0 is installed, so in that case you will need to install .NET 8.0 as well.
+> - It is possible to run multiple .NET versions on a single machine without a problem. You can have .NET 7, .NET 8 & .NET 9 installed all at the same time. The above is just the minimum requirements per DataMiner version.
```
MariekeGO: OK if I combine these two lines like this?
```suggestion
> - Multiple .NET versions can run on the same machine without any problem. You can for example have .NET 7, .NET 8, and .NET 9 installed at the same time. In fact, major .NET versions are not cross-compatible, which means that if .NET 8.0 is required, it is not sufficient to only have .NET 9.0 installed, but you will need to install .NET 8.0 as well.
```

MariekeGO (approved):

PR #5427, opened 2025-10-21
Add WebUIVersion property with XML comments (RN 43840+43875)

AaronVanVyve:
Feature introduced in DataMiner 10.5.12 (RN 43840)


MariekeGO:
> Feature introduced in DataMiner 10.5.12 (RN 43840)

Could you double-check and confirm that this is the correct release note for this change? The internal record only mentions the ScriptName property but says nothing about "WebUIVersion".

PR #5459, opened 2025-10-28
RN 43922

MariekeGO:
@finestracheride, could you confirm if this is sufficient to add RN 43922 to the documentation? To my knowledge, all other instances where Copilot is mentioned have already been adjusted.


finestracheride (approved):
Every other mention of the DxM I am aware of has already been fixed

PR #5461, opened 2025-10-28
Security overview page

SeppeDejonckheere:
Migration (and update) of the dojo security expert hub.


MariekeGO:
> @SkylineCommunications/team-docs-reviewers I was asked to migrate and update the security expert hub on dojo. However, I'm not entirely certain if adding a dedicated page for this is the right approach or if I should have added this content to existing pages. What are your thoughts on this?

This looks fine to me. A dedicated page for this is indeed best, and the spot you picked is perfect. I'd just add another hyperlink to the new page here: https://docs.dataminer.services/dataminer/Administrator_guide/Security/About_security.html.

PR #5470, opened 2025-10-28
Added code pages (RN 43929)

LaurensMT:
Gives a deeper explanation about the different code pages, like Unicode


MariekeGO (commented):
@LaurensMT, I've reviewed your pull request and made some changes. Could you check if everything still looks OK? Note also that we'll still need to specify the main release version for 43929 as well, so users on both release tracks know when this is available.

LaurensMT (commented):
Changes look good. I thought that "e.g." was followed by a comma in American English.
Indeed, the RN still needs to be merged and determined in which DataMiner version it will be eventually added. This documentation commit can be merged once the code of the RN is released.

PR #5471, opened 2025-10-28
Enhance STaaS features documentation with Size Tool info

MicBroSky:
Added information about using the DataMiner Size Tool for database usage details as it is currently not clear to an user what the next steps they should take when running into this STaaS issue.

Let me know if what I suggested is inefficient or would put the user in the wrong area for debugging this issue please.

### DIFF
--- a/dataminer/Administrator_guide/Databases/STaaS/STaaS_features.md
+++ b/dataminer/Administrator_guide/Databases/STaaS/STaaS_features.md
@@ -74,7 +74,7 @@ From DataMiner 10.4.0 [CU14]/10.5.0 [CU2]/10.5.5 onwards<!-- RN 42387 -->, when 
 
 ![ThrottlingAlarmExample](~/dataminer/images/throttling_alarm_example.png)
 
-If you encounter such an alarm, to ensure smooth operation of your DataMiner System, try to identify and resolve the root cause of this higher load. You can run the [DataMiner Size Tool](xref:DataMinerSizeTool) to get details on the database usage for each element on your DataMiner System.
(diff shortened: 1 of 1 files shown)


LouisSkyline (changes requested):
DataMiner Size tool might give an indication on what elements have lots of parameters but having lots of parameters doesn't immediately correlate to having a high STaaS usage. The amount of updates being done on these parameters is the only good indication why you might be throttled. The Cloud ecosystem team has tooling to get insights on the amount of updates and recently something had been added to the CDMR offloads that could help with identifying these issues.

MariekeGO:
> DataMiner Size tool might give an indication on what elements have lots of parameters but having lots of parameters doesn't immediately correlate to having a high STaaS usage. The amount of updates being done on these parameters is the only good indication why you might be throttled. The Cloud ecosystem team has tooling to get insights on the amount of updates and recently something had been added to the CDMR offloads that could help with identifying these issues.

What should we recommend for users who want to troubleshoot this then? Contact support@dataminer.services?

LouisSkyline:
> > DataMiner Size tool might give an indication on what elements have lots of parameters but having lots of parameters doesn't immediately correlate to having a high STaaS usage. The amount of updates being done on these parameters is the only good indication why you might be throttled. The Cloud ecosystem team has tooling to get insights on the amount of updates and recently something had been added to the CDMR offloads that could help with identifying these issues.
> 
> What should we recommend for users who want to troubleshoot this then? Contact [support@dataminer.services](mailto:support@dataminer.services)?

This would indeed be the best option, they will have access to CDMR or can forward these requests to the Cloud team if necessary.

PR #5474, opened 2025-10-29
Update Failover configuration instructions for DataMiner

AVRSkyline:
Setting up a Failover requires the NATS credentials to be reset when running BrokerGateway. Since this is enforced in 10.6, it should become a default step when setting up a failover.

### DIFF
--- a/dataminer/Administrator_guide/Failover/Configuring_Failover/Failover_configuration_in_Cube.md
+++ b/dataminer/Administrator_guide/Failover/Configuring_Failover/Failover_configuration_in_Cube.md
@@ -39,7 +39,7 @@ To enable Failover using DataMiner version **10.1.8 or higher**:
 
       Once you have completed the configuration, the IP addresses will be changed, which may take a while.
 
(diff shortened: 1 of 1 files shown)


dataminer/Administrator_guide/Failover/Configuring_Failover/Failover_configuration_in_Cube.md
```diff
@@ -37,7 +37,9 @@ To enable Failover using DataMiner version **10.1.8 or higher**:
 
    1. Click *Apply* or *OK* to save the configuration (depending on your DataMiner version).
 
-      Once you have completed the configuration, the IP addresses will be changed. Because of this, it could take a while before you can reconnect to the system.
+      Once you have completed the configuration, the IP addresses will be changed, which may take a while.
+
+   1. If you are running DataMiner 10.6.0 or higher, run NATSRepair to update your NATS credentials with the virtual IP. See [BrokerGateway Migration](xref:BrokerGateway_Migration).
```
MariekeGO: Can we also specify the feature release version where this will be needed? And will it also be needed in prior versions if people have already run the migration to BrokerGateway?

PR #5488, opened 2025-11-03
Update migration instructions for DataMiner 10.6.0

AVRSkyline:
Emphasize the importance of completing migration before upgrading to DataMiner 10.6.0.


MariekeGO:
Included in #5509

PR #5582, opened 2025-11-24
RN 44136

MariekeGO:
@simonmestdagh2, this PR is intended to add the web API info from RN [44136](https://collaboration.dataminer.services/releasenotes/44136). Note that the RN mentions 10.5.0 CU10 as the main release version instead of 10.6.0, which seems a bit odd as #5532 only mentions 10.6.0. I've used the version mentioned in the RN here, so let me know if I should change this to 10.6.0.

### DIFF
--- a/develop/webservices/WS_v1/WS_Methods_v1/GetAnnotationsLinkForElement.md
+++ b/develop/webservices/WS_v1/WS_Methods_v1/GetAnnotationsLinkForElement.md
@@ -4,7 +4,7 @@ uid: GetAnnotationsLinkForElement
 
 # GetAnnotationsLinkForElement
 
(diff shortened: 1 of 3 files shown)


simonmestdagh2 (commented):
Annotations are only removed when installing a full DataMiner upgrade package of 10.6.

MariekeGO:
> Annotations are only removed when installing a full DataMiner upgrade package of 10.6.

Do you mean that just installing 10.6.1 isn't enough? Or do you mean that there should be some indication that it's the server version that matters here rather than the web upgrade version? If it's the latter, as I mentioned "DataMiner 10.6.0/10.6.1" rather than "DataMiner web 10.6.0/10.6.1", would this be clear enough? This is the notation we usually use to refer to the server version.

PR #5631, opened 2025-12-04
Firewall Configuration prerequisite update

MariekeGO:
This PR updates the prerequisites info, because I noticed that the Firewall Configuration check is no longer included when you upgrade. @SeppeDejonckheere, could you check if everything is correct?


SeppeDejonckheere:
Maybe it's best to ask the  cloud team. The FirewallConfigurationBPA was indeed replaced with the SecuritAdvisory one, but I can't find anything about the Firewall configuration prerequisite.

PR #5649, opened 2025-12-09
IDP 1.6.0 CU1

MariekeGO:
@SkylineThomasCR, should I still add a preview label for 1.6.0 CU1, or will this CU be released as is soon?


SkylineThomasCR (approved):
No preview label needed. The change is indeed going to be released soon.

PR #5681, opened 2025-12-16
RN 44347+44355

MariekeGO:


edalslc (approved):
Ok with the description of the the 2 LIVE orchestration use cases ( node-by-node , global ). Other aspects of the RN describes what PLAN offers as functionalities to integrate orchestration event which is a PLAN thing.

PR #5682, opened 2025-12-16
Added an explanation on how the Interactivity auto detection works.

JorgeFilipeDias:
Added an explanation on how the Interactivity auto detection works.
Removed commented code from the IAS Toolkit examples as the Interactivity should now be controlled by the tag.

### DIFF
--- a/develop/devguide/Automation/Howto/Getting_Started_with_the_IAS_Toolkit.md
+++ b/develop/devguide/Automation/Howto/Getting_Started_with_the_IAS_Toolkit.md
@@ -27,7 +27,7 @@ To start creating your script:
 
 1. Fill in the solution name and specify a target folder. A folder will be created with the solution name in the target folder. This will contain your solution. Click *OK*.
 
-1. Once your solution has been created, open the script XML file in the Solution Explorer and adjust the Interactivity tag to Always or Optional, depending on the types of execution that the script can have. More info on the [Interactivity](xref:DMSScript.Interactivity) tag.
+1. Once your solution has been created, open the script XML file in the *Solution Explorer* and set the [Interactivity](xref:DMSScript.Interactivity) tag to *Always* or *Optional*, depending on the types of execution the script can have.
 
 1. Then, open the first Exe block by clicking the icon next to it.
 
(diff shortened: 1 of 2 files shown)


develop/devguide/Automation/Howto/Getting_Started_with_the_IAS_Toolkit.md
```diff
@@ -68,12 +70,6 @@ public class Script
     /// <param name="engine">Link with SLAutomation process.</param>
     public void Run(Engine engine)
     {
-        // DO NOT REMOVE THIS COMMENT OR THE SCRIPT WON'T RUN!
```
slc-peter-vanpoucke: Since this comment is removed, I might be good to review [the note about it](https://github.com/SkylineCommunications/dataminer-docs/blob/1e804c3abd97756caac6ab64c11b4eebc1198d57/develop/devguide/Automation/Howto/Getting_Started_with_the_IAS_Toolkit.md?plain=1#L86C55-L86C76).
Maybe it's worth to explain in that note that the ".FindInteractiveClient(" comment is required for Dataminer versions prior to 10.5.9/10.6.0\.

develop/schemadoc/Automation/InteractivityOptions.md
jaklke: Under Content type, I would already mention 
It's advised to set an explicit value Never, Optional, Always because Auto needs to process the complete script and use simple string detection. See section "How Auto-detection determines interactivity" for more information

jaklke: imo, having this information on the top would be better instead of only mentioning at the end in the note.

develop/schemadoc/Automation/InteractivityOptions.md
```diff
@@ -25,3 +25,20 @@ Setting a value different from "Auto" will replace the [options](xref:DMSScript-
 - Setting "Optional" will enable the option flag "HasFindInteractiveClient".
 
 - Setting "Always" will enable the option flag "RequireInteractive". This is the preferred option when Automation scripts are used in an [IAS component](xref:InteractiveAutomationScript) in low-code apps.
+
+## How Auto-detection determines interactivity
```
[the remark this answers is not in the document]
JorgeFilipeDias: Done

PR #5723, opened 2026-01-06
Update QAction documentation for RN44420

BramDevlaminck-Skyline:
Update QAction documentation with the changes from [DCP261954](https://collaboration.dataminer.services/task/261954), [RN44420](https://collaboration.dataminer.services/releasenotes/44420)

Also added documentation on how to debug reoccurring SLScripting and SLProtocol crashes and how to pinpoint which element is causing the crashes. From the code we will put a link to [this](https://docs.dataminer.services/dataminer/Troubleshooting/Procedures/Troubleshooting_procedures.html) page, which in this PR will also have a SLProtocol section. This should make it easy for customers to get to the right procedure depending on if they have an slscripting or slprotocol crash

Wait with merging this change until the mentioned task/RN are merged.

### DIFF
--- a/develop/devguide/Connector/LogicQActions.md
+++ b/develop/devguide/Connector/LogicQActions.md
@@ -169,15 +169,15 @@ All scenarios below assume that the data does **not** need to persist when the e
 
 - **Between elements running the same protocol**: Use a **static QAction** and share data via [**InterApp calls**](xref:InterAppCalls_Introduction).
 
-> [!WARNING]
-> Do **not** share data between elements running the same protocol version using static fields. This approach does not work when DataMiner uses multiple SLScripting processes, which is the default starting from DataMiner 10.6.3/10.7.0<!--RN 44420-->.
+  > [!IMPORTANT]
+  > Do **not** share data between elements running the same protocol version using static fields. This approach does not work when DataMiner uses multiple SLScripting processes, which is the default starting from DataMiner 10.6.3/10.7.0<!--RN 44420-->.
 
 - **Within a single element across multiple executions of the same QAction**: Use an **instanced QAction** and share data through a **non-static field**.
 
-- **Within a single element across multiple executions of two or more different QActions**: Use a **precompiled QAction** with a **singleton or static class** containing a **static dictionary field**. Store and retrieve data per element using this dictionary. The key of the dictionary should be a combination of the **Root DataMiner ID** and the **Element ID**. This combination will guarantee uniqueness of the dictionary key even when the DMS contains multiple agents. The **Root DataMiner ID** is used instead of the Host DataMiner ID to make sure the key stays unique even after an element is swarmed to a different agent.
+- **Within a single element across multiple executions of two or more different QActions**: Use a **precompiled QAction** with a **singleton or static class** containing a **static dictionary field**. Store and retrieve data per element using this dictionary. The key of the dictionary should be a combination of the **Root DataMiner ID** and the **Element ID**. This combination will guarantee uniqueness of the dictionary key even when the DMS contains multiple Agents. The **Root DataMiner ID** is used instead of the host DataMiner ID to make sure the key stays unique even after an element is swarmed to a different Agent.
 
-> [!IMPORTANT]
-> Implement cleanup logic when the element is stopped to remove all entries related to that element from the dictionary, as the lifetime of the static field is independent of the element lifecycle.
+  > [!IMPORTANT]
+  > Implement cleanup logic when the element is stopped to remove all entries related to that element from the dictionary, as the lifetime of the static field is independent of the element lifecycle.
 
 ## SLProtocol(Ext) instance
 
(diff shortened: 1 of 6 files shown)


develop/devguide/Connector/LogicQActions.md
```diff
@@ -107,39 +107,14 @@ The entry point method must be a public method of a public class.
...
+
+> [!IMPORTANT]
+> Since DataMiner 10.6.3/10.7.0, multiple SLScripting processes are active by default. This causes unexpected behavior when static fields are used to share data as not all elements running the same protocol will be located in the same SLScripting process. Furthermore, restarting the element can cause it to move to another SLScripting processes. **Because of this, static fields should not be used to exchange data between elements running the same protocol.** Protocols developed with the assumption that all elements are running in the same SLScripting process can configure DataMiner to only use 1 SLScripting process as described [here](https://docs.dataminer.services/dataminer/Administrator_guide/DataMiner_Agents/Configuring_a_DMA/Configuration_of_DataMiner_processes.html#setting-the-number-of-simultaneously-running-slscripting-processes). It is strongly recommended to adapt the problematic protocol(s) with the approaches described in [Sharing and persisting data](#sharing-and-persisting-data), as this allows DataMiner to use multiple SLScripting processes and increases overall resilience.
```
SimonV-Skyline:
```suggestion
> Since DataMiner 10.6.3/10.7.0, multiple SLScripting processes are active by default. This causes unexpected behavior when static fields are used to share data as not all elements running the same protocol will be located in the same SLScripting process. Furthermore, restarting the element can cause it to move to another SLScripting process. **Because of this, static fields should not be used to exchange data between elements running the same protocol.** Protocols developed with the assumption that all elements are running in the same SLScripting process can configure DataMiner to only use 1 SLScripting process as described [here](https://docs.dataminer.services/dataminer/Administrator_guide/DataMiner_Agents/Configuring_a_DMA/Configuration_of_DataMiner_processes.html#setting-the-number-of-simultaneously-running-slscripting-processes). It is strongly recommended to adapt the problematic protocol(s) with the approaches described in [Sharing and persisting data](#sharing-and-persisting-data), as this allows DataMiner to use multiple SLScripting processes and increases overall resilience.
```

develop/devguide/Connector/LogicQActions.md
```diff
@@ -184,6 +159,30 @@ For example, in the following QAction two parameters can trigger a QAction. Howe
...
+
+#### Between elements running the same protocol
+Use a **static QAction** and share data via [**InterApp calls**](xref:InterAppCalls_Introduction).
+
+#### Within a single element across multiple executions of the same QAction
+Use an **instanced QAction** and share data through a **non-static field**.
+
+#### Within a single element across multiple executions of two or more different QActions
+Use a **precompiled QAction** with a **singleton or static class** containing a **static dictionary field**.  
+Store and retrieve data from this dictionary using the **DataMiner ID** and **Element ID** as a unique key to associate data with the correct element.
```
SimonV-Skyline:
```suggestion
Store and retrieve data per element using this dictionary. The key of the dictionary should be a combination of the **Root DataMiner ID** and the **Element ID**. We need both to make sure the key is unique even in scope of a DMS containing multiple Agents and we need to use the **Root DataMiner ID** as opposed to using the Host DataMiner ID to make sure keys stay unique even after an element gets swarmed to a different Agent.
```

BramDevlaminck-Skyline: I updated the text with a slight variation of what you mentioned here.
I did add the part about the "root dataminer id vs host dataminer id in the context of swarming", but I don't think that should matter as the entries should have been cleaned up when stopping the element as a first step in the swarming. But it is probably safer to use the root dataminer ID to be sure in case something is not properly cleaned up.

MariekeGO (commented):
Would it be useful to also add some information about this (and possible a link to the new info) here: https://docs.dataminer.services/dataminer/Administrator_guide/DataMiner_Agents/Configuring_a_DMA/Configuration_of_DataMiner_processes.html#setting-the-number-of-simultaneously-running-slscripting-processes?

SimonV-Skyline (approved):

BramDevlaminck-Skyline:
@MariekeGO, as the troubleshooting steps for a reoccurring SLProtocol and reoccurring SLScripting crash are almost identical, I reused the same file, and just referenced that file twice from the `toc.yml`. I'm not sure if this is something that has been done before, or if you have any objections against this. It would also be OK for me if this is split up, but that would introduce a lot of duplication.

If you have a better alternative that still keeps a good structure without having to duplicate the whole file, that would also be OK

MariekeGO:
> @MariekeGO, as the troubleshooting steps for a reoccurring SLProtocol and reoccurring SLScripting crash are almost identical, I reused the same file, and just referenced that file twice from the `toc.yml`. I'm not sure if this is something that has been done before, or if you have any objections against this. It would also be OK for me if this is split up, but that would introduce a lot of duplication.
> 
> If you have a better alternative that still keeps a good structure without having to duplicate the whole file, that would also be OK

The same file should never be referenced twice in the toc.yml, as this will cause odd behavior that users may experience as a bug. What we usually do in a case like this is add the information in one place, and then add a cross-reference to that place where appropriate, as duplicate information is indeed to be avoided.

LaurensMT (approved):

PR #5729, opened 2026-01-08
Update DataAPI change log for version 1.4.2

AiltonLuz:
Added multiple entries for DataAPI 1.4.2 fixes and enhancements, including connector deletion notifications, Serilog configuration changes, cluster sync improvements, and installer updates.


MariekeGO:
@AiltonLuz, I've reviewed your PR and made some changes. Could you check if everything still looks OK for you? Note also that for RN 44293, an entry had already been added to the feature release notes, which should not happen as DataAPI is still in soft launch. To prevent this in the future, could you make sure that DataAPI RNs are correctly marked as Documentation Hold = Soft-launch?

PR #5782, opened 2026-01-19
LargeAlarmTrees: update docs

XanderWA:


XanderWA:
Ah yes, my bad, this changed recently due to RN44565.

Any suggestions on what the clearest way is to document this for multiple versions?

PR #5784, opened 2026-01-20
Add docs around use of SAML in clusters and Failover

MichielMN:


dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
```diff
@@ -62,7 +62,7 @@ To set up external authentication, you first need to create an enterprise applic
 
    1. Set *Entity ID* to the IP address or the DNS name of your DataMiner System, for example ``https://dataminer.example.com/``.
 
-      If your DMS consists of DMAs accessible via different URLs or IP addresses, choose one URL or IP address to use as the entity ID.
+      If your DMS consists of DMAs accessible via different URLs or IP addresses, choose one URL or IP address to use as the entity ID. Alternatively, you could use the cluster name of you DMS as entity ID. Azure will use a URL as default but this can be changed to anything without issue. **The only requirement is that it is unique across the identity provider**.
```
MariekeGO: This explanation looks like it might be a bit more complicated than necessary. Would it still be correct if I simplify it to this?
```suggestion
Either choose a specific URL or IP address as the entity ID, or use the cluster name of the DMS. By default, Azure uses a URL, but you can change this to the IP address or cluster name according to preference, as long as the specified entity ID is **unique across the identity provider**.
```

dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_Example_Config_DMS.md
```diff
@@ -0,0 +1,94 @@
...
+      <Surname>http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname</Surname>
+      <Groups claims="true">http://schemas.microsoft.com/ws/2008/06/identity/claims/groups</Groups>
+    </AutomaticUserCreation>
+  </ExternalAuthentication>
+  ...
+</DataMiner>
+```
+
+> [!IMPORTANT]
+> Do not mix Groups claims true and false in the cluster as this will lead to problems with >  user access. This config needs to be the exact same across the cluster for SAML agents.
```
MariekeGO: This isn't entirely clear to me. Would it still be correct if I rephrase it like this? Also, if this is important, it will need to be mentioned in the general procedure as well.
```suggestion
> This configuration must be exactly the same for all Agents in a cluster. Mixing different types of configuration across Agents (e.g. `Groups claims="true"` for one Agent and `Groups claims="false"` for another) will cause problems with user access.
```

dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_Example_Config_DMS.md
MariekeGO: I'm a little bit concerned that this example might not be clear enough yet. Someone might think that what is on this page is all they need to get this setup running, and then they could get in trouble. Perhaps you could adjust it so that the headers on this page mirror those of the general procedure, but everything that is vague and generic there gets filled in with a clear example here? If something is not worth repeating (e.g. the detailed instructions for "Setting up a Microsoft Entra ID Enterprise application), we can include a cross-reference to the general procedure and mention that those are the steps to follow, possibly clarifying what exactly needs to be filled in a specific step. (Try not to refer to the steps by number though, as that's very risky in case of updates).

dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_using_Entra_ID.md
MichielMN: This looks fine to me

dataminer/Administrator_guide/Security/Advanced_security_configuration/Configuring_SAML/SAML_Example_Config_DMS.md
MichielMN: this is what I was going for indeed, that's the reason I used the full XML provisioning tag too to make this point clear. and since this isn't automatically synced but a manual action on each agents, mistakes could happen

PR #5791, opened 2026-01-20
RN 44233 - DOM - Instance-level security

spetebroodt:


spetebroodt:
Hi @Ghysbrecht , can you please review this RN? And can you also explain what is meant with that last note regarding old and new instances?

PR #5879, opened 2026-02-06
RAD 4.0.1 RNs (44238)

MariekeGO:
Does the RAD 4.0.1 release only include RN 44238, or should additional RNs be included?


SkylineDennisDr:
Hey Marieke,

I only merged that release note (RN 44238).
I did notice that there are 2 other release notes that were actually already merged in the RAD Manager 4.0.0: namely RN44246 and RN 44244. Both are on "Ready to be tested" but maybe I can just put them to completed? I also notice that Main release version is set to 4.0.0, but actually I guess it's the Solution version that should have this value? 
If you confirm, I don't mind updating it!

PR #5887, opened 2026-02-09
Revise backup file names in RollbackSwarming.md

MicBroSky:
Updated backup file naming conventions for clarity based on a recent Swarming performed on 10.5.12


wdm-slc:
FYI: Looks like the extra "ElementFolder_BeforeSwarmingMigration" got replaced by a hashed version of that string as part of changes to how recycling works in RN40565 (10.5.5 / 10.6.0)

PR #6160, opened 2026-04-10
Update Changing the IP of a DMA documentation

samasdevops:
Added a tip about reconnecting from Cube Client after changing DMA IP

Trying to connect to a hostname previously used on the old IP will fail and make you think you didn't do the server Ip address change correctly.

image


MariekeGO:
@samasdevops, after consulting with our DevOps Engineers, I received feedback that this is not the expected behavior of Cube and you have most probably encountered a bug. As such, instead of documenting this, it would be better if you [contacted DataMiner Support](https://aka.dataminer.services/tech-support) in order to investigate this issue.

PR #6189, opened 2026-04-16
RN 45318 + 45320

MariekeGO:
@ConanVanHove, as this release note appears to depend on RN [44714](https://docs.dataminer.services/release-notes/Web_apps/Web_apps_Feature_Release_10.6/Web_apps_Feature_Release_10.6.4.html#dashboardslow-code-apps---gqi-enhanced-filtering-when-using-the-gqi-dxm-id-44714), I have also changed the mentioned required DataMiner version. If a higher version will actually be required, let me know.


ConanVanHove (commented):
Hi Marieke, I discussed it with Harun who implemented the fix. We don't know if the RN 44714 you refer to is in any way related to this problem. The issue was found when installing Ticketing on Ziine which ran indeed another version of DataMiner. And that is why we fixed it. But it does not rely on that version and the fix will also still work on a previous version.

MariekeGO:
> Hi Marieke, I discussed it with Harun who implemented the fix. We don't know if the RN 44714 you refer to is in any way related to this problem. The issue was found when installing Ticketing on Ziine which ran indeed another version of DataMiner. And that is why we fixed it. But it does not rely on that version and the fix will also still work on a previous version.

I thought it was that release note because it's mentioned in the [breaking changes](https://docs.dataminer.services/release-notes/Breaking_changes/Breaking_changes.html) with the same issue as described in RN 45318. But if Ticketing doesn't rely on that version, I'll adjust the required version again.

ConanVanHove (approved):
All OK, thank you.

PR #6193, opened 2026-04-17
Update Ticketing 2.0.3 release notes

ConanVanHove:
Added important installation permissions and tips for Ticketing.

### DIFF
--- a/release-notes/Ticketing/Ticketing_1.0.2.md
+++ b/release-notes/Ticketing/Ticketing_1.0.2.md
@@ -4,10 +4,19 @@ uid: Ticketing_1.0.2
 
 # Ticketing 1.0.2
 
-> [!NOTE]
-> This version of the Ticketing app is only intended for use with DataMiner version **10.5.0 up to (and including) 10.5.8**.
->
-> Note that Ticketing needs the **GQI DxM** to be enabled. For information on how to enable this prior to DataMiner 10.5.8, refer to [Enabling or disabling the use of the GQI DxM](xref:GQI_DxM#enabling-or-disabling-the-use-of-the-gqi-dxm).
+## Prerequisites
+
+- This version of the Ticketing app is only intended for use with DataMiner version **10.5.0 up to (and including) 10.5.8**.
+
+- Ticketing needs the **GQI DxM** to be enabled. For information on how to enable this prior to DataMiner 10.5.8, refer to [Enabling or disabling the use of the GQI DxM](xref:GQI_DxM#enabling-or-disabling-the-use-of-the-gqi-dxm).
+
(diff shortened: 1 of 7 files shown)


MariekeGO (commented):
A couple of questions:
- This new large "important" block at the top of the RNs may be a little intimidating for users. Would it be OK for you if I change this prerequisites section to a regular list of prerequisites instead, without putting the content in alert blocks? 
- Should this also be added for the other Ticketing RN documents?

Note also that I changed the references to the user permissions to hyperlinks. We try to do this as much as possible because this way, if a user permission is renamed or removed, we'll get a notification that the cross-reference no longer works, which works as a reminder that the user permission also needs to be updated here.

ConanVanHove:
Yes, these can be changed to regular prerequisites. At the time this seemed more important, but I think I was a little bit misled. This was because the installation on Ziine failed. But this was because of insufficient rights. On most systems where an administrator installs stuff this shouldn’t be a problem. But it is still good to mention.

Yes, this will also be needed for the other Ticketing RN entries. But I wanted to wait for the feedback on this first.

Thank you for adding the hyperlinks.

MariekeGO:
> Yes, these can be changed to regular prerequisites. At the time this seemed more important, but I think I was a little bit misled. This was because the installation on Ziine failed. But this was because of insufficient rights. On most systems where an administrator installs stuff this shouldn’t be a problem. But it is still good to mention.
> 
> Yes, this will also be needed for the other Ticketing RN entries. But I wanted to wait for the feedback on this first.
> 
> Thank you for adding the hyperlinks.

I've adjusted the prerequisites to a list format now. Does this look OK for you? If yes, I'll add the same info about the user permissions to the other Ticketing RN pages using the same formatting.

ConanVanHove (commented):
The user installing Ticketing needs the following **user permissions**: ...
...
In addition, they will need to have **write access** to the root view ...

Related to the usage of "they". This seems to imply this is only for the user installing Ticketing. But it is also required for any user of Ticketing.

So perhaps it would be best to also add this permission to the list and then add an extra remark that this write access is also needed for every user of Ticketing as well.

MariekeGO:
> The user installing Ticketing needs the following **user permissions**: ... ... In addition, they will need to have **write access** to the root view ...
> 
> Related to the usage of "they". This seems to imply this is only for the user installing Ticketing. But it is also required for any user of Ticketing.
> 
> So perhaps it would be best to also add this permission to the list and then add an extra remark that this write access is also needed for every user of Ticketing as well.

I've adjusted this. Let me know what you think!

ConanVanHove (commented):
Perfect! Thank you.
This can now be added to the other versions as well.

MariekeGO (approved):

PR #6196, opened 2026-04-20
RN 45138 updated + 45332/45398 added

MariekeGO:


ConanVanHove (commented):
Your remark about still being able to view the details for Rejected/Cancelled/Closed tickets is not fully relevant. These details can all be viewed already on the information page. In fact we should prevent the editing altogether. But that is technically not possible. So we just disable the Save button. Here you make it sound like a feature/capability, whereas it is more like a shortcoming. So better not to mention it I would say.

ConanVanHove (commented):
I made a remark on this on another Submit review. Can you pick it up from there?

MariekeGO:
> Your remark about still being able to view the details for Rejected/Cancelled/Closed tickets is not fully relevant. These details can all be viewed already on the information page. In fact we should prevent the editing altogether. But that is technically not possible. So we just disable the Save button. Here you make it sound like a feature/capability, whereas it is more like a shortcoming. So better not to mention it I would say.

OK, adjusted!

ConanVanHove (commented):
OK, thank you.

PR #6200, opened 2026-04-21
Conditions on groups linked to triggers after and before groups

ThomasW-Skyline:
Extension on the docs for conditions on groups linked with triggers before and after groups. Simon V to verify content first!

### DIFF
--- a/develop/devguide/Connector/LogicConditions.md
+++ b/develop/devguide/Connector/LogicConditions.md
@@ -132,7 +132,10 @@ Conditions can be defined on the following items:
 > [!NOTE]
 > Avoid using conditions on a Timer. Instead, use a condition on the timer group(s).
 >
-> For a group condition, the evaluation of the condition will be performed when the group is executed by the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
+> For a group condition, the evaluation of the condition will be performed when the group is dequeued by the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
+
+> [!IMPORTANT]
+> Triggers linked to the group (before and after) will not go off when the condition result is 'false'. Behavior is happening in full from 10.4.8 (CU1) and 10.4 (CU5). Before these versions, the trigger before could be executing before the condition is validated.
 
 To build more advanced conditional constructs, you can use brackets.
 
--- a/develop/devguide/Connector/LogicConditions.md
+++ b/develop/devguide/Connector/LogicConditions.md
@@ -135,7 +135,7 @@ Conditions can be defined on the following items:
 > For a group condition, the evaluation of the condition will be performed when the group is dequeued from the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
 
 > [!IMPORTANT]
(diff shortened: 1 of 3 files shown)


develop/devguide/Connector/LogicConditions.md
```diff
@@ -132,7 +132,10 @@ Conditions can be defined on the following items:
 > [!NOTE]
 > Avoid using conditions on a Timer. Instead, use a condition on the timer group(s).
 >
-> For a group condition, the evaluation of the condition will be performed when the group is executed by the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
+> For a group condition, the evaluation of the condition will be performed when the group is dequeued by the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
```
SimonV-Skyline: I think "dequeued by" is not correct as the "group execution queue" is a queue, not the process/thread. the queue is not the do-er. So either we change it to "dequeued from" or "picked up from" either we change it to "dequeued by the SLProtocol thread". But I think one of the first 2 alternative is better because, in this context, we care more about the location than about what process/thread does the work.

```suggestion
> For a group condition, the evaluation of the condition will be performed when the group is dequeued from the group execution queue. This means the group will always be added to the queue of the protocol thread anyway. If the condition value changes between the moment the group is added to the queue and the moment the group is executed, it is possible that the behavior of the group is different than was intended when the group was added. This also means that it is no problem to have a poll group with a condition as the last group in a timer.
```

develop/devguide/Connector/LogicConditions.md
```diff
@@ -132,7 +132,10 @@ Conditions can be defined on the following items:
...
+
+> [!IMPORTANT]
+> Triggers linked to the group (before and after) will not go off when the condition result is 'false'. Behavior is happening in full from 10.4.8 (CU1) and 10.4 (CU5). Before these versions, the trigger before could be executing before the condition is validated.
```
SimonV-Skyline: Trying to make the sentence more digestible.

```suggestion
> When the group has a condition, `before group` and `after group` triggers will only go off if the condition result is 'true'. Note that prior to DataMiner 10.4.8 (CU1) and 10.4 (CU5), `before group` triggers would go off no matter the group condition result, this has now been fixed.
```

develop/schemadoc/Protocol/Protocol.Triggers.Trigger.Time.md
```diff
@@ -48,6 +48,9 @@ Not all Protocol.Triggers.Trigger.Time values can be used in combination with th
 |session|timeout|The trigger will go off when a timeout occurs on the specified session.<!-- RN 12542 -->|
 |timer|before|The trigger will go off before the specified timer is executed.|
 
+> [!IMPORTANT]
+> Conditions on the group cause triggers linked to the group (before and after) not to go off when the condition result is 'false'. Behavior is happening in full from 10.4.8 (CU1) and 10.4 (CU5). Before these version, the trigger before could be executing before the condition is validated.
```
SimonV-Skyline: For now, suggested the same rephrasing as I did on the logicConditions page. However, I wonder if it wouldn't be nicer to put this info directly within the appropriate table entries:

- group | before | The trigger will go off before the specified group is executed. Note that, starting from DM 10.4.8 (CU1) and 10.4 (CU5), the trigger will only go off if the group condition is evaluated to 'true'.
- group | after | The trigger will go off after the specified group is executed. Note that the trigger will only go off if the group condition is evaluated to 'true'.

```suggestion
> When the group has a condition, `before group` and `after group` triggers will only go off if the condition result is 'true'. Note that prior to DataMiner 10.4.8 (CU1) and 10.4 (CU5), `before group` triggers would go off no matter the group condition result, this has now been fixed.
```

SimonV-Skyline (changes requested):
I believe it would also be beneficial to add this info to the [Group/Condition tag page](https://docs.dataminer.services/develop/schemadoc/Protocol/Protocol.Groups.Group.Condition.html).

MariekeGO (approved):

PR #6221, opened 2026-04-27
GoogleMaps instead of gmaps

Joachim-Ally-Skyline:
I think gmaps is outdated, based on a recent test with the OptusPoc?


wimme:
Both "gmaps" and "GoogleMaps" are being used and are supported, to avoid confusion it might be better to document both.

MariekeGO (approved):

PR #6251, opened 2026-05-04
RN 45295

EdithLansens:
@Skyline-ThomasGH, could you please review this pull request to confirm the accurate inclusion of release note 45295 in the documentation? Thanks in advance!


Skyline-ThomasGH:
RN45295 is a change that did a small tweak in the Web APIs so when the Web APIs are interacting with the DOM back-end, and DOM returns a NoPermission error, that the error returned to the actual client by the Web APIs is more clear. The action that was not allowed, was already available on the DOM NoPermission back-end error. I don't think there documentation on this web API call, so I believe docs change is needed for this RN.

EdithLansens:
> RN45295 is a change that did a small tweak in the Web APIs so when the Web APIs are interacting with the DOM back-end, and DOM returns a NoPermission error, that the error returned to the actual client by the Web APIs is more clear. The action that was not allowed, was already available on the DOM NoPermission back-end error. I don't think there documentation on this web API call, so I believe docs change is needed for this RN.

Okay, thanks for the feedback! Closing this PR.

PR #6277, opened 2026-05-11
Update Alarm ID retrieval instructions

MicBroSky:
Clarified instructions for obtaining the Alarm ID, specifying the section from which to copy it.


dataminer/DataMiner_Tools/SLNetClientTest_tool/SLNetClientTest_tool_advanced_procedures/SLNetClientTest_clear_from_event_cache.md
```diff
@@ -10,7 +10,7 @@ To do:
 
 1. In DataMiner Cube, get the alarm ID of the alarm you want to clear.
 
-   You can do so by [making sure the Alarm ID column is displayed in the Alarm Console](xref:ChangingTheAlarmConsoleLayout#adding-or-removing-columns) and then copying the ID from there.
+   You can do so by [making sure the Alarm ID column is displayed in the Alarm Console](xref:ChangingTheAlarmConsoleLayout#adding-or-removing-columns) and then copying the ID from there. The Alarm ID needed for the command in the alarm column will be from the **third section**. Ex: 000000/00/**0000000**.
```
MariekeGO: @saatsky, based on this proposed change from Micah, I'd like to further clarify this procedure by adding a screenshot that illustrates where users can find the alarm ID. However, looking in Cube, I see that the "Alarm ID" column shows two alarm IDs now — I assume one being the root and the other being the latest record in the alarm tree. Is it the root alarm ID that should be used here? If yes, it might be better if we recommend opening the "Root Alarm ID" column instead, to prevent possible confusion.

Image

saatsky: Agree with the remark please proceed

MariekeGO (approved):

PR #6278, opened 2026-05-12
Update Ticketing 2.0.4 release notes

ConanVanHove:
Added important note about version usage restrictions.

### DIFF
--- a/release-notes/Ticketing/Ticketing_2.0.4.md
+++ b/release-notes/Ticketing/Ticketing_2.0.4.md
@@ -5,8 +5,7 @@ uid: Ticketing_2.0.4
 # Ticketing 2.0.4
 
 > [!IMPORTANT]
-> Note that this version can only be used on a STaaS/DaaS DataMiner system (so not on a self hosted and managed system)!
-
+> This version of the Ticketing Solution can only be used on a DataMiner System using [Storage as a Service](xref:STaaS) (with self-managed DataMiner nodes or [DaaS](xref:DaaS_hosting)). Systems with self-managed data storage are not supported.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


release-notes/Ticketing/Ticketing_2.0.4.md
```diff
@@ -4,6 +4,10 @@ uid: Ticketing_2.0.4
 
 # Ticketing 2.0.4
 
+> [!IMPORTANT]
+> Note that this version can only be used on a STaaS/DaaS DataMiner system (so not on a self hosted and managed system)!
```
MariekeGO: Wouldn't it be better to add this as another prerequisite in the list below? It seems odd to add one specific prerequisite in an "important" box, but not the rest. (Same also goes for #6279.)

ConanVanHove: Personally I see a big difference between these two preconditions. A prerequisite in the context of deploying an application typically implies you could potentially run it if you are on the right DataMiner version. But here you simply cannot run it, whatever version you are on.

MariekeGO: I'm not sure I follow. As a STaaS/DaaS system is something that is needed for the Ticketing version to work, this looks like a prerequisite to me: if a prerequisite is not met, the version doesn't work — prerequisites are by definition not optional. Or do you mean that we should highlight this particular prerequisite more because users might not expect it, as previous versions were supported on self-managed systems?

ConanVanHove: It is actually the other way round. We did not test it until very recently and were not aware that it did not work on a self-hosted system. So it is not so much as that STaaS/DaaS is a requirement, but rather that it does not work on a self-hosted system as would normally be expected. Of course it should and it will in the next upcoming release. So for those reasons I would still prefer to see it as an important message.

MariekeGO (approved):

PR #6288, opened 2026-05-15
Added Billing and Subscriptions installation information

GelberMSkyline:


MariekeGO (commented):
Some of the changes in this pull request are unnecessary and will either cause issues or create an inconsistency with our usual syntax:

- We deliberately included the blank lines after the metadata at the top and between list items. There should always be a blank line after the metadata, and the blank lines between list items here prevent inconsistencies in the layout when there are complex lists with indented paragraphs (as is the case here). Could you revert the change removing these blank lines?
- All items in numbered lists should start with ".1", as this improves the maintainability of the list. Could you revert the changes adjusting the numbering?

MariekeGO (approved):

PR #6312, opened 2026-05-21
Add descriptions to custom solutions

spetebroodt:


solutions/custom_solutions/META_WhatsApp_Messaging/META_WhatsApp_Messaging_Installation.md
```diff
@@ -1,5 +1,6 @@
 ---
 uid: META_WhatsApp_Messaging_Installation
+description: Prerequisite and setup steps to install META WhatsApp Messaging, including Meta app creation, WhatsApp setup, and user configuration.
```
MariekeGO: To make this more attractive in the Google search results, I'd prefer to adjust this to an extremely shortened version of the procedure instead of a dry description of the page. As Google describes it, the description has to be a kind of pitch that convinces the user that the page is what they're looking for, so that requires a different kind of language.

```suggestion
description: To install META WhatsApp Messaging, make sure the prerequisites—such as a correctly configured alarm notification app—are met, then install the connector
```

solutions/custom_solutions/PLM_Tool/Installing_PLM_tool.md
```diff
@@ -1,5 +1,6 @@
 ---
 uid: Installing_PLM_tool
+description: Instructions on how to set up the DataMiner Planned Maintenance tool, with prerequisites and initial element configuration in Cube.
```
MariekeGO: Same as for the comment above, if a call to action like "discover X" or "Learn about Y" doesn't make sense, instead of a dry description of the content, we should go for a very brief summary as a kind of "teaser".

```suggestion
description: To install the PLM tool, make sure the prerequisites are met, deploy the package from the DataMiner Catalog, and then configure the element.
```

PR #6335, opened 2026-05-28
KI: SLSNMPAgent RTEs

MariekeGO:


dataminer/Troubleshooting/Known_issues/KI_SLSNMPAgent_RTE_SNMP_manager_not_responding.md
```diff
@@ -0,0 +1,33 @@
...
+
+# SLSNMPAgent RTEs when SNMP manager is not responding
+
+## Affected versions
+
+- 10.5.0 Main Release versions starting from DataMiner 10.5.0 [CU8].
+
+## Cause
+
+When an SNMP manager is not responding and there is a locking issue between the polling thread and the sending thread, the latter does not release the lock, causing SLSNMPAgent runtime errors (RTEs).
```
DieterDegrande: 'When an SNMP manager is not responding and there is a locking issue between the polling thread and the sending thread' is wrong.
It's because of an introduced locking issue due to the mismerge that the behaviour occurs, better worded in my opinion as:
'When an SNMP manager is not responding, a locking issue occurs between the polling thread and the sending thread because the latter does not release the lock, ...'

DieterDegrande (approved):

PR #6357, opened 2026-06-03
Update to overview structure & terminology

MariekeGO:
- Adjusted headers & TOC to mention cluster and nodes terminology.
- Moved system layout page to "About DataMiner" section and renamed it to "DataMiner System architecture and components", keeping the "DMA peripherals" section as a separator page in the Administrator guide.
- Removed obsolete page about eventing vs. polling (as this is irrelevant now that gRPC is used by default).


MariekeGO:
@EdithLansens, could you update the DataMiner Systems icon on the Administrator guide index page so it mentions "DataMiner clusters" instead?

PR #6404, opened 2026-06-11
Reorganize Swarming limitations section

jaklke:
Move and rewrite the "Limitations" section in SwarmingElements.md: relocate it further down the page, convert the paragraph into an explicit bullet list of element types that cannot be swarmed, and preserve the smart-serial server-mode bypass snippet and tip. Clarifies technical restrictions (localhost polling, SNMP trap distribution, element connections, smart-serial server mode) and points to upcoming features for future support.

### DIFF
--- a/dataminer/Administrator_guide/Swarming/SwarmingElements.md
+++ b/dataminer/Administrator_guide/Swarming/SwarmingElements.md
@@ -8,27 +8,6 @@ With DataMiner Swarming, you can swarm elements from one DataMiner Agent to anot
 
 When you are swarming an element so it gets hosted on a different DataMiner Agent, a temporary transition occurs. The element will be stopped and then started again on a new host. While this happens, a message will be displayed to inform users that the element is currently swarming. The ability to open element cards or change the element configuration for the involved element will be temporarily suspended. Once the element migration is complete, it will become accessible again.
 
-## Limitations
-
-At present, Swarming is not yet supported for certain specific types of elements. Refer to [Upcoming features](xref:Swarming#upcoming-features) for information on which types of elements are supported already and which will be supported in the future.
-
-Because of the way Swarming functions, it is not possible to swarm elements polling localhost and elements receiving SNMP traps in a DMS with trap distribution disabled on at least one DMA.
-
-By default, smart-serial elements in server mode also cannot be swarmed. However, from DataMiner 10.6.6/10.7.0 onwards<!--RN 45173-->, you can explicitly allow such elements to be swarmed by adding the following configuration to the *Protocol.xml* file:
-
-```xml
-<Swarming>
-    <BypassChecks>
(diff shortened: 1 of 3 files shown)


jaklke:
@PieterVanCompernolle I did this to make it easier for people to understand what's currently not supported.

PieterVanCompernolle:
I feel this change makes it less clear that some limitations are temporary while others are 'by design'.

jaklke:
Users are exposed to limitations regardless of whether they are permanent or temporary. In practice, some temporary limitations can persist for a long time, and labeling them as upcoming improvements may create expectations that are too distant or uncertain.
What users need is a clear view of what is currently supported, without having to cross-reference limitations with future improvements. While this indeed reduces visibility on planned changes, limitations that are also on the roadmap can be rephrased to explicitly indicate that they apply _for now_ or _currently_. The list of improvements is also kept explicitly btw.

dataminer/Administrator_guide/Swarming/SwarmingElements.md
```diff
@@ -85,3 +64,31 @@ By default, this checkbox is not selected for new and existing elements.
...
+            <Check>smartSerialAsServer</Check>
+        </BypassChecks>
+    </Swarming>
+
+    ```
+
+     With this configuration in place, at startup, the element can send a message to the data source to indicate where data should be sent. As a result, the fact that the smart-serial connection is in server mode will no longer be considered a valid reason to prevent the element from swarming.
+
+    > [!TIP]
+    > See also: [Making changes to a protocol.xml file](xref:Advanced_protocol_functionality#making-changes-to-a-protocolxml-file)
```
MariekeGO: To improve the findability of this info for AI, I'd prefer to move the info on how to enable swarming of smart-serial elements in server mode to a separate section. Would this be OK for you?

jaklke: Sure!

MariekeGO: I've now moved this info to a separate page. As this configuration requires changes to the protocol.xml, it made more sense to include this in the development guide. OK to merge like this?

jaklke: the protocol xml changes can be used in case devices support being reconfigured (so they send their data to the new DMA). Only with this specific condition, smart-serial in server mode is possible. It would still make sense to me that elements with smart-serial connections in server mode have their dedicated bullet item in the list : the majority of connectors doesn't fit the condition.

MariekeGO: Adjusted! Can you take another look?

jaklke: Perfect for me. Thanks!

MariekeGO (approved):

PR #6430, opened 2026-06-19
gRPC vs .NET Remoting topic

MariekeGO:
@bertbsky, as your answer to [this question](https://community.dataminer.services/question/slnet-to-grpc-migration-path-hints/) seemed very interesting for a larger DataMiner audience, I've added a page with this info in this pull request. Does this look OK for you or are there any changes you'd like to propose? Note that I've currently used the image you added as-is, but if everything is OK with this, I'll have its layout adjusted a bit to fit better on docs (I just wanted to wait with this in case changes are needed first).


bertbsky (approved):
Looks good. I can provide the image source in draw.io format if that helps.

PR #6467, opened 2026-06-29
Moved ISO compliance to top of the page

SeppeDejonckheere:
Moved ISO compliance to the top of the page since this is the most meaningful item.


SeppeDejonckheere:
Discussed and approved by Bert

MariekeGO (approved):

PR #6468, opened 2026-06-29
Document remote log collection via LogCollector

XanderWA:


dataminer/DataMiner_Tools/SLLogCollector.md
```diff
@@ -106,6 +111,14 @@ SL_LogCollector.exe -c -h
 >
 > From DataMiner 10.4.0 [CU21]/10.5.0 [CU9]/10.5.12 onwards<!-- RN 43588 -->, an option is available that makes it possible to skip these BPA tests (see table above).
 
+## Remote log collection
+
+From DataMiner 10.6.3/10.7.0 onwards<!-- RN 44384 -->, SLLogCollector can collect logs from other DataMiner Agents in the cluster and optionally upload the resulting packages to the cloud. This requires SupportAssistant 1.8.0.0 or higher.
+
+- Select *Trigger remote LogCollector* to remotely collect logs from one or more other DataMiner Agents. A list with checkboxes is displayed where you can select the Agents on which LogCollector should run. The collected packages are placed at the selected location (the default folder or a custom folder).
+
+- Select *Offload packages to cloud* to upload the resulting log packages to the cloud. This also triggers the remote LogCollector, so you can select the Agents on which LogCollector should run in the same way. After you click *Start*, a GUID will be displayed, which you can provide to Skyline via an existing support email thread so the package can be fetched.
```
MariekeGO, line 127: When I checked these options on my own DMA, I noticed that the only option I could actually select was "Offload packages to Cloud". This also automatically selected "Trigger remote LogCollector". However, I wasn't able to select that last option independently. Is this indeed how Log Collector is expected to work? If yes, we will need to adjust this text a bit, because now users will want to select "Trigger remote LogCollector" first, and they may think it's not working or something is wrong.

XanderWA: I added some more clarifications, there are indeed some edge cases where the UI dynamically is adjusted which I forgot about. Copilot completed the list based on codebase so everything should be covered now.

MariekeGO (approved):

PR #6493, opened 2026-07-02
Add APP item type to GitHub guidelines

saatsky:
Propose adding APP to the Skyline repository naming conventions for application repositories (e.g. UI/Vibe Coded Web Apps), enabling clear names like SLC-APP-DocumentHub-UI instead of overloading EXT (SLC-EXT-DocumentHub-UI)

### DIFF
--- a/develop/CICD/Skyline Communications/Github/Use_Github_Guidelines.md
+++ b/develop/CICD/Skyline Communications/Github/Use_Github_Guidelines.md
@@ -72,6 +72,7 @@ If the repository is private, the name should look like this (using "-" as separ
   | UDAPI | User-Defined APIs |
   | V | Visio files |
   | EXT | External tools - software  |
+  | APP | Application code (web/mobile/desktop/Vibe Coded Apps)  |
 
   > [!NOTE]
   > If you think an item type should be added, please contact us so we can add it before you create the repository.


develop/CICD/Skyline Communications/Github/Use_Github_Guidelines.md
```diff
@@ -72,6 +72,7 @@ If the repository is private, the name should look like this (using "-" as separ
   | UDAPI | User-Defined APIs |
   | V | Visio files |
   | EXT | External tools - software  |
+  | APP | Application code (web/mobile/desktop/Vibe Coded Apps)  |
```
jarno-lernou-skyline, line 75: @saatsky , I think we want to frame it as 'Custom Apps' and typically they will be shipped as a solution towards the customers, so I think we can keep 'SOL' for this. Note: Vibe Coded Apps should not really be mentioned as this has an bad connotation towards customers. Let me know what you think on my comment.

@SebastiaanSL  for awareness

SebastiaanSL: Thanks Jarno, the "Custom apps" is very much a working title. Not sure what the name will be. However, from the core software perspective we actually try to not care if it's a LCA vs a custom app. I wonder if this is something you want to apply it as well and just call it "App", be it a custom app or LCA...

jarno-lernou-skyline: I agree, App or Solution, in essence they are the same so for me, nothing should change. I rather include a GitHub topic if we want to mention this more clearly, but does not have to be part of the name of the repository.

MariekeGO: @jarno-lernou-skyline, do I understand it correctly that it would be better not to implement the proposed change here?

saatsky: Recently @leanderdruwel-skyline requested that on any repository that contains a Custom App, to add the topic _dataminer-custom-app_  . 
I think this "releases" the need to state on the repo name the wording "custom app". 
For this particular PR the intention is to add a item type that can cover the UI repos been LCA or Vibe codded to the guidelines aligned with the already existing ones. 

I can give the example of _**DocumentHub**_, where we have: 

- _SLC-S-DocumentHub_ repo
- _Skyline.DataMiner.Dev.Utils.Solutions.DocumentHub_ repo 

And now need to decide the repo name for the custom app in between:
 _**SLC-APP-DocumentHub**_  or _**SLC-S-DocumentHub-CustomApp**_

I think the first seems the best approach and would like to suggest that on the guidelines.

PR #6505, opened 2026-07-06
Update PATCH API version from 10.6.7 to 10.6.8

jaklke:
According to RN, supported version for PATCH support is 10.6.8. @Skyline-ThomasGH to confirm


Skyline-ThomasGH (approved):
Indeed, code was merged on 22/05 in the afternoon, which is right after the seal of 10.6.7. This is indeed part of 10.6.8.

MariekeGO (approved):

PR #6525, opened 2026-07-09
Dead link detector

spetebroodt:


spetebroodt:
Hi @PedroDebevere , could you have a look at the custom prompt and workflow adaptation in this PR? Both do an external link check. The workflow action is non-blocking.

.github/workflows/azure-static-web-apps-victorious-sea-0e7978f03.yml
```diff
@@ -128,3 +128,38 @@ jobs:
...
+    needs: build_job
+    name: External Link Check
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true
+      - name: Check external links in Markdown
+        id: linkcheck
+        continue-on-error: true
+        uses: lycheeverse/lychee-action@v2
```
PedroDebevere: For improved security, consider targeting the corresponding commit instead of the tag version.

.github/workflows/azure-static-web-apps-victorious-sea-0e7978f03.yml
```diff
@@ -128,3 +128,38 @@ jobs:
         azure-search-indexer: indexer-1745835640388
         # Admin key used to connect to Azure Cognitive Search instance
         azure-search-admin-key: ${{ secrets.AZURE_SEARCH_KEY }}
+
+  external_link_check:
```
PedroDebevere: With every PR, this will now check for all dead links across the entire repo? If this is the case, I'd suggest to have dedicated pipeline to check for dead links e.g. once a day (optionally also with support to trigger it manually).
Otherwise, whenever a PR is created, or something is merged, the dead link detector will run every time.

.github/workflows/azure-static-web-apps-victorious-sea-0e7978f03.yml
spetebroodt: copilot suggested to use the following: uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8

spetebroodt: @PedroDebevere , would this be OK?

.github/workflows/azure-static-web-apps-victorious-sea-0e7978f03.yml
spetebroodt: @PedroDebevere , I now split off the external link check in a separate workflow file. Can you double-check?

.github/workflows/external-link-check.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    name: External Link Check
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true
+
+      - name: Check external links in Markdown
+        id: linkcheck
+        continue-on-error: true
```
PedroDebevere, line 19: suggestion: remove continue-on-error
When removing this, the pipeline will fail. Is this enough to go unnoticed, or should some additional steps be added? 

On, [https://github.com/lycheeverse/lychee-action](https://github.com/lycheeverse/lychee-action) the following is mentioned: 
_When used in conjunction with [Create Issue From File](https://github.com/peter-evans/create-issue-from-file), issues will be opened when the action finds link problems (make sure to specify the issues: write permission in the [workflow](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions) or the [job](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions))._

But if you want to have issues created automatically, we'll need to be sure that false positive results will be rare.

.github/prompts/dead-link-detector.prompt.md
```diff
@@ -0,0 +1,109 @@
+---
```
PedroDebevere, line 1: General remark: I gave this custom prompt a try and it eventually found dead links. However, during the session, I had to do quite some confirmations on many steps and the process took quite some time until the results showed up.

The custom prompt nicely describes what needs to be checked and also what can be skipped or should not be considered a dead link. The external-link-check is probably more limited. It's probably not capable of e.g. knowing that some placeholders should not be considered. Ideally, that same results show up in the workflow and the prompt.

To have uniform results on both the prompt and workflow, I'm wondering if it would be a better design to e.g. create e.g. a custom MCP server/agent/skill that executes the logic of detecting dead links. The MCP server could then for example run some scripts (that were generated by an agent) which implement all the logic currently described in thee custom prompt. The workflow could then invoke these scripts as well.

This will probably also save some tokens/credits as it will now run some pre-created scripts.

PR #6538, opened 2026-07-13
DCP Changelog 14-07-2026

SkylineKevinMOR:


release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+A back button is now available on the create organization step in the DaaS deploy wizard.
+
+### 14 July 2026 - Enhancement - Admin - Documentation button on DMS Overview for expiring DaaS system
+
+On the DMS Overview page in the Admin app, a documentation button is now shown when a DaaS system is about to expire.
+
+### 14 July 2026 - Enhancement - Home - Increased line allowance for Open in desktop notification
+
+The *Open in desktop* notification now has increased line allowance.
```
MariekeGO: This is a bit too vague. What is the "Open in desktop notification"? And do you mean that it can now consist of more lines? Why is this change needed?

release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+### 14 July 2026 - Enhancement - Home - Remote URL Indicator now shows preview of remote URL
+
+The *Remote URL Indicator* now displays a preview of the remote URL below the input field.
```
MariekeGO: What is this Remote URL Indicator and where is it shown?

release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+### 14 July 2026 - New feature - Admin - Created by and Created At info in subscription detail panel
+
+The detail panel of a subscription on the Billing page of the Admin app now includes the *Created by* and *Created At* information.
```
MariekeGO: It's a bit strange that "by" is not capitalized here but "At" is. Possibly it should be "Created at"?

release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+In the Catalog app, inline code in the description of Catalog items was not rendered correctly. This issue has been resolved.
+
+### 14 July 2026 - Enhancement - Admin - API key references renamed to Key in Audit
+
+In the Audit section of the Admin app, the initiator type has been renamed from *ApiKey* to *Key*. Related error messages have also been updated.
+
+### 14 July 2026 - Enhancement - Admin - Usage page removed
+
+The usage page has been removed from the Admin app.
```
MariekeGO: Does this mean that this page is no longer up to date: [https://docs.dataminer.services/dataminer/dataminer_services/Admin/Usage_and_Billing/Viewing_info_on_system_usage.html](https://docs.dataminer.services/dataminer/dataminer_services/Admin/Usage_and_Billing/Viewing_info_on_system_usage.html)? 
If yes, could you adjust it so it will remain correct after this change?

release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+The Settings page in the Admin app has been updated to include an auto-save option for all settings. Settings that involve managing lists of items, such as IP addresses or domains, are excluded from this auto-save. Some settings with significant system impact now also require confirmation before they are applied.
+
+### 14 July 2026 - Enhancement - General - Icons added to all apps in the header app menu
+
+All applications in the app menu of the header now have icons.
+
+### 14 July 2026 - Fix - Catalog - Clearing search value on Catalog home page not handled correctly
+
+An issue has been resolved where clearing the search value on the Catalog home page was not handled correctly.
```
MariekeGO: Would it be possible to add more details on what this incorrect handling meant for the user?

release-notes/Cloud/dataminer_services_change_log_2026.md
```diff
@@ -9,6 +9,66 @@ The dataminer.services platform gets updated continuously. This change log can h
...
+
+### 14 July 2026 - Fix - General - Notification counter appearing on top of popups
+
+An issue has been resolved where the notification counter was displayed on top of popups.
```
MariekeGO: This is a bit vague. What is the notification counter and which kinds of pop-ups are involved?

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: On home, there is an open on desktop button to open the system in cube.
This will launch an notification that its trying to open the cube application with a link to install it if the user doesnt have it.
This message was due to its length cut off and showing ...
With the increased line allownace that is not the case anymore.

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: On Home, when uyou click add a dataminer system. In the second step of the wizard.
image

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: That indeed a typo from me :p

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: Correct, we will no longer show raw usage.
The only usage the user will be able to see is the billable usage thats accessible through the billing page

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: Yes, basicly on the home page of catalog when you enter a search term and then remove it completely.
If you then would click the browse catalog button next to it it would go to the browse page but the last letter of the searchterm would activate instead of the empty value. Causing the browse page to filter on that last character

release-notes/Cloud/dataminer_services_change_log_2026.md
SkylineKevinMOR: Its in the header bar on the notification bell. If a notification triggers a small colored circle with the amount of unseen notifications will appear on the bell.

Our more recent side panels would be display above the header but the counter of the notification had a higher number in the layering order which cause it to be on top of those panels

release-notes/Cloud/dataminer_services_change_log_2026.md
MariekeGO: Does this mean that the page I referred to is actually completely outdated and should be removed?

dataminer/dataminer_services/Admin/Usage_and_Billing/Viewing_info_on_system_usage.md
MariekeGO, line 1: I get the impression that this fully overlaps with the existing page [Viewing billing information](https://docs.dataminer.services/dataminer/dataminer_services/Admin/Usage_and_Billing/Viewing_billing_information.html), so I think it would be better to update that page with any new information that is relevant and to remove the current page. That way we avoid duplicate info, which can lead to inconsistenties and confusion.

SkylineKevinMOR: Thats sounds indeed like a better solution to extend that page and remove the one i updated.

MariekeGO (approved):

PR #6596, opened 2026-07-30
Document script for GPON OLT creation

EstebanSkyline:
Added documentation for the CreateOLTs automation script, detailing how to create GPON OLT elements individually or in bulk from a CSV file, including supported vendors and configuration steps.


solutions/standard_solutions/EPM/EPM_GPON/GPON_CreateOLTs.md
MariekeGO, line 1: As elsewhere in this solution documentation GPON was adjusted to xPON, should this also happen here? Or is GPON OK here?

EstebanSkyline: GPON here it is ok ?

MariekeGO (commented):
@EstebanSkyline, I've reviewed your pull request and made some changes. Could you confirm if everything still looks OK for you?
Note that I'm still not entirely sure if the use of "GPON" is OK here. We had to change all references to the solution from "GPON" to "xPON" a while back, so it seems a bit odd that in this particular instance, "GPON" is still OK. But possibly there's a specific reason that I don't know about?

EstebanSkyline:
Hello Marieke.

Now I get the issue about the xPON and GPON.
I used the term GPON because it was inside the GPON folder, but if we have been using the term xPON on all our documentation so far, I think it would be a great Idea to leave it as xPON. 
Sorry I had no context about the documentation uniformity and the xPON-GPON confusion

Thank you for your explanation and your attention to this major detail.!

MariekeGO (approved):

PR #6607, opened 2026-08-04
Update synchronization details in Overview_of_the_different_subfolder…

PieterVanCompernolle:
…s.md

Added important notes on file synchronization and modification notifications for DMAs.


dataminer/Reference/Skyline_DataMiner_Folder/Overview_of_the_different_subfolders.md
MariekeGO, line 1: As this "important" block is getting rather large, and it also provides a lot of information that users might not necessarily expect to find on a page that provides an overview of folders, I'd prefer to move this to a separate file. I was thinking of adding a "Data synchronization" page under "[About DataMiner clusters](https://docs.dataminer.services/dataminer/Administrator_guide/DataMiner_Systems/About_DataMiner_Systems.html)", with basic info on the midnight sync and the information from this "important" block. We can then add a link to that page here instead. Would this be OK for you?

PieterVanCompernolle: sure np

MariekeGO: I've now added the separate data synchronization page, and I've added a link to it both on the files overview and the subfolders overview page. Could you take a look at the new page and let me know if everything looks OK or if any changes are needed?

PieterVanCompernolle (commented):
approved

MariekeGO (approved):

PR #6614, opened 2026-08-05
RN 45843

EdithLansens:
@RobbeBultynck, I'm not entirely sure how to document release note 45843. According to the release note, reference traces and their makers can now be used and saved in presets, but this was already listed on the [*Using Spectrum Analysis presets* page](https://docs.dataminer.services/dataminer/Functions/Spectrum_Analysis/Working_with_Spectrum_Analyzer_elements/Using_Spectrum_Analysis_presets.html#:~:text=the%20trace%20visibility-,A%20reference%20trace,-Marker%20positions). Reference trace markers aren't specifically mentioned there, so for now I've added those, but maybe this falls under "Marker positions", which is mentioned in the list. Let me know if I should remove this again.

Could you tell me what has concretely changed for the user? What will they notice in the Spectrum Analysis UI that wasn't possible before?

Are there any other pages in the [Spectrum Analysis documentation](https://docs.dataminer.services/dataminer/Functions/Spectrum_Analysis/Working_with_Spectrum_Analyzer_elements/Spectrum_analyzer_cards.html) that need updating because of RN 45843?

### DIFF
--- a/dataminer/Functions/Spectrum_Analysis/Working_with_Spectrum_Analyzer_elements/Using_Spectrum_Analysis_presets.md
+++ b/dataminer/Functions/Spectrum_Analysis/Working_with_Spectrum_Analyzer_elements/Using_Spectrum_Analysis_presets.md
@@ -12,7 +12,7 @@ Spectrum Analysis presets can be used to reload settings, to reload a set of mar
 
 - Display settings, including the trace visibility
 
-- A reference trace (including reference trace markers, from DataMiner 10.5.0 [CU18]/10.6.0 [CU6]/10.6.9 onwards<!--RN 45843-->)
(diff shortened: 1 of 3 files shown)


RobbeBultynck:
Lemme explain what is changed. 
So what we had before was indeed the possibility to save a reftrace already, but this was a snapshot of the current trace at the moment you save the preset.

What we've changed: 
- We've added buttons to save/reset the reftrace at any moment in time. 
image

- We've added the possibility to use markers on the reftrace. 
image

EdithLansens:
Based on the extra info you provided, I gave it another go 😊 What do you think, @RobbeBultynck?

MariekeGO (approved):

PR #6640, opened 2026-08-07
Split up and updated DxM & DcM pages to match platform vision

SLC-brechtseys:


dataminer/Reference/toc.yml
```diff
@@ -119,8 +119,13 @@ items:
         topicUid: Activating_SLProtocol_as_a_64_Bit_Process
       - name: Activating SLProtocol as a 32-bit process
         topicUid: Activating_SLProtocol_as_a_32_Bit_Process
-  - name: DataMiner Extension Modules (DxMs)
-    topicUid: DataMinerExtensionModules
+  - name: DataMiner modules
```
MariekeGO, line 122: I'd prefer to move this info to the Administrator guide section, instead of keeping it under Reference. This mainly because I'm rather concerned about the ambiguity of the term "DataMiner modules", given that we also have "modules" in the Cube sidebar, and those are something completely different. I'm hoping that if we move this to the Administrator guide, it will be more clear that this is related to how DataMiner functions, rather than what you can do with it. OK for you?

MariekeGO:
@EdithLansens, could you add an icon for "DataMiner modules" to the Admin guide index page?

PR #6641, opened 2026-08-07
Fix deadlock issue in CloudGateway 3.3.2

ruben12321:
Resolved a deadlock that could keep a broken connection alive, causing for an unstable cloud connection.

### DIFF
--- a/release-notes/DxMs/cloudgateway_change_log.md
+++ b/release-notes/DxMs/cloudgateway_change_log.md
@@ -4,9 +4,9 @@ uid: cloudgateway_change_log
 
 # Cloud Gateway change log
 
-#### TBD - Fix - CloudGateway 3.3.2 - Resolved possible deadlock
+#### 10 August 2026 - Fix - CloudGateway 3.3.2 - Resolved possible unstable connection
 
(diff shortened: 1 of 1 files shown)


release-notes/DxMs/cloudgateway_change_log.md
```diff
@@ -4,6 +4,10 @@ uid: cloudgateway_change_log
 
 # Cloud Gateway change log
 
+#### TBD - Fix - CloudGateway 3.3.2 - Resolved possible deadlock
+
+Resolved a deadlock that could keep a broken connection alive, causing for an unstable cloud connection.
```
MariekeGO: This should be rephrased to better match the format of other fix descriptions (i.e., focus on the issue from the perspective of the user, for optimal findability of fixes).

JordyGit (approved):

MariekeGO:
@ruben12321, I've adjusted the PR to further make the fix match our usual format for fixes and also clarify when the issue could occur. Could you confirm if everything still looks OK for you?

JordyGit (approved):

MariekeGO (approved):

PR #6647, opened 2026-08-10
MediaOps Plan: RNs 46146/46151/46214/46215/46225

MariekeGO:


release-notes/MediaOps/MediaOps_Plan/MediaOps_Plan_2.0.0.md
```diff
@@ -45,6 +45,14 @@ The selected job type can be used to filter the timeline on the *Job View* page,
 
 If only one category is available, it is automatically selected for the job.
 
+#### Scheduling: Custom job types [ID 46151]
+
+Instead of all jobs being automatically assigned the *Scheduled* type, you can now define custom job types and assign them when creating or editing a job.
+
+Job types are defined using the Categories Solution. From MediaOps 2.0.0 onwards, a new *Job Type* scope is created in Categories with two default types: *Ad Hoc* and *Scheduled*. You can freely add additional types on top of these defaults.
```
SkylineCookieMonster: Only the Scheduled job type is being created. The Ad Hoc one was legacy (from demo versions) which was never used eventually, so we removed it.

release-notes/MediaOps/MediaOps_Plan/MediaOps_Plan_2.0.0.md
```diff
@@ -45,6 +45,14 @@ The selected job type can be used to filter the timeline on the *Job View* page,
...
+
+The icon of the selected job type is displayed in the *Job Type* filter of the timeline view. If no icon has been assigned to a type, a default icon is displayed.
```
SkylineCookieMonster: Maybe we should make it more clear that the icon can be configured from the Categories app? For example, "The icon defined on the Job Type from the Categories application, will be used from the Scheduling application as filter option (name will appear when hovering over the icon). If no icon is defined on the Job Type, then the default filter icon will be used"

SkylineCookieMonster (approved):

PR #6668, opened 2026-08-13
Update connection instructions with internet access note

ruben12321:
Added note about internet access requirements for login.

### DIFF
--- a/dataminer/dataminer_services/Connecting_to_cloud/Connect_to_dataminer_services.md
+++ b/dataminer/dataminer_services/Connecting_to_cloud/Connect_to_dataminer_services.md
@@ -46,6 +46,9 @@ Before you connect your DMS to dataminer.services (unless you [connect with a DM
 1. On the System Center \> *Cloud* page, click the *Connect* button. A pop-up browser window will open.
 
    > [!NOTE]
+   > If the server or PC you are on does not have general access to the internet, you will not be able to login. Open cube from a server or pc that does have general access to the internet, or copy the url from the header bar before clicking the login button and browse to it from a server or pc that does. 
+
+   > [!NOTE]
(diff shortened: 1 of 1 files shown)


dataminer/dataminer_services/Connecting_to_cloud/Connect_to_dataminer_services.md
```diff
@@ -45,6 +45,9 @@ Before you connect your DMS to dataminer.services (unless you [connect with a DM
 
 1. On the System Center \> *Cloud* page, click the *Connect* button. A pop-up browser window will open.
 
+   > [!NOTE]
+   > If the server or PC you are on does not have general access to the internet, you will not be able to login. Open cube from a server or pc that does have general access to the internet, or copy the url from the header bar before clicking the login button and browse to it from a server or pc that does.
```
MariekeGO, line 49: I'd like to rephrase this a bit, but I'm not sure I understand everything correctly. Do you mean that when the pop-up window opens, the user should copy the URL from the header bar, paste this in a browser on a computer with internet access, and then continue with the next step?
```suggestion
   > This requires general access to the internet. If you cannot access Cube on a computer with direct access to the internet, copy the URL from the header bar of the pop-up window before clicking the login button and browse to that URL from a computer with general access to the internet.
```
I also find it a bit odd that we're mentioning a login button in this note, but not elsewhere in the procedure. Is something missing in the procedure? Or is this actually a different button?

ruben12321: yes, that's indeed what i mean. The login button is indeed not mentioned, as it's just part of following the registration flow. I guess it could be excluded with the way your suggestion rephrases it?

MariekeGO (approved):

PR #6689, opened 2026-08-19
Update MediaOps.Plan version requirement in release notes

ConanVanHove:
Updated the required version for MediaOps.Plan to 1.6.x or higher for ticket assignment functionality.

### DIFF
--- a/release-notes/Ticketing/Ticketing_2.0.0.md
+++ b/release-notes/Ticketing/Ticketing_2.0.0.md
@@ -20,7 +20,7 @@ uid: Ticketing_2.0.0
 
 - Installing the following apps alongside the Ticketing Solution will provide access to **additional functionality**:
 
-  - [MediaOps.Plan](https://catalog.dataminer.services/details/1b67a623-4ca6-4d25-8b3d-ed4e39496a75) 1.6.x or higher (Requiring DataMiner 10.6.4/10.7.0 or higher!): Required to be able to assign people to tickets.
+  - [MediaOps.Plan](https://catalog.dataminer.services/details/1b67a623-4ca6-4d25-8b3d-ed4e39496a75) 1.4.x or higher: Required to be able to assign people to tickets.
   - [InfraOps](https://catalog.dataminer.services/details/5a1edac2-45aa-4498-8ab7-ee322d07da27): Required to be able to link assets to tickets.
 
 ## New features
--- a/release-notes/Ticketing/Ticketing_2.0.0.md
+++ b/release-notes/Ticketing/Ticketing_2.0.0.md
@@ -20,7 +20,7 @@ uid: Ticketing_2.0.0
 
 - Installing the following apps alongside the Ticketing Solution will provide access to **additional functionality**:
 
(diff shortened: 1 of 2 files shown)


release-notes/Ticketing/Ticketing_2.0.0.md
```diff
@@ -20,7 +20,7 @@ uid: Ticketing_2.0.0
 
 - Installing the following apps alongside the Ticketing Solution will provide access to **additional functionality**:
 
-  - [MediaOps.Plan](https://catalog.dataminer.services/details/1b67a623-4ca6-4d25-8b3d-ed4e39496a75) 1.4.x or higher: Required to be able to assign people to tickets.
+  - [MediaOps.Plan](https://catalog.dataminer.services/details/1b67a623-4ca6-4d25-8b3d-ed4e39496a75) 1.6.x or higher (Requiring DataMiner 10.6.4/10.7.0 or higher!): Required to be able to assign people to tickets.
```
MariekeGO: Should I implement this same change in Ticketing release notes for 2.0.1 to 2.1.0?

ConanVanHove: No, since the 2.0.x should no longer be used now that 2.1.0 is available.

MariekeGO: I'm afraid I can't quite follow. You're increasing the minimum MediaOps Plan version in the release notes for **2.0.0**, so surely that would mean that this also needs to be adjusted in the subsequent release note documents? Or did you mean to increase the minimum MediaOps Plan version for 2.1.0 instead?

ConanVanHove: Ah, I must have added it on the wrong entry then. Indeed, only for 2.1.0. My bad. Thank you for spotting this and pointing this out.

MariekeGO: I've moved the change to the 2.1.0 release notes and also changed the DataMiner version requirement mentioned at the top, as it wouldn't make much sense to first mention 10.5.9/10.6.0 as the required version and then add later on that it's actually 10.6.4/10.7.0. OK to merge like this?

ConanVanHove:
Hm, I understand, but this is a bit tricky/sensitive. It is only when a customer also optionally installs MediaOps Plan (in order to get access to the People&Organizations app) that the "higher" versions are required. Most customers will indeed want to have P&O so they can assign tickets to a person or team, but it could well be that for a more backend-targeted integration P&O is not required and then there is no need for the higher versions and we would then unnecessarily require those.

MariekeGO:
> Hm, I understand, but this is a bit tricky/sensitive. It is only when a customer also optionally installs MediaOps Plan (in order to get access to the People&Organizations app) that the "higher" versions are required. Most customers will indeed want to have P&O so they can assign tickets to a person or team, but it could well be that for a more backend-targeted integration P&O is not required and then there is no need for the higher versions and we would then unnecessarily require those.

I've adjusted this. OK to merge like this?

ConanVanHove (commented):
OK, perfect.

MariekeGO (approved):

PR #6699, opened 2026-08-21
Support lifecycles update

MariekeGO:


dataminer/Reference/Software_support_life_cycles.md
```diff
@@ -48,11 +48,18 @@ The following features currently have the "End of Engineering" status. They will
...
 | SRM Service Info | This object with booking-related service info is no longer needed since SRM 2.0.2 ([RN 40666](xref:SRM_2.0.2#skipservicehandling-automatically-enabled--srmserviceinfo-objects-no-longer-used-id-40666)). | End of Life to be confirmed. |
 
+<!-- 
+When 10.7.0.x is released, add the following items:
+| Polling of data in DataMiner web apps | WebSockets will be used instead. | End of Support to be confirmed. |
+| SLOffload tool | This tool is obsolete and cannot be used in systems with a Cassandra general database | End of Life as of DataMiner 10.8.x (Q4 2027). |
+| Visual overviews in DataMiner web apps | The lightweight version of Visual Overview shown in DataMiner web apps is now considered legacy UI. [Several capabilities](xref:DashboardVisualOverview#unsupported-capabilities) of Visual Overview in Cube are not supported in the web apps. | End of Support to be confirmed. | -->
```
MariekeGO, line 61: As the table lists the current status, it might be confusing to add these already, hence why I added them in comment so we can add them during the 10.7 update early next year. Is this OK for you, or would you prefer to have them in the table already, but with the clarification "End of Engineering as of DataMiner 10.7.x"? (Same also for .NET Remoting below.)

dataminer/Reference/Software_support_life_cycles.md
```diff
@@ -124,6 +136,11 @@ The following features currently have the "End of Life" status. They are no long
 | Support for Visual Studio 2017 | End of Life as of DIS 2.41. DIS will no longer be compatible with this Visual Studio version. |
 | Support for Visual Studio 2019 | End of Life as of DIS 3.0. DIS will no longer be compatible with this Visual Studio version. |
 
+<!-- 
+When 10.7.0.x is released, add the following item:
+| Support for OpenSearch 1.x | OpenSearch 1.x reached community end of life on May 6, 2025 and should no longer be used. |
```
MariekeGO: For now, I've added this in comment since the version when this goes into effect is DataMiner 10.7; however, given that it's already no longer supported, perhaps we should add this one in this EoL table already instead of waiting for the DataMiner 10.7 release?

MariekeGO (commented):
@SimonRaine, could you take a look at my comments for this PR?

dataminer/Reference/Software_support_life_cycles.md
SimonRaine: OK

dataminer/Reference/Software_support_life_cycles.md
SimonRaine: OK

SimonRaine (approved):

PR #6711, opened 2026-08-21
Update docs with latest configurator changes and image changes

ThomasVerschuere:


dataminer/Getting_started/Installing_a_DMA/Unattended_DM_installation.md
```diff
@@ -18,17 +18,32 @@ cd "C:\Skyline DataMiner\Tools\FirstStartupChoice"
 > [!TIP]
 > For details about the different fields that can be configured, refer to [Running the Installer](xref:Installing_DM_using_the_DM_installer#running-the-installer).
 
+### Perpetual licensing
```
MariekeGO: Could you also include a full example for perpetual licensing? As the introduction at the top states that users have to configure a JSON file as indicated in the examples, they'll expect to find one for this as well.

ThomasVerschuere: It's already included in the examples below?

MariekeGO: I see! The new page structure is confusing, because as a user, you'll expect a series of examples only, not a subsection with separate information on specific examples first. I'll move the info so it becomes part of the example sections.

MariekeGO (approved):

PR #6715, opened 2026-08-25
Update migration tool version for DataMiner

plehotarik:
The reindexing tool has been updated to .NET 10

The reindexing tool has been adapted to force a rollover for TTL rollover indices at the end of reindexing. This is to ensure that the index provided name is preserved on the newest empty index. The reindexed indices can not have this name preserved due to this field being controlled by elastic itself and we can only preserve the old date name of the index with a string literal provided name.

This should help prevent loss of data and generation of erroneous indices.


MariekeGO (commented):
I'm a little confused by this change. Has the tool been removed from DataMiner versions prior to 10.6.0 [CU7]/10.6.10? That's what seems to be suggested here. 
Based on the RN, I get the impression that the tool has been improved, and would be best to use the latest version, but it's nevertheless still possible to run the tool if you are using an older DataMiner version. If this is indeed the case, could you adjust your pull request to prevent this confusion?

MariekeGO (approved):

PR #6716, opened 2026-08-25
Update migration guide with regex configuration note

plehotarik:
Added note about enabling regex for reindexing tool in Elasticsearch 7.10.0.

### DIFF
--- a/dataminer/Administrator_guide/Databases/Indexing_database/Migrating_from_Elasticsearch_to_Opensearch.md
+++ b/dataminer/Administrator_guide/Databases/Indexing_database/Migrating_from_Elasticsearch_to_Opensearch.md
@@ -35,7 +35,7 @@ To use this tool, follow the instructions below:
 
 ## Take a snapshot of the Elasticsearch 6.8.22 cluster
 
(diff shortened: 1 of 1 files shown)


dataminer/Administrator_guide/Databases/Indexing_database/Migrating_from_Elasticsearch_to_Opensearch.md
```diff
@@ -109,6 +109,14 @@ Using Kibana, you can take a snapshot in the following way:
 
 ## Restore the snapshot on an Elasticsearch 7.10.0 cluster
 
+> [!NOTE]
+> For the reindexing tool to work properly, we need to remove all complexity and character limits on regular expressions used within Elasticsearch. To do this, add    
+```txt
+   script.painless.regex.enabled: true
+   ```
+to your Elasticsearch 7.10 configuration.
```
MariekeGO: Is this an extra step users should always take before they do the procedure below? Because if this is the case, it would be better to just include it as the first step of the procedure, instead of putting the info in a separate note.

MariekeGO (approved):

PR #6717, opened 2026-08-25
Clarify DB.xml sync behavior for CloudStorage (STaaS)

LuisFreitas-Skyline:
The general database settings in DB.xml are also fully synced across the cluster when type=CloudStorage is used, same as CassandraCluster. Proxy settings are never synced and remain agent-specific, and for CassandraCluster the DBServer settings are excluded from the sync.


LuisFreitas-Skyline:
FYI @ruben12321

MariekeGO (approved):

PR #6726, opened 2026-08-27
Cost & Billing release notes

MariekeGO:
@Daniela-Prada, since Cost & Billing has not been released yet, should I add a preview label to these release notes? Or can we simply wait until the release to publish them?


Daniela-Prada:
Yes we can publish the release notes.

PR #6740, opened 2026-08-31
Support lifecycle update

MariekeGO:


dataminer/Administrator_guide/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_a_DMS_Cassandra_cluster.md
```diff
@@ -21,22 +21,20 @@ The Cassandra Cluster Migrator tool (called *SLCCMigrator.exe*) is available on
...
+- A Cassandra cluster must be available using [a supported version](xref:DataMiner_Compute_Requirements#cassandra-software). For information on how to install Cassandra, see [Installing Cassandra on a Linux machine](xref:Installing_Cassandra).
 
-  > [!NOTE]
-  > Previously Cassandra 3.11.8 was also supported. This will remain supported for existing installations; however, because of its increased performance, for new Cassandra cluster installations, Cassandra 4.0 or higher is required and **Cassandra 5.0 is recommended**. If you have a Cassandra 3.11.8 database and you have not yet migrated your DataMiner data, we recommend upgrading to Cassandra 5.0 first.
-
-- Either an OpenSearch cluster should be available (recommended), or an Elasticsearch cluster using version 6.8.0 or higher, but lower than 7.0.
+- Either an OpenSearch cluster should be available using [a supported version](xref:DataMiner_Compute_Requirements#opensearch-software) (recommended), or an Elasticsearch cluster using version 6.8.0 or higher, but lower than 7.0.
```
OnyxBit, line 26: Whilst supported on the 10.6 still. Do we want to drop Elasticsearch here so that for new migrations we always recommend or go with OpenSearch as the standard?

OnyxBit (approved):
Minor comment regarding Elastic being documented in the migration to clustered DB still, as it might be better to urge towards solely OpenSearch there imho.

dataminer/Administrator_guide/Databases/Configuring_dedicated_clustered_storage/Migrating_existing_systems/Migrating_the_general_database_to_a_DMS_Cassandra_cluster.md
MariekeGO: The reason why I kept it here is because lower in the text, it mentions "If there is an **existing OpenSearch/Elasticsearch cluster** connected to the DMS, this cluster will be reused" - and in most cases this will actually be an Elasticsearch cluster, which is why I then also added the note "Keep in mind that **support for Elasticsearch ends in DataMiner 10.7.x**, so any existing Elasticsearch cluster will need to be migrated to OpenSearch."

PR #6775, opened 2026-09-04
Remove icons from RS_Resource_Icons.md

JensVandewalle:
Removed LNB, SSPA and Transponder icon as they are now supported by SatOps solution


MariekeGO (commented):
If these icons aren't used anywhere else on docs, could you also remove them from the images folder? We want to avoid orphan images taking up space while they're actually no longer used.

JensVandewalle:
> If these icons aren't used anywhere else on docs, could you also remove them from the images folder? We want to avoid orphan images taking up space while they're actually no longer used.

@Joachim-Ally-Skyline will you use these icons in SatOps documentation? If not, I'll remove them from the images folder.

Joachim-Ally-Skyline:
Aha my excuses Jens for the late reply and thanks for the reminder. I understand your question. Yes it's the goal to make a icon page "at our side" with those three icons, and the two new ones, so it is indeed good if you can just keep the icons there. now they are unreferenced but that it then only temporary, I put it on my list to have this on the docs soon when we are going to release our 2.0 version.

Joachim-Ally-Skyline:
But maybe... since we need to add two images anyway, maybe it makes sense to remove them now, and later I will just add all 5 in my pullr equest, it will be cleaner now that I thought of it more.

MariekeGO (approved):

PR #6776, opened 2026-09-04
Update semantic versioning guidelines for NuGet packages and Catalog items

jarno-lernou-skyline:
Clarify the definitions of major, minor, and patch version increments in the semantic versioning guidelines to enhance understanding and adherence. Adjustments include specifying conditions for breaking changes and compatible dependency updates.

### DIFF
--- a/develop/TOOLS/NuGet/Producing_NuGet.md
+++ b/develop/TOOLS/NuGet/Producing_NuGet.md
@@ -149,9 +149,9 @@ We want to adhere to the [Semantic Versioning for NuGet packages](https://learn.
 
 A specific version number is in the form *Major.Minor.Patch[-Suffix]*, where the components have the following meanings:
 
-- *Major*: Breaking changes (API breaking changes, DataMiner minimum dependency changes, etc.)
-- *Minor*: New features, but backwards compatible
-- *Patch*: Backwards compatible bug fixes only
+- *Major*: Breaking changes, including API breaking changes or an increase to the minimum required dependency version
+- *Minor*: New features or compatible dependency updates
+- *Patch*: Backwards-compatible bug fixes only
 - *-Suffix* (optional): A hyphen followed by a string denoting a pre-release version.
 
 ## Package Metadata Conventions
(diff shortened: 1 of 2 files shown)


jarno-lernou-skyline:
@abelleite , what do you think about this change?

I saw two different explanations about version semantics and wanted to streamline. If you increase the prerequisites (e.g. DM version), do you increase A or B in A.B.C?

jarno-lernou-skyline (commented):
Given version **A.B.C**, Guidance use case:

- **_People forgot to set the right minimum DM version and will correct it now_**: 
  - **Action**: Increase version A
  - **Rationale**: Although the catalog item itself may not introduce a new feature or breaking code change, increasing the minimum supported DataMiner version removes support for a previously advertised compatibility range. Customers running a lower DataMiner version will no longer be able to deploy the updated catalog item and will need to upgrade before benefiting from future fixes or features.
- **_Including a new soft dependency._**
  - **Explanation**: An example of a soft dependency is a catalog item that is used but not needed for a given catalog item. It will still function, but by having the soft dependency, you will get more features.
  - **Action**: Increase version B
  - **Rationale**: A soft dependency provides additional functionality but is not required for the catalog item to function correctly. The catalog item must continue to operate when the dependency is unavailable. Adding such support introduces new functionality without breaking existing usage.
- **_Updating a soft dependency_**.
  - **Action**: Increase version B or C, depending on the impact.
  - **Rationale**:Use B if the update provides new functionality to users. Use C if it is purely an internal, backward-compatible fix.
- **_A hard dependency does not automatically mean a major version bump._**.
  - Patch update of a hard dependency
    - **Action**: Increase version C
    - **Example**: The new version of Library X only fixes an internal bug. Its public interface and behaviour relevant to your catalog item remain unchanged.
  - **_New functionality in a hard dependency_**
    - **Action**: Increase version B
    - **Rationale**: update your catalog item to use that new API to provide an additional feature. Existing functionality remains unchanged.
  - **_Increasing the minimum dependency version_**
    - **Action**: Increase version A
    - **Rationale**: removed a previously supported configuration which is no longer working

PR #6801, opened 2026-09-09
Add documentation regarding fetching credentials in GQI extensions (RN46279)

KlaasSL:


develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -0,0 +1,38 @@
...
+# ICredentialProvider interface
+
+## Definition
+
+- Namespace: `Skyline.DataMiner.Core.GQI.Extensions`
+- Assembly: `Skyline.DataMiner.Core.GQI.Extensions.dll`
+
+Provides access to username and password credentials that have been granted to a GQI extension library.
+
+Available from DataMiner 10.6.11/10.7.0 [CU0] onwards.
```
skyline-rge: The underlying feature is only available on those versions, but the ICredentialProvider itself is a valid service available in GQI in older main releases.
To be consistent we should use the versions in which it can be used, but then add an important note or something that DataMiner support for it is only available from the new feature and main release versions.

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Methods
+
+### Task\<UsernamePasswordCredential\> GetUsernamePasswordCredential(string name, CancellationToken cancellationToken = default)
+
+Retrieves a username and password credential by its configured name.
+
+#### Parameters
+
+- `string` `name`: The name of the credential reference granted to the extension library.
```
skyline-rge, line 28: Could we link to a page here about how to link credentials to an automation script?

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -0,0 +1,38 @@
...
+- `CancellationToken` `cancellationToken`: A token that can be used to cancel the credential request.
+
+#### Returns
+
+A task whose result contains the requested [UsernamePasswordCredential](xref:GQI_UsernamePasswordCredential).
+
+#### Exceptions
+
+- `GenIfException`: The credential name is empty or the named credential has not been granted to the extension library.
```
skyline-rge: Should say "...is null" (in that case, it should probably throw an `ArgumentNullException` instead) or otherwise more generally "...is invalid"

develop/devguide/GQI_Extensions/API_Reference/UsernamePasswordCredential.md
```diff
@@ -0,0 +1,31 @@
...
+# UsernamePasswordCredential class
+
+## Definition
+
+- Namespace: `Skyline.DataMiner.Core.GQI.Extensions`
+- Assembly: `Skyline.DataMiner.Core.GQI.Extensions.dll`
+
+Represents a username and password credential returned by the [ICredentialProvider](xref:GQI_ICredentialProvider).
+
+Available from DataMiner 10.6.11/10.7.0 [CU0] onwards.
```
skyline-rge: Idem about the versioning.

develop/devguide/GQI_Extensions/GQI_Extension_API.md
```diff
@@ -18,3 +18,7 @@ In order to create a GQI extension, you use the GQI extension API. This API can
 
 > [!TIP]
 > To ease the transition of extensions that were built against the legacy API, it is possible to combine both APIs within extension libraries and within extensions. If a data source is implemented with both the Core.GQI package and the legacy API, GQI uses the Core.GQI implementation when it is supported.
+
+## Accessing credentials
+
+The Core GQI extension API provides the [ICredentialProvider](xref:GQI_ICredentialProvider) service for retrieving username and password credentials that have been granted to an extension library. See [Using credentials in GQI extensions](xref:GQI_Extensions_Credentials) for configuration and usage details.
```
skyline-rge: Random to add this here?
This page is just all about the API, not specific features.

develop/devguide/GQI_Extensions/GQI_Extensions_Services.md
```diff
@@ -182,6 +182,7 @@ The following types can be used as constructor parameters:
 |--|--|--|
 | [IGQILogger](xref:GQI_IGQILogger) | A logger for the extension or service. | Extensions and services |
 | [IGQIFactory](xref:GQI_IGQIFactory) | Factory methods to create GQI objects. | Extensions and services |
+| [ICredentialProvider](xref:GQI_ICredentialProvider) | Access to username and password credentials granted to the extension library. | Extensions |
```
skyline-rge: We can be more general here in anticipation of future credential types.

```suggestion
| [ICredentialProvider](xref:GQI_ICredentialProvider) | Access to credentials granted to the extension library. | Extensions |
```

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+description: Learn how to grant and retrieve DataMiner Credentials Library username and password credentials in a GQI extension.
+---
+
+# Using credentials in GQI extensions
+
+The Core GQI extension API can retrieve credentials from the [DataMiner Credentials Library](xref:Credentials_Library).
+
+Currently, only credentials of type *Username and password credentials* are supported. Other credential types in the Credentials Library cannot be retrieved through this API.
+
+This functionality is available from DataMiner 10.6.11/10.7.0 [CU0] onwards when using version 1.5.0 or later of the `Skyline.DataMiner.Core.GQI.Extensions` NuGet package.
```
skyline-rge: I would explain the difference between when the types are available in GQI vs. availability in DataMiner here.

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+uid: GQI_Extensions_Credentials
+keywords: GQI extension credentials
+description: Learn how to grant and retrieve DataMiner Credentials Library username and password credentials in a GQI extension.
+---
+
+# Using credentials in GQI extensions
+
+The Core GQI extension API can retrieve credentials from the [DataMiner Credentials Library](xref:Credentials_Library).
+
+Currently, only credentials of type *Username and password credentials* are supported. Other credential types in the Credentials Library cannot be retrieved through this API.
```
skyline-rge: Sounds a bit negative.
Suggestion:

> Currently, the API already supports *username and password* credentials. More credential types may be added in the future.

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+
+This functionality is available from DataMiner 10.6.11/10.7.0 [CU0] onwards when using version 1.5.0 or later of the `Skyline.DataMiner.Core.GQI.Extensions` NuGet package.
+
+> [!IMPORTANT]
+> Credentials are only available to an extension library when they have been explicitly granted to that library.
```
skyline-rge, line 17: Link to docs?

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Granting a credential to an extension library
+
+1. Add a credential of type *Username and password credentials* to the DataMiner Credentials Library in DataMiner Cube. For more information, see [Credentials Library](xref:Credentials_Library).
+
+2. In the Automation module in DataMiner Cube, configure the credential for the Automation Script that contains the extension library. In the script's *CREDENTIALS* section, declare the credential as described in [Using credentials in an automation script](xref:Using_credentials_in_an_automation_script#declaring-a-set-of-credentials). Give the credential reference a name, such as `MyUserNamePasswordCredential`, and select the corresponding credential from the Credentials Library.
+
+The name configured for the credential reference is the name that the extension uses to request the credential.
```
skyline-rge: Isn't this duplicate info from the links mentioned here?

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+    public ExternalDataSource(ICredentialProvider credentialProvider)
+    {
+        _credentialProvider = credentialProvider;
+    }
+
+    // Implement the IGQIDataSource lifecycle methods.
+}
+```
+
+## Retrieving the credential
```
skyline-rge:
```suggestion
## Retrieving credentials
```

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+        .GetUsernamePasswordCredential("MyUserNamePasswordCredential")
+        .GetAwaiter()
+        .GetResult();
+}
+```
+
+> [!NOTE]
+> `GetUsernamePasswordCredential` returns a task. GQI extension lifecycle methods are synchronous, so call `.GetAwaiter().GetResult()` there. If you call it from an asynchronous helper, await it directly. Retrieve the credential once before fetching data rather than requesting it for every row or page.
+
+Treat the values retrieved from the credential library as secrets: never log them, include them in query results, or add them to exception messages.
```
skyline-rge: Add this in a important/warning note?

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+{
+    return _credentialProvider
+        .GetUsernamePasswordCredential("MyUserNamePasswordCredential")
+        .GetAwaiter()
+        .GetResult();
+}
+```
+
+> [!NOTE]
+> `GetUsernamePasswordCredential` returns a task. GQI extension lifecycle methods are synchronous, so call `.GetAwaiter().GetResult()` there. If you call it from an asynchronous helper, await it directly. Retrieve the credential once before fetching data rather than requesting it for every row or page.
```
skyline-rge: I would rephrase this a little, and also split it into 2 separate unrelated notes: 1 about awaiting the task, and 1 about caching the credential.

```suggestion
> Because GQI extension lifecycle methods are synchronous and `GetUsernamePasswordCredential` returns a task, you might need to block the thread to await the result with `.GetAwaiter().GetResult()`. If you call it from an asynchronous context, you can just await the task instead.
```

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -0,0 +1,69 @@
...
+
+Treat the values retrieved from the credential library as secrets: never log them, include them in query results, or add them to exception messages.
+
+## Troubleshooting
+
+- If the requested name is empty or is not configured for the extension library, the provider throws a `GenIfException`.
+- If the credential was removed from the Credentials Library or the DataMiner version does not support script credentials, credential retrieval fails. Verify the credential reference and the DataMiner version.
```
skyline-rge: Not sure how helpful this is.
A troubleshooting guid only works I think if we can phrase it as a list of `symptom -> problem/action/solution` instead of `problem -> symptom`

skyline-rge (changes requested):

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
KlaasSL: I don't really see the value of linking to the docs here again, the title below is "Granting a credential to an extension library" which contains all the info?

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
KlaasSL: Is it valuable to link here again, on line 18 we already point to the full documentation of the credentials in the extensions?

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -12,7 +12,7 @@ description: Reference for the ICredentialProvider interface, which retrieves us
 
 Provides access to username and password credentials that have been granted to a GQI extension library.
 
-Available from DataMiner 10.6.11/10.7.0 [CU0] onwards.
+Available from DataMiner Web 10.5.0 [CU20]/10.6.0 [CU8]/10.6.11 onwards when using version 1.5.0 or later of the `Skyline.DataMiner.Core.GQI.Extensions` NuGet package. <!-- RN 46279 -->
```
skyline-rge, line 15: Previously, we never mentioned the version of the NuGet directly in the docs before.
We relied on the version table in the NuGet description to map the DataMiner version to the API version.

If we want to start doing that, I would suggest a slighly less verbose wording:

Available from DataMiner Web 10.5.0 [CU20]/10.6.0 [CU8]/10.6.11 and API version 1.5.0.

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
skyline-rge: Users don't want to go to the full docs here, but specifically they want to know what a credential reference is and how to grant it.

develop/devguide/GQI_Extensions/API_Reference/UsernamePasswordCredential.md
```diff
@@ -12,7 +12,7 @@ description: Reference for the UsernamePasswordCredential class, which contains
 
 Represents a username and password credential returned by the [ICredentialProvider](xref:GQI_ICredentialProvider).
 
-Available from DataMiner 10.6.11/10.7.0 [CU0] onwards.
+Available from DataMiner Web 10.5.0 [CU20]/10.6.0 [CU8]/10.6.11 onwards when using version 1.5.0 or later of the `Skyline.DataMiner.Core.GQI.Extensions` NuGet package. <!-- RN 46279 -->
```
skyline-rge, line 15: Same as previous version comment.

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
skyline-rge: Yes, but it would at least be nice for users to link to that section explicitly, no?
Currently, the user implicitly needs to know that this will be explained in the next section and just needs to continue reading and in most cases, users like to read the docs in full that way.

+ I was also unsure whether we needed that section at all if we could just link directly to the "declaring a set of credentials section" for scripts.

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -21,8 +22,6 @@ This functionality is available from DataMiner 10.6.11/10.7.0 [CU0] onwards when
 
 2. In the Automation module in DataMiner Cube, configure the credential for the Automation Script that contains the extension library. In the script's *CREDENTIALS* section, declare the credential as described in [Using credentials in an automation script](xref:Using_credentials_in_an_automation_script#declaring-a-set-of-credentials). Give the credential reference a name, such as `MyUserNamePasswordCredential`, and select the corresponding credential from the Credentials Library.
 
-The name configured for the credential reference is the name that the extension uses to request the credential.
```
skyline-rge, line 24: Not sure why this was removed?
Note: my previous comment was a multi-line comment about the whole section, not just this line.

skyline-rge (changes requested):

skyline-rge (approved):

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -1,16 +1,19 @@
 ---
 uid: GQI_Extensions_Credentials
 keywords: GQI extension credentials
-description: Learn how to grant and retrieve DataMiner Credentials Library username and password credentials in a GQI extension.
+description: Learn how to grant and retrieve Credentials Library credentials in GQI extensions. Currently supported types are "username and password" and "token".
```
skyline-rge: **The** currently supported types...

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -1,16 +1,19 @@
...
 The Core GQI extension API can retrieve credentials from the [DataMiner Credentials Library](xref:Credentials_Library).
 
-Currently, the API already supports *Username and password credentials*. More credential types may be added in the future.
+The API supports the following credential types:
 
-The credential functionality within DataMiner is available from 10.7.0 [CU0]/10.6.10.
+- *Username and password*: Contains a username and password and is returned as a [UsernamePasswordCredential](xref:GQI_UsernamePasswordCredential).
+- *Token*: Contains an access token and is returned as a [TokenCredential](xref:GQI_TokenCredential).
+
+The credential functionality within DataMiner is available from 10.7.0 [CU0]/10.6.10 onwards.
```
skyline-rge: Since we need to mention so many different versions for different reasons, a "Prerequisites" section at the top with bulleted list might be better.

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -57,11 +65,25 @@ private UsernamePasswordCredential GetCredential()
...
 > [!NOTE]
-> Because GQI extension lifecycle methods are synchronous and `GetUsernamePasswordCredential` returns a task, you might need to block the thread to await the result with `.GetAwaiter().GetResult()`. If you call it from an asynchronous context, you can just await the task instead.
+> Because GQI extension lifecycle methods are synchronous and the credential retrieval methods return a task, you might need to block the thread to await the result with `.GetAwaiter().GetResult()`. If you call them from an asynchronous context, you can just await the task instead.
 
 > [!TIP]
 > To reduce latency and server load, retrieve the credential once and cache it before fetching data instead of requesting it for every row or page.
 
 > [!WARNING]
-> Treat the values retrieved from the credential library as secrets: never log them, include them in query results, or add them to exception messages.
+> Treat the values retrieved from the Credentials Library as secrets: never log them, include them in query results, or add them to exception messages.
```
skyline-rge: Since the examples for different credentials are in different sections, it now seems as if these notes are only relevant for the last section...

skyline-rge (changes requested):

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -0,0 +1,58 @@
...
+- `CancellationToken` `cancellationToken`: A token that can be used to cancel the credentials request.
+
+#### Returns
+
+A task whose result contains the requested [UsernamePasswordCredential](xref:GQI_UsernamePasswordCredential).
+
+#### Exceptions
+
+- `ArgumentNullException`: The credentials name is null.
+- `GenIfException`: The named credential has not been granted to the extension library.
```
MariekeGO: This isn't very clear to me. Do you mean that the extension library did not have permission to retrieve the named set of credentials? (Same for line 57.)
Note also that "credentials", in the sense of a set of data used to identify oneself, should ideally always be used as a plural. The singular form feels very unnatural (a bit like talking about "a scissor" instead of "scissors"). I've already adjusted this in several instances in this pull request, but not yet yere, as the meaning of this sentence isn't clear to me.

develop/devguide/GQI_Extensions/API_Reference/ICredentialProvider.md
```diff
@@ -1,6 +1,6 @@
 ---
 uid: GQI_ICredentialProvider
-description: "API reference for the ICredentialProvider interface used to retrieve username/password and token credentials granted to DataMiner GQI extension libraries."
+description: "API reference for the ICredentialProvider interface, which retrieves username and password credentials and token credentials for GQI extension libraries."
```
skyline-rge, line 3: This might be a difficult sentence to read with the multiple "and"s.
We could also just say something like "... to retrieve credentials for GQI extension libraries."

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -1,38 +1,43 @@
...
 keywords: GQI extension credentials
-description: Learn how to grant and retrieve Credentials Library credentials in GQI extensions. Currently supported types are "username and password" and "token".
+description: Learn how to grant and retrieve Credentials Library credentials in GQI extensions. The currently supported types are "username and password" and "token".
 ---
 
 # Using credentials in GQI extensions
 
+## Prerequisites
+
+- DataMiner 10.7.0 [CU0]/10.6.10 or later for the Credentials Library.
```
skyline-rge, line 11: Can we link this once to a RN too?

develop/devguide/GQI_Extensions/GQI_Extensions_Credentials.md
```diff
@@ -1,38 +1,43 @@
...
+- DataMiner Web 10.5.0 [CU20]/10.6.0 [CU8]/10.6.11 or later. <!-- RN 46279 -->
+- GQI extensions API version 1.5.0 or later. <!-- RN 46279 -->
```
skyline-rge, line 13: We could link to the API docs page here in case it would be unclear what this refers too.

skyline-rge (approved):

MariekeGO (approved):

PR #6834, opened 2026-09-15
DCP 302502: Update SRM resource status and obsolete availability modes

AndreRodriguesSkyline:
Updated docs to identify obsolete resource modes and clarify that the 'Maintenance' resource mode can be used to make resources unavailable for new bookings.


JorgeFilipeDias (approved):
lgtm

PR #6842, opened 2026-09-16
RN linked to softlaunch option

spetebroodt:


MariekeGO:
@spetebroodt, can you check if this can be incorporated in the add-RN prompt, so it first checks whether an RN is linked to a soft-launch option, and if yes, first verifies with the user if it's OK to add the RN before proceeding?

PR #6851, opened 2026-09-17
RN45976

mmphs:


release-notes/DxMs/DataAPI_change_log.md
```diff
@@ -7,6 +7,12 @@ uid: DataAPI_change_log
 > [!IMPORTANT]
 > At present, the Data API feature is only available in preview, if the soft-launch option *DataAPI* is enabled. For more information, see [Getting started with the Data Sources module](xref:Data_Sources_Setup).
 
+### 17 September 2026 - New feature - DataAPI 1.5.0 - Added support for pushing data to existing elements [ID 45976]
```
MariekeGO, line 10: There seems to be a mismatch between the solution version mentioned for this RN on Collaboration and here. Should the version on Collaboration be adjusted to 1.5.0?

release-notes/DxMs/DataAPI_change_log.md
```diff
@@ -7,6 +7,12 @@ uid: DataAPI_change_log
...
+
+Data API can now push data directly to an existing element, including a specific sub-target of that element. This allows data to be sent to elements that use an existing connector, rather than requiring Data API to create an auto-generated connector and element.
+
+When data is pushed to an existing element, its connector is reused without being changed.
+
```
MariekeGO, line 15: Should the "how to use" details from the Collaboration RN entry be included here as well? Or is there a specific reason why these should be left out for now?

release-notes/DxMs/DataAPI_change_log.md
mmphs: To get this fully working, this RN requires changes to both DataAPI and EdgeManager. EdgeManager has recently been migrated to GitHub, where the existing pipelines automatically set the solution version and do not support specifying multiple versions. As a result, the DataAPI 1.5.0.0 version had been overwritten.

I have manually updated it for now to include both versions.

release-notes/DxMs/DataAPI_change_log.md
mmphs: Personally, I think this is a feature that should be reserved for DataMiner EdgeNodes for now, which is why I left it out of the documentation. I’ve double-checked with the team to get their thoughts on this and will get back to you once I have more clarity.

If we decide not to document it, perhaps we should explicitly mention that this functionality is currently intended for scripts running on an EdgeNode.

mmphs: The team agreed to keep the implementation details internal for now. I added a note clarifying that this functionality is currently reserved for scripts running on an Edge Node.

MariekeGO (approved):

